PR #8, opened 2015-01-22
exclude packrat directory from deployment

jjallaire:
Assuming this is the behavior we want but submitting as a PR just as a sanity check.

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -11,6 +11,7 @@ bundleFiles <- function(appDir, rmdFile, fullNames) {
   files <- files[!grepl(glob2rx("*.Rproj"), files)]
   files <- files[!grepl(glob2rx(".DS_Store"), files)]
   files <- files[!grepl(glob2rx(".gitignore"), files)]
+  files <- files[!grepl(glob2rx("packrat/*"), files)]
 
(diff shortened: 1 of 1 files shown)


trestletech:
LGTM. Just so we're all on the same page, the `R/packrat` directory is where we can manage our own RSC-internal version of Packrat. If we go to load Packrat to do operations on an uploaded bundle and find that it's not yet installed into that directory, we'll install it from RStudio CRAN.

This will give us later flexibility if we need to install a particular version/branch, as we can do so at install time via R commands.

The only reason we'd want to include this dir in the bundle is if we had Packrat as a Git subrepo and wanted to lock versions of RSC to versions of Packrat (which we're not doing now).

jjallaire:
It seems like we could control the version of packrat on the server side by
just noting it in our upload manifest (rather than including the entirety
of the packrat directory which includes installed packages, etc.). Or if
that wasn't satisfactory we could specially include *just* the source code
to the version of packrat that was used to create the bundle.

It seems like this change is okay for now then. Correct?

trestletech:
Sounds right to me. We can refine exactly what we want. I think the current arrangement should get us safely to v1.

PR #10, opened 2015-01-30
Include app mode in bundle

jcheng5:
Possible values are "shiny", "rmd", and "staticRmd".
These values may need to change to align with connect.


R/bundle.R
```diff
@@ -145,6 +184,7 @@ createAppManifest <- function(appDir, files, users) {
   # create the manifest
   manifest <- list()
   manifest$version <- 1
+  manifest$appmode <- appMode
```
kippandrew: FWIW shinyapps uses schema validation for manifests. I believe this will break deploying to shinyapps.

jjallaire:
Two possible resolutions:

1. shinyapps does more liberal schema validation

2. we make appmode conditional on ! shinyapps

I'm open to either. Thoughts?

jjallaire:
Here's a PR to handle it on the rsconnect side: https://github.com/rstudio/rsconnect/pull/13

R/bundle.R
jcheng5: Nice catch @kippandrew

PR #12, opened 2015-02-09
Refine absolute path lint mechanics

kevinushey:
Previously, we used a set of regular expressions applied on strings in documents indiscriminately to try and pull entries out that looked like absolute paths. This caused a lot of false positives, especially with URIs beginning with `/`, for example (as those can 'look like' absolute paths to files on the local filesystem).

This PR refines that behaviour by:

1. Extracting all (single and double quoted) strings in the document of at least 5 characters in length, and
2. Warning only if they 'look like' absolute paths, _and_ they exist on the local filesystem.


jjallaire:
This looks good to me. @jmcphers could you take a quick look as well and merge if you think it's good to go.

PR #13, opened 2015-02-10
don't include appmode in manifest for shinyapps

jjallaire:
shinyapps.io does strong validation of deployment manifests so will fail if an extra field (appmode in this case) is included. This PR makes the accountInfo available to the manifest generation code and only includes appmode if we aren't running against shinyapps.io.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rsconnect
 Type: Package
 Title: Deployment Interface for Shiny Applications
-Version: 0.3.72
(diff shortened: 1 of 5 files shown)


jcheng5:
LGTM. Might even consider making it an error to try to publish a staticRmd to shinyapps.

jjallaire:
Good idea. Let's take care of that when we refactor to have a report specific deployment API (e.g. deployReport). I'm guessing this will be part of the work Jonathan does in the next few weeks.

PR #31, opened 2015-05-08
Locale detection fixes from shinyapps

kippandrew:
A few fixes to the locale detection code from shinyapps package. @jjallaire these really should go in before cutting the RC for the IDE as it fixes breakage for some locales.


jjallaire:
Thanks for getting that in! Merged and also moved over to v0.99 branch. This will be in the RC by the end of the day.

PR #42, opened 2015-06-11
Add all dependencies' DESCRIPTION files to bundle

jcheng5:
Related:
https://github.com/rstudio/connect/pull/514 (see here for summary)
https://github.com/rstudio/packrat/pull/225

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -338,6 +338,34 @@ addPackratSnapshot <- function(bundleDir, appMode) {
   if (file.exists(tempDependencyFile)) {
     unlink(tempDependencyFile)
   }
+
+  # Copy all the DESCRIPTION files we're relying on into packrat/desc.
+  # That directory will contain one file for each package, e.g.
+  # packrat/desc/shiny will be the shiny package's DESCRIPTION.
+  #
+  # The server will use this to calculate package hashes. We don't want
+  # to rely on hashes calculated by our version of packrat, because the
+  # server may be running a different version.
+  lockFilePath <- file.path(bundleDir, "packrat", "packrat.lock")
+  descDir <- file.path(bundleDir, "packrat", "desc")
+  tryCatch({
+    dir.create(descDir)
+    packages <- na.omit(read.dcf(lockFilePath)[,"Package"])
+    lapply(packages, function(pkgName) {
+      descFile <- system.file("DESCRIPTION", package = pkgName)
+      if (!file.exists(descFile)) {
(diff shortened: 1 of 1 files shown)


R/bundle.R
```diff
@@ -338,6 +338,34 @@ addPackratSnapshot <- function(bundleDir, appMode) {
...
+
+  # Copy all the DESCRIPTION files we're relying on into packrat/desc.
+  # That directory will contain one file for each package, e.g.
+  # packrat/desc/shiny will be the shiny package's DESCRIPTION.
+  #
+  # The server will use this to calculate package hashes. We don't want
+  # to rely on hashes calculated by our version of packrat, because the
+  # server may be running a different version.
+  lockFilePath <- file.path(bundleDir, "packrat", "packrat.lock")
+  descDir <- file.path(bundleDir, "packrat", "desc")
```
aronatkins: It has been requested that this not occur for shinyapps deployments.

R/bundle.R
```diff
@@ -338,6 +338,34 @@ addPackratSnapshot <- function(bundleDir, appMode) {
...
+  tryCatch({
+    dir.create(descDir)
```
aronatkins: Should we ever cleanup/prune this directory?

R/bundle.R
jcheng5: Oh, does addPackratSnapshot occur for shinyapps deployments? If so--does it need to?

R/bundle.R
jcheng5: It's in the bundle directory, I assumed the bundle directory got deleted after deploy. I'll check.

R/bundle.R
aronatkins: Sorry. I didn't follow through the whole call chain on this one. Andy wanted to make sure this data wasn't added to the bundle for shinyapps. I think the feeling is that shinyapps does not need it since it already has the package listings in `manifest.json`.

R/bundle.R
aronatkins: Ah, perfect. I couldn't remember what was next to the app and what was in the temp directory.

R/bundle.R
kippandrew: addPackratSnapshot does currently get called for shinyapps.io deploy, but I fixed that in #43

R/bundle.R
jcheng5: Confirmed that bundle directory gets deleted. https://github.com/rstudio/rsconnect/blob/a00776881338e1e1d168eb7730efca9fe766c003/R/bundle.R#L36

PR #48, opened 2015-06-15
shinyapps usage metrics

kippandrew:
This adds a new function showUsage() which returns application usage metrics (.e.g number of active hours) as a data frame.


jjallaire:
Let's just note in the docs that this is shinyapps only (e.g. as we do here: https://github.com/rstudio/rsconnect/blob/86fb562cd821300e4e210c3428d735969e89c431/R/configureApp.R#L86)

PR #58, opened 2015-06-16
remove server parameter from showLogs

jjallaire:
Since `showLogs` only works for shinyapps.io the `server` parameter is unnecessary.


jjallaire:
Address this issue : https://github.com/rstudio/rsconnect/issues/57

@kippandrew Could you review when you have a chance?

PR #68, opened 2015-07-14
Implement sendEmail option for addAuthorizedUser

kippandrew:
You can now choose not to automatically have an email sent when adding users to apps hosted on shinyapps.io. This is because some users prefer to send their own messages. Using `addAuthorizedUser(..., sendEmail=FALSE)` to do this. If the user does not have an account you can use  `showInvited()` to retrieve the invite link and send it to them via whatever means suites you (a.k.a carrier pigeon, smoke signals, telepathy, vulcan mind-meld, etc.).

### DIFF
--- a/R/auth.R
+++ b/R/auth.R
@@ -40,11 +40,13 @@ cleanupPasswordFile <- function(appDir) {
 #'   system then this parameter can be omitted.
 #' @param server Server name. Required only if you use the same account name on
 #'   multiple servers.
+#' @param sendEmail Send and email letting the user know the application
+#'   has been shared with them.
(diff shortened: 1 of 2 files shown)


R/auth.R
```diff
@@ -40,11 +40,13 @@ cleanupPasswordFile <- function(appDir) {
 #'   system then this parameter can be omitted.
 #' @param server Server name. Required only if you use the same account name on
 #'   multiple servers.
+#' @param sendEmail Send and email letting the user know the application
```
jjallaire: "Send an email"

PR #71, opened 2015-07-21
Add an has_parameters property to the manifest... and other param-Rmd support

aronatkins:
~~~Not yet done: Scanning the parameters for library dependencies and files we should include. Thoughts about how to approach that?~~~

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -149,10 +149,13 @@ inferDependencies <- function(appMode) {
   if (grepl("\\brmd\\b", appMode)) {
     deps <- c(deps, "rmarkdown")
   }
-  if (grepl("\\bshiny\\b", appMode) || identical("rmd-param", appMode)) {
+  if (identical("rmd-param", appMode)) {
     deps <- c(deps, "shiny")
   }
-  deps
+  if (grepl("\\bshiny\\b", appMode)) {
+    deps <- c(deps, "shiny")
+  }
+  unique(deps)
 }
 
 createAppManifest <- function(appDir, appMode, contentCategory, accountInfo,
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -50,7 +50,7 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
   if (!isShinyapps(accountInfo)) {
     # infer package dependencies for non-static content deployment
     if (appMode != "static") {
-      addPackratSnapshot(bundleDir, appMode)
+      addPackratSnapshot(bundleDir, inferDependencies(appMode))
     }
   }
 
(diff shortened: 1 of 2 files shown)


jjallaire:
For scanning the parameters for dependencies, I think all you need to do is extract the `expr` from the knit_params and then put those through the standard expression dependency codepath (https://github.com/rstudio/rsconnect/blob/master/R/dependencies.R#L137)

aronatkins:
`expressionDependencies` takes an expression, not a string but the `knitr::knit_params` gives `param$expr` as a string. I've tried `as.expression(param$expr)` and `expression(param$expr)` without luck. Do you have a recommended way to get the parsed expression?

Basically, I need to get this to work and then everything else falls into place (it's contrived, but perhaps sufficient?):

```
expr <- "shiny::actionButton('a','b')"
deps <- rsconnect:::expressionDependencies(as.expression(expr))
```

jjallaire:
I think parse(text = expr) should do it.

aronatkins:
We haven't decided if we want a new app mode.. This PR is still WIP.

aronatkins:
Related: https://github.com/rstudio/packrat/pull/232

aronatkins:
@trestletech @jmcphers @jjallaire I think this is ready for eyes. Note that you should have:

```
devtools::install_github("rstudio/rmarkdown", ref="aron/knit_params_ask")
devtools::install_github("aronatkins/knitr", ref="knit_params.yaml")
devtools::install_github("aronatkins/packrat", ref="aron/process-rmd-parameter-expressions")
```

R/bundle.R
```diff
@@ -80,7 +82,24 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
   bundlePath
 }
 
-isShinyRmd <- function(filename) {
+appHasParameters <- function(appDir, files) {
+  rmdFiles <- grep("^[^/\\\\]+\\.rmd$", files, ignore.case = TRUE, perl = TRUE,
```
jmcphers: Are parameters supported for R Markdown documents other than the primary one in an application?

aronatkins: `rmarkdown` only supports parameters in the top-level document right now. This scans everything. Do you think we should only scan the primary doc?

jmcphers: No, I don't think we need to try to figure out if a doc is nested (seems like a lot of engineering and the worst thing that could happen is a superfluous dependency on Shiny). I'm asking about the case wherein you publish several top-level docs as part of a multi-page app.

aronatkins: Ah. We only support single-doc rendering. It's fine to have multiple HTML artifacts, but we only render a single Rmd at present. I suppose when we eventually have multiple docs, we might want metadata about each file.

jmcphers: Gotcha. Maybe we just want to scan the primary Rmd for now, then, but not a big deal either way.

aronatkins:
@jmcphers This is set again. Do we want to have it only scan the primary Rmd?

jmcphers:
Nope, I don't think that's necessary. LGTM as-is.

PR #76, opened 2015-10-14
Use packrat to generate shinyapps.io manifests

kippandrew:
Previously content being deployed to connect used `packrat` to detect dependencies whiling still generating the manifest from code within the `rsconnect` package.  The goal of this PR is to generate the manifest using the same `packrat` snapshot that connect uses. This eliminates the need for the old dependency detection code. This also has the benefit of allowing content deployed to shinyapps.io to understand dependencies detected by `packrat` including 3rd-party CRAN-like repositories and other sources in the future.

NOTE: There is shinyapps.io server side changes that must be deployed for this to work so don't merge without checking with me.

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -27,17 +27,11 @@ bundleFiles <- function(appDir) {
   files
 }
 
-bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
-                      contentCategory, accountInfo) {
-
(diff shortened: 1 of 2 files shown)


kippandrew:
Related connect PR: https://github.com/rstudio/connect/pull/912/files

kippandrew:
@jjallaire i think this is good to go.

PR #85, opened 2015-12-03
retain global deployment history

jjallaire:
We now record all deployments in a centralized deployments/history.dcf file. This is done so that front ends can show a view of all recent / previously published content.

Note that we record the data in a single file that is arranged in reverse chronological order (this is so that a front end interested in e.g. only the last 30 deployments could read that far and stop).

There is no limit to the size of the file but note that 4 deployments take about 1K of space so a 1MB file would cover ~ 4000 deployments.

@jmcphers

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rsconnect
 Type: Package
 Title: Deployment Interface for Shiny Applications
-Version: 0.4.1.9
+Version: 0.4.1.10
 Date: 2015-11-03
 Author: JJ Allaire
 Maintainer: JJ Allaire <jj@rstudio.com>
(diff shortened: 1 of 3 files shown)


jmcphers:
Tried this out locally, looks good to me. Do we want to make any effort to coalesce deployment records of the same `appPath`? I think this would go a long way towards keeping the content from growing too large and would also simplify the parsing task for front ends (which are unlikely to want to display repeated deployments of the same content), although obviously at the expense of either parsing on every deployment or moving away from a single-file system.

jjallaire:
I'm okay with the single-file system for now (I think coalescing will be
easy for front-ends, just ignore subsequent deployments of any appPath
already encountered). Moving to a multiple file system requires us to have
some sort of key-generate scheme, which could be a hash of the appPath but
that wouldn't be guaranteed conflict-free.

aronatkins:
Front-ends won't want to show repeated deployments of the same content to the same destination, but it can be valuable to show different destinations. For example, I might have the same content deployed to a staging and production server. Or I might (for whatever reason) deploy a piece of content under two different names.

PR #96, opened 2016-03-24
Implement deploySite function

jjallaire:
This PR makes use of the [`site_generator`](https://github.com/rstudio/rmarkdown/pull/653) work in **rmarkdown** to implement a high-level `deploySite` function. A couple of assumptions regarding how RSC will handle these sites on the server are baked in here so we should discuss and then revise as appropriate. 

We now deploy with `contentCategory = "site"` (and this is not customizable by design so that RSC can trigger different logic for this category of content). We had previously gone in the direction of `contentCategory = "book"` then simply treating it as a "Document" in the UI. For the time being we can allow "site" to default into the "Documents" category as "book" did however I'd propose that we eventually create a new category named "Websites" which will include books, blogs, sites generated automatically from data analysis projects, etc. 

We now have a `sourceCode = FALSE` parameter that indicates whether this is a static publish or a publish with source code. Static publish works today without any changes to RSC, however source code publish will need something like the following:

1. If a new bundle is deployed with `contentCategory = "site"` and `sourceCode = TRUE` then call the `rmarkdown::site_generator` function in order to build the site and ascertain what directory it's output should be served from (e.g. '.', '_site', or '_book')

2. Serve requests to the deployed application from the output directory (e.g. _book) rather than the root.

3. Realize that this is a class of "report" that cannot be attached to an email and behave accordingly when sending out emails.

Paramaterized websites are not yet addressed (and will take some work in a number of packages to get right so I'd suggest we add this later if there is demand for it).

@jmcphers @aronatkins @jcheng5 @yihui @hadley @tareefk @trestletech @CBarrafod @jonyoder @chaitac


jjallaire:
Talking with @tareefk, he is in agreement that we can classify books as "Websites" but would like to preserve the flexibility to distinguish them in the UI down the line. A can think of a couple ways to do this:

1. Introduce a "contentSubCategory" construct

2. Allow the contentCategory field to carry a subcategory (e.g. `site:book`)

I personally like #2. Interested in what other folks think.

jmcphers:
I prefer #2 as well. Alternately, there may not be a need to fully express the taxonomy in the manifest at all -- we could just say `book` and have the knowledge that `book` is a subcategory of `site` live on Connect. (Of course this implies that you can't upload a `book` to a version of Connect that doesn't yet have `book` in its taxonomy, so perhaps it'd create more version coupling than we want.)

jjallaire:
The problem is that Connect then needs to have a static list of all content
types that qualify as "site" (e.g. when "blog", "ProjectTemplate", etc.
types come along we'll need to update Connect so it knows to treat them as
sites.

To be clear, I'd be content to not have any subcategories at all right now
and just call everything "site" but I just wanted to make sure others
didn't feel that overly limiting.

aronatkins:
If we are just talking about pre-rendered content, single-page and multi-page structures are just fine when classified as "static" within Connect. This already how Connect understands and serve pre-rendered HTML from R Markdown documents and plots. The manifest contains "content category" metadata that helps with some presentation details (mostly for organization and iconography). There is no difference when it comes to serving the content.

Further, if Connect sees a content category that it does not understand, it will group it as a document with a generic document icon. So there's not really a strict coupling. If we start receiving interesting content category tags, we can add grouping/icons for those as we go. This means that if `content_category=blog`, we will display that in the dashboard as static HTML for now but could easily alter the iconography once it appears. This is really all we did when we started receiving "book".

I think my preference is for `rsconnect` to send a very specific content category in the manifest. If the type of content is not understood, then sending "site" (noting a multi-page collection) or an empty string (like we receive for R Markdown HTML) would be fine.

Jonathan and I talked last year about having "content_category" be a list of tags rather than just a singleton, but I think that's a more involved modeling change.

jjallaire:
Okay, I think this argues for just using "site" since that indicates a
special behavior (as opposed to a type to display in the UI). In this case
"site" means that RSC needs to invoke rmarkdown::site_generator to discover
exactly how to render the multi-page site and exactly where it's content is
generated into (e.g. "_book").

So net I'd vote to just keep it "site" and worry about sub categories later
and/or ascertain them through heuristics if we really want to distinguish.

If everyone agrees then I think we can merge (assuming there is not other
feedback/concerns).

How much work will it be on the RSC site to call rmarkdown::site_generator,
etc? I'd love to see us do this soon as it would make bookdown publishing
no longer a special case but just our first use case of website publishing.
I've got another website type cooking in rmarkdown (just a simple
multi-page html_document with a navbar) and have also updated the IDE to
automatically recognize website projects and call their site generator for
"Build All". The last step will be to hookup the Publish button so then
bookdown books, html_document sites, and any other type of website that is
cooked up can be both published to RSC via rsconnect::deploySite and the
IDE, but also scheduled for execution on the server like any other report.

aronatkins:
Ah, right. We're talking about rendering. Then I think I agree that we need a special app mode in the manifest. We can probably have something working early next week. Let me confirm with the team.

jjallaire:
@aronatkins Made the changes we discussed and cleaned up some other things. This PR should be good to go, let me know if there are any other issues you see. 

Note that we do (effectively) detect the site generator in packrat by adding a dependency for the default output format. It's technically possible to have a site generator in a different package from the default output format so we've logged a packrat bug to take care of that once a new version of rmarkdown with the `all_output_formats` function is on CRAN (see https://github.com/rstudio/packrat/issues/286)

The change to `rmarkdown::render_site` to return the output file will also go in shortly.

jjallaire:
rmarkdown::render_site change here: https://github.com/rstudio/rmarkdown/commit/b050b3fa79e6c33ab81b490a5f492913148e4f7e

It will always return a path relative to the input directory (e.g. "index.html" or "_book/index.html")

PR #103, opened 2016-05-23
improve error reporting for overly-large bundles

kevinushey:
This PR attempts to detect errors due to overly large bundles, and provide a prescription in the case where users are not using `curl` for deployment.

Should add some tests before merge.

### DIFF
--- a/R/http.R
+++ b/R/http.R
@@ -352,11 +352,31 @@ httpRCurl <- function(protocol,
     error = function(e, ...) {
       # ignore errors resulting from timeout or user abort
       if (identical(e$message, "Callback aborted") ||
-          identical(e$message, "transfer closed with outstanding read data remaining"))
-        return
+          identical(e$message, "transfer closed with outstanding read data remaining")) {
+        return()
+      }
+
+      # detect errors due to overly-large bundles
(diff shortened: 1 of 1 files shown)


R/http.R
```diff
@@ -352,11 +352,31 @@ httpRCurl <- function(protocol,
...
+        return()
+      }
+
+      # detect errors due to overly-large bundles
+      if (grepl("long vectors not supported yet", e$message, fixed = TRUE)) {
+
+        msg <- "Deployment failed due to overly-large bundle."
+
+        prescription <- if (!identical(getOption("rsconnect.http"), "curl"))
+          "Try setting 'options(rsconnect.http = \"curl\")' and re-deploying your application."
```
jjallaire: Note that this option *only* works reliably on OS X (it will probably never work on Windows and would only work on Linux if the user had installed the curl command line utility).

jjallaire:
I think we are going to want to replace the RCurl handler within http.R w/ a curl package handler (hopefully this would resolve the issue + other issues we are currently seeing).

R/http.R
kevinushey: Good point; I've modified the prescription to only report `curl` as a candidate if `curl` is available on the PATH.

PR #116, opened 2016-09-15
Modernise dependencies: use openssl pkg

jeroen:
The `PKI` package does not build anymore on MacOS (See e.g. https://github.com/s-u/PKI/issues/17).  This replaces `PKI` with the `openssl` package (which also has all of the hash/hmac/base64 functions).

I am not sure what is the best way to test this package, but from what I can see there should be no side effects.

If it is useful, I will gladly try to replace RJSONIO/RCurl with `jsonlite/curl` as well.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -10,8 +10,7 @@ Description: Programmatic deployment interface for 'RPubs', 'shinyapps.io', and
 Depends:
     R (>= 3.0.0)
 Imports:
-    digest,
-    PKI,
+    openssl,
     RCurl,
     RJSONIO,
     packrat (>= 0.4.8-1),
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -421,8 +421,7 @@ createAppManifest <- function(appDir, appMode, contentCategory, hasParameters,
   # provide checksums for all files
   filelist <- list()
   for (file in files) {
-    checksum <- list(checksum = digest::digest(file.path(appDir, file),
-                                               algo = "md5", file = TRUE))
+    checksum <- list(checksum = as.character(openssl::md5(base::file(file.path(appDir, file)))))
     filelist[[file]] <- I(checksum)
   }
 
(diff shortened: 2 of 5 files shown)


jjallaire:
We would very much welcome this modernization. One proviso: we are just about to go GA on RStudio Connect (and submit rsconnect to CRAN to sync with that) so I don't want to make this change right now. After we submit to CRAN we'll do it which will give us a bit more time to make sure there are no unanticipated side effects or compatibility issues.

jeroenooms:
Sure that makes sense. Is there a way for me to test this package? Most of the unit tests are skipped because I don't have a connect server or so.

jjallaire:
You can add yourself to the beta server here:
https://beta.rstudioconnect.com/connect/#/welcome

You can also download an evaluation version that will work for 45 days from
here (and we'll be glad to extend it infinitely and/or just give you a
regular license): https://www.rstudio.com/products/connect/

The way HTTP works in the package is that there are multiple "methods"
(RCurl, curl via system, and raw sockets) which are dispatched to from
here: https://github.com/rstudio/rsconnect/blob/master/R/http.R#L413-L424.
To add support for the curl package you'd just implement the http function
for curl in a manner analogous to the other methods. Since we already have
a "curl" method that uses system perhaps the new method could be called
"curlpkg" or something like that. I'd imagine we'd want to make "curlpkg"
the new default method.

We use multiple methods because there end up being customers that for one
reason or another can't connect with RCurl. Perhaps the curl package will
be more robust and we can scrap the whole multi-method thing once we
establish that it works better.

J.J.

jmcphers:
Thanks for the update!

It looks like this needs a few changes to work; in particular when computing MD5 signatures, it's opening files in text mode. It's probably not easy for you to iterate on this so I'll do so and then merge once I have a working set of changes.

jmcphers:
Okay, I've got it working locally. Very much appreciate the contribution!

jeroen:
Awesome, thanks. Yes it's difficult to properly test this without a connect server :)

Issue #137, opened 2016-10-28
setProperty fails with authInfo error


jspiewak:
```
> rsconnect::setProperty('application.visibility', 'private')
Error in authInfo$secret : $ operator is invalid for atomic vectors
```

kevinushey:
If I understand correctly, this occurs because `PUT_JSON` is called incorrectly in the `setApplicationProperty` function (https://github.com/rstudio/rsconnect/blob/1230e532294cf7f9e341533631d8270aae3a00d9/R/lucid.R#L86-L94):

```
    setApplicationProperty = function(applicationId, propertyName,
                                      propertyValue, force=FALSE) {
      path <- paste("/applications/", applicationId, "/properties/",
                    propertyName, sep="")
      v <- list()
      v$value <- propertyValue
      query <- paste("force=", if (force) "1" else "0", sep="")
      handleResponse(PUT_JSON(authInfo, path, v, query))
    }
```

Note that we're not passing along the `service` argument to `PUT_JSON`. I'll submit a PR.

PR #139, opened 2016-10-28
add missing 'service' parameters

kevinushey:
Should resolve https://github.com/rstudio/rsconnect/issues/137.

### DIFF
--- a/R/lucid.R
+++ b/R/lucid.R
@@ -80,7 +80,7 @@ lucidClient <- function(service, authInfo) {
 
     listApplicationProperties = function(applicationId) {
       path <- paste("/applications/", applicationId, "/properties/", sep="")
-      handleResponse(GET(authInfo, path))
+      handleResponse(GET(service, authInfo, path))
     },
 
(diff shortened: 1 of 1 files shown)


kevinushey:
Although the request is made successfully, the server doesn't seem to like it:

```
> rsconnect::setProperty('application.visibility', 'private', account = 'lodashkev', appName = 'rplot')
Error: HTTP 400
PUT https://beta.rstudioconnect.com/__api__/applications/2048/properties/application.visibility?force=0
The requested endpoint or method is invalid. 
```

Note that this code path uses the `lucid` client; perhaps we also need to be detecting + using the `connect` client instead?

jspiewak:
#137 was against shinyapps.io, so `lucid` is correct.
I am not sure whether or not connect supports this.

kevinushey:
Got it -- then it should definitely work in that case. Thanks!

jspiewak:
poke?

kevinushey:
@jjallaire can you sanity check and merge if all looks fine?

PR #147, opened 2016-12-08
Open Config URL after Deployment Regardless of needs_config Flag

jonyoder:
See rstudio/connect#4306 and rstudio/connect#4230. We've decided to redirect users to the Dashboard URL on both initial deployments and subsequent updates. 

Connect (as of rstudio/connect#4306) will return the `config_url` regardless of whether `needs_config` is `TRUE`. Compatibility with older Connect versions is preserved by testing to see if a `config_url` is returned.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -329,11 +329,9 @@ deployApp <- function(appDir = getwd(),
     # if this client supports config, see if the app needs it
     if (!quiet && !is.null(client$configureApplication)) {
       config <- client$configureApplication(application$id)
-      if (config$needs_config) {
-        # app needs config, finish deployment on the server
-        showURL(config$config_url)
-        return(invisible(TRUE))
-      }
+      # open app in Dashboard for publishing or further configuration
+      showURL(config$config_url)
+      return(invisible(TRUE))
     }
 
     # launch the browser if requested
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -332,7 +332,7 @@ deployApp <- function(appDir = getwd(),
       # Open app in Dashboard for publishing or further configuration.
       # Preserve compatibility with older versions of connect by checking
       # to see if config$config_url is set.
-      if (!(is.na(config$config_url) || config$config_url == '')) {
+      if (!(is.null(config$config_url) || config$config_url == '')) {
         showURL(config$config_url)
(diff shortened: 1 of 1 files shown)


jjallaire:
@jmcphers I'm not sure about the possible values for the config_url as well as whether checking for `NA` and empty string are enough (couldn't it ever by `NULL`?). Could you take a look at this also?

jonyoder:
@jmcphers - I was working on the assumption that this block of code is only reached by Connect deploys (as opposed to shinyapps.io). If I'm correct, I think we're probably ok, since Connect historically returned an empty string for `config_url`. However, if this code block is reached by non-Connect deploys, then all bets are off.

jjallaire:
What is the `NA` check for? (I don't know if there is even a way for JSON
to end up as NA rather than NULL)

jonyoder:
I was thinking (likely incorrectly) that `is.na` would return true if the `config_url` property was missing from the JSON response. After searching, I see that we're elsewhere using a combination of `is.null`, `is.na`, and `length(x)==0` (see https://github.com/rstudio/rsconnect/blob/5565386d2de92f50cc3c8b4a56f0146d88609ec0/R/deployments.R#L8). However, that example doesn't involve a JSON property, and I'm not sure what's best in this case.

jonyoder:
I switched from using `is.na` to using `is.null`. @jmcphers, if you think we need an additional `is.na` test, let me know.

jmcphers:
Yes, this code is only going to get hit for Connect. It's guarded by `client$configureApplication`, which is `NULL` for the lucid client.

`is.null` is indeed the correct test. The output of the JSON parser is a list, so like any list it will report `NULL` if you access a named element that doesn't exist. 

Idiomatically I prefer `identical(x, '')` to test for the empty string as `x == ''` has pathological cases that can result in runtime errors (if e.g. `x <- character(0)`) but there's no chance of those here.

Presuming we've tested this against both new and old Connect servers, this change LGTM.

jjallaire:
@jonyoder Once you give the word on testing I'll merge.

jonyoder:
I tested this code against both new and old Connect servers when using the `is.na` test. I haven't retested since I switched to `is.null`. I can retest if you think it's necessary.

PR #158, opened 2017-02-24
Added note that terminating also archives the application

martinjhnhadley:
It isn’t clear from the documentation that termination is analogous
with archiving.


martinjhnhadley:
Failed to rebuild documentation before PR - fixed in new PR

PR #159, opened 2017-02-24
Added note that terminating also archives the application

martinjhnhadley:
It isn’t clear from the documentation that termination is analogous
with archiving.

### DIFF
--- a/R/terminateApp.R
+++ b/R/terminateApp.R
@@ -1,6 +1,6 @@
 #' Terminate an Application
 #'
-#' Terminate an application currently running on a remote server, this will also archive the application.
+#' Terminate and archive a currently deployed ShinyApps application.
 #'
 #' @param appName Name of application to terminate
 #' @param account Account name. If a single account is registered on the system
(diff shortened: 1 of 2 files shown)


jjallaire:
@trestletech Does terminateApp work on RSConnect or just shinyapps.io? If it does, does the archiving behavior differ between platforms? Just want to make sure that the docs reflect accurately what happens in both environments.

kippandrew:
Terminate and Archive are the same. FWIW, I would be in favor of deprecating the `terminateApp` function and renaming it to `archiveApp`. "Terminate" was never a good name to begin with, which is why it was changed in the web UI.

kippandrew:
An update to my pervious comment: 'terminate' and 'archive' mean the same things in shinyapps.io, not sure about RSC.

jjallaire:
Okay, so let's revise the PR to add a new function "archiveApp" and to have the old function alias the new one with a deprecation warning. Jeff, let us know about RSConnect.

trestletech:
We don't use Terminate in RSC today.

jjallaire:
Okay, @martinjhnhadley if you could modify the PR to also mention that this only works on shinyapps.io that would be great (we'll take care of the rename / deprecation).

martinjhnhadley:
@jjallaire that's great - thanks for handling the deprecation!

PR #162, opened 2017-03-03
long-polling for connect task requests

aronatkins:
Long-polling was added to Connect in the 1.2.0 release. This change substantially decreases the number of requests made by rsconnect against Connect when looking for new task output.

Hold on merging for now; need to test the behavior against older Connect releases. 

Two open questions:
* Might need to retain the `Sys.sleep` to avoid fast-polling prior to 1.2.0.
* Are we comfortable waiting up to five seconds (or until task completion) to receive the next batch of output? Might need to adjust the Connect long-poll behavior when data is already available.

### DIFF
--- a/R/connect.R
+++ b/R/connect.R
@@ -160,7 +160,7 @@ connectClient <- function(service, authInfo) {
     waitForTask = function(taskId, quiet) {
       start <- 0
       while (TRUE) {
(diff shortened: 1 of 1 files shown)


aronatkins:
Closing this PR for now after determining that we need to change the Connect polling implementation before using it from `rsconnect`.

PR #186, opened 2017-06-13
Manually cast rbind args to factorless data.frame

trestletech:
This actually was covered in our automated tests but I hadn't executed in old versions of R so I didn't catch this :(. The HTTP tests now pass in R3.2.1 (but be aware that some of the bundle tests do still fail in that older version. I can open a separate ticket if that's unexpected).

R 3.2.1 (at least) doesn't support stringsAsFactors in rbind, so you have to create the
factorless data.frame yourself then pass that into rbind if you want to get out of SAF.

Unfortunately our StringsAsFactors call was being interpreted as another row that might
be rbound.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rsconnect
 Type: Package
 Title: Deployment Interface for R Markdown Documents and Shiny Applications
-Version: 0.8.2
+Version: 0.8.3
 Author: JJ Allaire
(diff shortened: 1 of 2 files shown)


trestletech:
Looks like this was introduced in R 3.2.4 https://cran.r-project.org/doc/manuals/r-release/NEWS.html

> The data.frame method of rbind() gains an optional argument stringsAsFactors (instead of only depending on getOption("stringsAsFactors")).

PR #217, opened 2017-11-07
Emit X-Auth-Token HTTP header in every request to RStudio Connect

scottmmjackson:
Connected to #216 

The fake X-Auth-Token is provided to allow anonymous access through a properly configured SSO proxy.

To take advantage of this functionality, the proxy should permit any request that provides `X-Auth-Token` to pass through to Connect.

`deployApp` tested on a local copy of Connect running master branch, proxy auth.


jspiewak:
Will this be sent to shinyapps.io as well?

scottmmjackson:
@jspiewak the header? Yes. I just tested a deployment to shinyapps.io, and it works fine there too.

jspiewak:
I don't think there are any unauthenticated requests to shinyapps.io, so this should be okay.
@kippandrew FYI

PR #259, opened 2018-03-14
only consider the primary doc when computing parameterization

aronatkins:
The primary document is used when launching the rmarkdown::knit_params_ask UI.
The `has_parameters` manifest field indicates if the configuration UI is
permitted.

If the primary document does not contain parameters, rmarkdown::knit_params_ask
exits before even launching its Shiny app. It is better to avoid setting
`has_parameters`.

Non-default documents may be parameterized, but are no longer considered when
marking the bundle as parameterized.

Note: Hold on merging; still need to do a bunch of testing.


aronatkins:
Force-pushed to this branch with latest changes from master.

PR #288, opened 2018-08-30
write publish record to website root when publishing rendered files

gtritchie:
This is in support of fixes here: https://github.com/rstudio/rstudio/pull/3409

When publishing the rendered files for a website, write the rsconnect folder (publish record) to the website root instead of putting it in the _site folder where it is (1) prone to being deleted when rebuilding the site and (2) not found by the IDE when looking for previous publishes.


gtritchie:
Rescinding this PR for the moment. I want to see if I can accomplish this without needing a rev the `rsconnect` package.

PR #289, opened 2018-09-04
Ask RSC for Title Updates

slopp:
This PR introduces a change to `rsconnect::deployments` to account for the fact that in RSC 1.6.8 and above the title can be changed on the Connect end.

This change to `rsconnect` will cause the IDE's publish dialogs to pick up title changes from RSC _each time the content source file is opened_ or _any time content is published_. 

Validated against shinyapps.io to ensure nothing broke.

Also validated the case where the deployment record (rsconnect) folder is deleted, and the case where RSC doesn't respond successfully with the title name (e.g. if the content has been deleted or the publisher's permissions have been changed to viewer).

Small validation note: 

The title change on RSC will not be picked up by the IDE if the source file is left open and the user repeatedly: selects the publish drop down, changes the title on RSC,  reselects the publish drop down _all without ever publishing_.

### DIFF
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -184,7 +184,7 @@ deployments <- function(appPath, nameFilter = NULL, accountFilter = NULL,
     }
 
     # check to see if the title has changed for RSC deploys
-    if (!is.null(server) && server != "rpubs.com" && server != "shinyapps.io")
+    if (length(server) > 0 && server != "rpubs.com" && server != "shinyapps.io")
       deployment$title <- confirmTitle(deployment)
 
     # append to record set to return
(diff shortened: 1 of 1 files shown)


slopp:
@aronatkins and @kevinushey could y'all take a look? (@kevinushey feel free to reassign to a different IDE representative)

slopp:
Following an offline conversation, we've decided to put this on hold until `rstudioapi` supports a public method for invoking async R code. 

At that time, the code will be changed so that instead of modifying the deployment record result set, the `deployments` function will fire an async function to update the deployment record file with the new title ( if file mod time is greater than a certain cache threshold)

PR #323, opened 2018-12-05
Add reticulate python support

mmarchetti:
This PR adds support for python content in apps, Rmds, sites, and APIs that use `reticulate`.

Content that uses `reticulate` can be deployed using `deployApp`, `deployDoc`, `deploySite`, or `deployAPI`. Each of these functions accepts an optional `python` argument which is the full path to a python binary. 

If the `python` argument is provided:
* The python binary is invoked to determine the version and list the installed python packages.
* A `python` section is added to the app manifest.
* A dependency on `reticulate` is inferred.


R/bundle.R
```diff
@@ -449,28 +456,81 @@ inferDependencies <- function(appMode, hasParameters) {
...
 }
 
+inferPythonEnv <- function(workdir, python) {
+  # run the specific python binary, and ensure that it's on the path
+  # so that pip etc. are available.
+  env <- paste0("PATH=", dirname(python), .Platform$path.sep, Sys.getenv("PATH"))
+
+  # run the python introspection script
+  sourceDir <- getSrcDirectory(function() {})
+  args <- c(file.path(sourceDir, "environment.py"), workdir)
```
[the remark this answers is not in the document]
mmarchetti: Excellent feedback, thank you

R/environment.py
```diff
@@ -0,0 +1,139 @@
...
+    Returns a dictionary containing the package spec filename
+    and contents if successful, or a dictionary containing 'error'
+    on failure.
+    """
+    result = (output_file(dirname, 'requirements.txt', 'pip') or
+              pip_freeze(dirname))
+
+    if result is not None:
+        result['python'] = get_python_version()
+        result['pip'] = get_version('pip')
```
jspiewak: Could you do:

```
import pip
pip.__version__
```

mmarchetti: As of pip 10.0, the pip module can no longer be imported (pip issue 5081) but we could probably run `python -m pip --version`.

R/imports.R
```diff
@@ -1,5 +1,5 @@
 
 
 #' @importFrom stats na.omit setNames
-#' @importFrom utils available.packages contrib.url file_test formatUL getFromNamespace glob2rx packageVersion read.csv
+#' @importFrom utils available.packages contrib.url file_test formatUL getFromNamespace getSrcDirectory glob2rx packageVersion read.csv
```
kevinushey: Since we're no longer using `getSrcDirectory()` I think we can get rid of this.

R/bundle.R
```diff
@@ -449,28 +456,77 @@ inferDependencies <- function(appMode, hasParameters) {
...
+  if (!is.null(python)) {
+    deps <- c(deps, "reticulate")
+  }
   unique(deps)
 }
 
+inferPythonEnv <- function(workdir, python) {
+  # run the python introspection script
+  env_py <- system.file("resources/environment.py", package = "rsconnect")
+  args <- c(env_py, workdir)
```
kevinushey: We might need to use `shQuote()` here to ensure arguments are shell-quoted as appropriate. (Unfortunately the `system2()` interface doesn't quote arguments passed in)

mmarchetti:
@kevinushey Are you set to approve, or are there other changes you'd like to see?

R/bundle.R
```diff
@@ -465,7 +465,7 @@ inferDependencies <- function(appMode, hasParameters, python) {
 inferPythonEnv <- function(workdir, python) {
   # run the python introspection script
   env_py <- system.file("resources/environment.py", package = "rsconnect")
-  args <- c(env_py, workdir)
+  args <- c(shQuote(env_py), shQuote(workdir))
```
kevinushey: FYI: `shQuote()` is vectorized so you can just call `shQuote(c(env_py, workdir))`. (The code as written is still correct so no need to change; just want to make sure you're aware of some of the R idioms)

PR #373, opened 2019-08-15
fix issue where package found in multiple repos (#372)

kevinushey:
Part of fix for https://github.com/rstudio/rsconnect/issues/372.


colearendt:
Can you not use the value from the `DESCRIPTION` to make the appropriate lookup? I'm not sure what the right approach is here, but choosing the first one definitely seems a little concerning, when the package said it came from RSPM.

kevinushey:
Based on some Slack discussion, we agree it's okay to take the PR as-is and revisit based on the `Repository` field in the DESCRIPTION if we find this causes some other downstream issue. (This at least unblocks cases where a package in use is found on multiple repositories set in the session)

colearendt:
Just FYI - testing with this branch has been going well so far!!

PR #382, opened 2019-09-19
md5 workaround

aronatkins:
Use `digest::digest` for MD5 file hashes rather than `openssl::md5`. First extracted all MD5 calls into helper functions with tests against those helpers. Once the MD5 calls were encapsulated, shifted from using `openssl` to `digest`.

Workaround until we have a migration path for #363.

Have tested deployments against both RStudio Connect and shinyapps.io.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -209,12 +209,28 @@ activeEncoding <- function(project = getwd()) {
   sub("^Encoding:\\s*", "", encodingLine)
 }
 
-md5sum <- function(path) {
+# Returns the MD5 for path as a sequence of 16 hexadecimal pairs with class md5 hash.
+fileMD5 <- function(path) {
+  if (is.null(path)) {
+    # Use raw(0) rather than the empty string so openssl::md5 returns the has as hex values and not
+    # a concatenated string of hex characters.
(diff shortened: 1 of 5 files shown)


R/utils.R
```diff
@@ -209,12 +209,27 @@ activeEncoding <- function(project = getwd()) {
...
-  on.exit(close(con), add = TRUE)
-
-  # compute md5 sum of contents and return as ordinary characters
-  unclass(as.character(openssl::md5(con)))
+# Returns the MD5 for path as a raw sequence of 16 hexadecimal pairs.
+fileMD5 <- function(path) {
+  if (is.null(path)) {
+    # Use raw(0) rather than the empty string so openssl::md5 returns the has as hex values and not
+    # a concatenated string of hex characters.
+    return(openssl::md5(raw(0)))
```
kevinushey: Does this call still work even in FIPS mode? (Why not just use `digest()` here?)

aronatkins: Oy. That's my mistake. I intended to make that use `digest` as well. PR shortly.

aronatkins: addressed in https://github.com/rstudio/rsconnect/pull/383.

PR #383, opened 2019-09-19
remove remaining openssl::md5 call and add tests forgotten previously

aronatkins:
finishing the work started in https://github.com/rstudio/rsconnect/pull/382

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -211,16 +211,15 @@ activeEncoding <- function(project = getwd()) {
 
 # Returns the MD5 for path as a raw sequence of 16 hexadecimal pairs.
 fileMD5 <- function(path) {
-  if (is.null(path)) {
-    # Use raw(0) rather than the empty string so openssl::md5 returns the has as hex values and not
-    # a concatenated string of hex characters.
-    return(openssl::md5(raw(0)))
-  }
-
   # Use digest::digest to compute file MD5. FIPS mode disables openssl::md5. Workaround until we can
   # migrate away from MD5 for file content checks.
   #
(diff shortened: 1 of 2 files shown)


aronatkins:
Here's a Dockerfile that I used to test this change. It forces FIPS mode and validates that we can still create MD5 hashes of files.

It's a little ugly because it's working around a couple issues, but it does the job!

```dockerfile
FROM centos:centos7

RUN yum -y install epel-release

RUN yum -y install \
        dracut-fips \
        libcurl-devel \
        openssl-devel \
        R-core \
        R-core-devel

# Fix the CentOS R installation to avoid doc copying issues when installing with remotes.
# See https://github.com/r-lib/devtools/issues/2084 for details.
RUN Rscript -e 'x <- file.path(R.home("doc"), "html"); if (!file.exists(x)) {dir.create(x, recursive=TRUE); file.copy(system.file("html/R.css", package="stats"), x)}'

# avoid curl issue on centos/rhel: https://github.com/jeroen/curl/issues/204
RUN R --slave --vanilla -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/curl/curl_4.0.tar.gz",repo=NULL,type="source")'

RUN R --slave --vanilla -e 'install.packages(c("remotes"), repos = "https://cran.rstudio.com/")'
RUN R --slave --vanilla -e 'remotes::install_github("rstudio/rsconnect", upgrade = FALSE, repos = "https://cran.rstudio.com/")'

ENV OPENSSL_FORCE_FIPS_MODE=1
CMD ["R", "--slave", "--vanilla", "-e", "rsconnect:::fileMD5.as.string('/etc/passwd')"]
```

PR #409, opened 2020-03-05
Add reticulate support for conda

scottmmjackson:
Added two new args to `deployApp` (and therefore every `deploy` function that uses `deployApp` and passes args) and `writeManifest`:

- `forceGeneratePythonEnvironment`: write a python environment file even if it exists
- `forceRequirementsTxtEnvironment`: write a `requirements.txt` file even if we're in a conda environment

Updated `environment.py` to the latest version from https://github.com/rstudio/rsconnect-python/blob/2b8c5fe249e254e4b0dfdfa21301241375fbe49b/rsconnect/environment.py

# Review Notes

`environment.py` doesn't need to be reviewed a second time

`paste0` isn't used because it's not supported under our minimum version of R. (is there a tool we can use to give us these hints?)

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -172,7 +172,7 @@ bundleFiles <- function(appDir) {
 
 bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
                       contentCategory, verbose = FALSE, python = NULL,
-                      compatibilityMode = F, forceGenerate = F) {
+                      compatibilityMode = FALSE, forceGenerate = FALSE) {
   logger <- verboseLogger(verbose)
 
   logger("Inferring App mode and parameters")
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -332,8 +332,9 @@ writeManifest <- function(appDir = getwd(),
   manifestPath <- file.path(appDir, "manifest.json")
   writeLines(manifestJson, manifestPath, useBytes = TRUE)
 
-  srcRequirementsFile <- file.path(bundleDir, "requirements.txt")
-  dstRequirementsFile <- file.path(appDir, "requirements.txt")
+  requirementsFilename <- manifest$python$package_manager$package_file
+  srcRequirementsFile <- file.path(bundleDir, requirementsFilename)
+  dstRequirementsFile <- file.path(appDir, requirementsFilename)
   if(file.exists(srcRequirementsFile) && !file.exists(dstRequirementsFile)) {
     file.copy(srcRequirementsFile, dstRequirementsFile)
   }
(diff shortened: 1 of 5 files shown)


R/bundle.R
```diff
@@ -263,12 +266,23 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
...
+#'   argument is `TRUE`.
+#'
 #' @export
 writeManifest <- function(appDir = getwd(),
                           appFiles = NULL,
                           appPrimaryDoc = NULL,
                           contentCategory = NULL,
-                          python = NULL) {
+                          python = NULL,
+                          forceGeneratePythonEnvironment = F,
```
kevinushey:
```suggestion
                          forceGeneratePythonEnvironment = FALSE,
```

R/bundle.R
```diff
@@ -263,12 +266,23 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
...
+                          forceRequirementsTxtEnvironment = F) {
```
kevinushey:
```suggestion
                          forceRequirementsTxtEnvironment = FALSE) {
```

R/bundle.R
```diff
@@ -503,13 +520,50 @@ inferDependencies <- function(appMode, hasParameters, python, hasPythonRmd) {
...
+  args <- c(shQuote(env_py))
+  if (compatibilityMode || forceGenerate) {
+    flags <- paste('-', ifelse(compatibilityMode, 'c', ''), ifelse(forceGenerate, 'f', ''), sep = '')
+    args <- c(args, flags)
+  }
+  args <- c(args, shQuote(workdir))
 
   tryCatch({
-    output <- system2(command = python, args = args, stdout = TRUE, stderr = NULL, wait = TRUE)
+    if(require('reticulate') && reticulate::py_config()$anaconda && !compatibilityMode) {
```
kevinushey:
```suggestion
    if (requireNamespace('reticulate', quietly = TRUE) && reticulate::py_config()$anaconda && !compatibilityMode) {
```

kevinushey: You might also want to check `reticulate::py_available(initialize = FALSE)` as otherwise calling `reticulate::py_config()` will force `reticulate` to initialize Python.

R/deployApp.R
```diff
@@ -109,7 +116,9 @@ deployApp <- function(appDir = getwd(),
                       metadata = list(),
                       forceUpdate = getOption("rsconnect.force.update.apps", FALSE),
                       python = NULL,
-                      on.failure = NULL) {
+                      on.failure = NULL,
+                      forceGeneratePythonEnvironment = F,
```
kevinushey:
```suggestion
                      forceGeneratePythonEnvironment = FALSE,
```

R/deployApp.R
```diff
@@ -109,7 +116,9 @@ deployApp <- function(appDir = getwd(),
...
+                      forceRequirementsTxtEnvironment = F) {
```
kevinushey:
```suggestion
                      forceRequirementsTxtEnvironment = FALSE) {
```

R/bundle.R
scottmmjackson: I'm using `'reticulate' %in% installed.packages()` to avoid including reticulate in the namespace where it doesn't belong.

PR #410, opened 2020-03-09
Fail manifest creation when no repo URL is found for a source.

jskress:
This change updates the `createAppManifest()` function to fail if it discovers that there is no repository URL for the source of a dependent package.  Prior to this change, `null` would be emitted when writing out a `manifest.json` under this condition.  Such a manifest cannot be deployed to RStudio Connect.


scottmmjackson:
Actually, I'm wrong about that- there are tests for `writeManifest`- you could make one!

PR #412, opened 2020-03-11
err when packages have no repository and do not come from an SCM

aronatkins:
This change replaces #410, as it does not err when a package
has an SCM source.

The warning in `snapshotDependencies` has been rewritten, noting
that we cannot determine the repository for the package.

The error in `createAppManifest` suggests that the "repos" option
might need adjusting.

Example output when a specific package cannot be found.

```
Warning: Unable to determine the repository for package ferris
 Error: 
* Unable to determine the location for some packages. Packages must come from a package
   repository like CRAN or a source control system. Check that options('repos') refers to
   a package repository containing the needed package versions. 
```

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -628,7 +628,7 @@ createAppManifest <- function(appDir, appMode, contentCategory, hasParameters,
                                 'Repository')])
 
       if (is.na(info$Repository)) {
-        if (is.scm.source(info$Source)) {
+        if (isSCMSource(info$Source)) {
           # ignore source+SCM packages
         } else {
           missing_url_sources <- unique(c(missing_url_sources, info$Source))
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -628,7 +628,7 @@ createAppManifest <- function(appDir, appMode, contentCategory, hasParameters,
                                 'Repository')])
 
       if (is.na(info$Repository)) {
-        if (tolower(info$Source) %in% c("github", "bitbucket", "source")) {
+        if (is.scm.source(info$Source)) {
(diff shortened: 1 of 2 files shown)


R/bundle.R
```diff
@@ -625,6 +627,14 @@ createAppManifest <- function(appDir, appMode, contentCategory, hasParameters,
       info <- as.list(deps[i, c('Source',
                                 'Repository')])
 
+      if (is.na(info$Repository)) {
+        if (tolower(info$Source) %in% c("github", "bitbucket", "source")) {
```
aronatkins: I don't love this copy-paste of the SCM types; I'll extract this into a helper function.

R/dependencies.R
```diff
@@ -122,15 +122,20 @@ snapshotDependencies <- function(appDir, implicit_dependencies=c()) {
...
+              call. = FALSE, immediate. = TRUE)
     }
     data.frame(Source = source, Repository = repository)
   })
   records[, c("Source","Repository")] <- do.call("rbind", tmp)
   return(records)
 }
 
+# Return TRUE when the source indicates that a package comes from a source control system.
+is.scm.source <- function(source) {
```
colearendt: I'm going to be super nitpicky here, just because it's R and weird. `method.class` is used for S3 dispatch... it seems ( from `random.repo.name` below) that there is some convention for this, but would `is_scm_source` work? I had a heck of a time extricating the dot convention out of a package one time after hitting weird S3 issues and I have never forgotten the pain 😅 Not a big deal to me. Just thought I'd mention it 😄

PR #447, opened 2020-07-28
RSCONNECT_TAR specifies an alternate tar implementation as expected by utils::tar

### LINKED ISSUE #446: utils::tar(tar="internal") can create invalid archives with long user/group names
See https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17871

rsconnect asks for the "internal" tar implementation to be used here:
https://github.com/rstudio/rsconnect/blob/68bf6a7b5292de49ac6e75cfbd1a8c7feedbdc58/R/bundle.R#L238

This was done because some tar implementations create archives with poor file encodings: https://github.com/rstudio/rsconnect/issues/150 & https://github.com/rstudio/rsconnect/pull/165

Unfortunately, the internal implementation is not always appropriate, so we want to create a way to opt-out.

aronatkins:
RSCONNECT_TAR is used when creating bundles for rpubs, shinyapps.io, and
RStudio Connect. When unset, "internal" is used.

Fixes #446

Use an alternate tar implementation by setting this environment variable; use either a full path or a binary in your PATH:

```r
Sys.setenv(RSCONNECT_TAR ="tar")
Sys.setenv(RSCONNECT_TAR ="/usr/bin/tar")
```

You can see what tar is used by deploying with a verbose logging level.

```r
rsconnect::deployApp(..., logLevel = "verbose")
```

When unset, you will see that "internal" is used:

```
[2020-07-28 14:45:07] Using tar: internal 
```

We report the tar executable when the environment variable is set:

```
[2020-07-28 14:45:25] Using tar: tar 
```

```
[2020-07-28 14:53:08] Using tar: /usr/bin/tar 
```

The environment variable applies both to `rsconnect::deploy*` calls from the R console and to deployment operations with output in the Deploy RStudio pane.

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -235,7 +235,9 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
 
   on.exit(setwd(prevDir), add = TRUE)
   bundlePath <- tempfile("rsconnect-bundle", fileext = ".tar.gz")
-  utils::tar(bundlePath, files = ".", compression = "gzip", tar = "internal")
+  tarImplementation <- Sys.getenv("RSCONNECT_TAR", "internal")
+  logger(sprintf("Using tar: %s",tarImplementation))
+  utils::tar(bundlePath, files = ".", compression = "gzip", tar = tarImplementation)
   bundlePath
 }
 
@@ -267,7 +269,7 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
 #'   its value will be used.
 #'
 #' @param forceGeneratePythonEnvironment Optional. If an existing
-#'   `requirements.txt` file is found, it will be overwritten when 
+#'   `requirements.txt` file is found, it will be overwritten when
 #'   this argument is `TRUE`.
 #'
 #'
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -236,7 +236,7 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
   on.exit(setwd(prevDir), add = TRUE)
   bundlePath <- tempfile("rsconnect-bundle", fileext = ".tar.gz")
   tarImplementation <- Sys.getenv("RSCONNECT_TAR", "internal")
-  logger(sprintf("Using tar: %s",tarImplementation))
+  logger(sprintf("Using tar: %s", tarImplementation))
   utils::tar(bundlePath, files = ".", compression = "gzip", tar = tarImplementation)
   bundlePath
 }
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -231,14 +231,21 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
 
   # create the bundle and return its path
   logger("Compressing the bundle")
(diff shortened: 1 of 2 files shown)


R/bundle.R
```diff
@@ -235,7 +235,9 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
 
   on.exit(setwd(prevDir), add = TRUE)
   bundlePath <- tempfile("rsconnect-bundle", fileext = ".tar.gz")
-  utils::tar(bundlePath, files = ".", compression = "gzip", tar = "internal")
+  tarImplementation <- Sys.getenv("RSCONNECT_TAR", "internal")
+  logger(sprintf("Using tar: %s",tarImplementation))
```
kevinushey, line 239:
```suggestion
  logger(sprintf("Using tar: %s", tarImplementation))
```

R/bundle.R
```diff
@@ -235,7 +235,9 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
...
+  utils::tar(bundlePath, files = ".", compression = "gzip", tar = tarImplementation)
```
kevinushey, line 240: The documentation for `?tar` has:

> Note that 'files = '.'' will usually not work with an external
> 'tar' as that would expand the list of files after 'tarfile' is
> created.  (It does work with the default internal method.)

We might need to use `getwd()` explicitly here instead, just to be safe? I'm not sure if we have tests already to verify that this does what we want.

R/rpubs.R
```diff
@@ -90,7 +90,8 @@ rpubsUpload <- function(title,
 
     # create the tarball
     tarfile <- tempfile("package", fileext = ".tar.gz")
-    utils::tar(tarfile, files = ".", compression = "gzip", tar = "internal")
+    tarImplementation <- Sys.getenv("RSCONNECT_TAR", "internal")
+    utils::tar(bundlePath, files = ".", compression = "gzip", tar = tarImplementation)
```
kevinushey, line 94: Do you think it would be worth creating a helper function here (just to share code)?

R/bundle.R
aronatkins: Hm. Maybe just `files = NULL`, given:

```
files | A character vector of filepaths to be archived: the default is to archive all files under the current directory.
```

aronatkins: https://github.com/wch/r-source/blob/8561f62c5d1f464d4a1e491ffc21a42796adf525/src/library/utils/R/tar.R#L352-L359

R/rpubs.R
aronatkins: Sure. I didn't do that initially because the verbosity options were not consistent.

aronatkins: Done, and tested publishing to Connect, shinyapps.io, and rpubs.

R/bundle.R
aronatkins: Using `files = NULL` and tested deploys to all the places.

R/bundle.R
aronatkins: Done. Thanks.

kevinushey:
Also updated NEWS in https://github.com/rstudio/rsconnect/commit/012905d132fc96576a18df1d08eb29cb5c6bfa56.

PR #448, opened 2020-07-29
fixed use of isShinyapps in showUsers

ronomal:
in showUsers changed the paramater for isShinyapps from the full account details to just the server. This was causing a crash whenever calling showUsers


ronomal:
Would you be able to review this @jmcphers? `showUsers()` is currently broken. Thanks.

PR #451, opened 2020-08-20
Allow finding packages with nonstandard contrib URLs

akgold:
For certain packages, there are multiple entries on CRAN. Some of these have nonstandard contrib URLs that don't exactly match to the contrib URLS for any of the configured `repos`, even when configured correctly. 

For example, the `Matrix` package appears twice, once with contrib URL https://colorado.rstudio.com/rspm/all/latest/src/contrib and once with contrib URL https://colorado.rstudio.com/rspm/all/latest/src/contrib/4.1.0/Recommended. If that second URL comes back, exact URL matching silently fails. 

I'm not sure if this is the best way to fix, but it enables manifest writing that was formerly breaking.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -109,7 +109,8 @@ snapshotDependencies <- function(appDir, implicit_dependencies=c()) {
       # Find this package in the set of available packages then use its
       # contrib.url to map back to the configured repositories.
       package.contrib <- repo.packages[pkg, 'Repository']
-      package.repo <- repo.lookup[repo.lookup$contrib.url == package.contrib, ][1, ]
+      package.repo.index <- vapply(repo.lookup$contrib.url, function(url) grepl(url, package.contrib), logical(1))
+      package.repo <- repo.lookup[package.repo.index, ][1, ]
       # If the incoming package comes from CRAN, keep the CRAN name in place
       # even if that means using a different name than the repos list.
       #
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -76,11 +76,19 @@ snapshotDependencies <- function(appDir, implicit_dependencies=c()) {
   # get Bioconductor repos if any
   biocRepos = repos[grep('BioC', names(repos), perl=TRUE, value=TRUE)]
   if (length(biocRepos) > 0) {
-    biocPackages = available.packages(contriburl = contrib.url(biocRepos, type = "source"), type = "source", filters = c("duplicates"))
+    biocPackages = available.packages(
+      contriburl = contrib.url(biocRepos, type = "source"),
+      type = "source",
+      filters = c("R_version", "duplicates")
+      )
   } else {
     biocPackages = c()
   }
-  repo.packages <- available.packages(contriburl = contrib.url(repos, type = "source"), type = "source", filters = c("R_version", "duplicates"))
+  repo.packages <- available.packages(
+    contriburl = contrib.url(repos, type = "source"),
+    type = "source",
+    filters = c("R_version", "duplicates")
+  )
   named.repos <- name.all.repos(repos)
   repo.lookup <- data.frame(
     name = names(named.repos),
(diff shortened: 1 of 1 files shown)


akgold:
Not really sure who would want to review, so tagged a few people...

slopp:
I'd have Aron review this when he returns

R/dependencies.R
```diff
@@ -80,7 +80,7 @@ snapshotDependencies <- function(appDir, implicit_dependencies=c()) {
   } else {
```
akgold, line 80: `available.packages` is also called in line 79 for `biocRepos`, do we need to filter on `R_version` there too?

akgold:
I'm in no rush since I can install from the branch, happy to wait for Aron's return.

R/dependencies.R
```diff
@@ -80,7 +80,7 @@ snapshotDependencies <- function(appDir, implicit_dependencies=c()) {
...
     biocPackages = c()
   }
-  repo.packages <- available.packages(contriburl = contrib.url(repos, type = "source"), type = "source", filters = c("duplicates"))
+  repo.packages <- available.packages(contriburl = contrib.url(repos, type = "source"), type = "source", filters = c("R_version", "duplicates"))
```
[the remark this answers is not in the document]
slopp: I think it is still possible for someone using a devel version of R plus RSPM (due to the differing order in PACKAGES that prefers the pathed contrib.url) to hit this, so I'd recommend including both fixes. 

The addition of the R version filter is an oversight that should have been in place to begin with, but only practically matters for the case when two packages with the same version exist but with different R version constraints.

slopp:
@aronatkins This was a bit of a perfect storm scenario. Happy to chat through the context if you want, but I think the fixes are appropriate as long as they look safe to you!

R/dependencies.R
```diff
@@ -109,7 +109,8 @@ snapshotDependencies <- function(appDir, implicit_dependencies=c()) {
       # Find this package in the set of available packages then use its
       # contrib.url to map back to the configured repositories.
       package.contrib <- repo.packages[pkg, 'Repository']
-      package.repo <- repo.lookup[repo.lookup$contrib.url == package.contrib, ][1, ]
+      package.repo.index <- vapply(repo.lookup$contrib.url, function(url) grepl(url, package.contrib), logical(1))
```
kevinushey, line 112: Should we be using `grepl(..., fixed = TRUE)` here?

trestletech:
Just dropping my +1. I just hit this and lost a couple of hours to it before I found this PR.

R 3.6 on Ubuntu 18 with RSPM. It blew up on the matrix package having an `NA` repository. For future searchers I just got:

```
2020-08-26 21:05:55  Error: 
2020-08-26 21:05:55  * Unable to determine the location for some packages. Packages must
2020-08-26 21:05:55     come from a package repository like CRAN or a source control system.
2020-08-26 21:05:55     Check that options('repos') refers to a package repository
2020-08-26 21:05:55     containing the needed package versions.
```

with no helpful warnings pointing me to the specific package, but I'm in a non-interactive session so maybe I'm an outlier.

R/dependencies.R
akgold: Yes, changed.

akgold:
Hi @aronatkins -- could you take a look here? Thanks!

R/dependencies.R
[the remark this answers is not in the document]
aronatkins: We may eventually find that "duplicates" is hiding the correct repository (since we are trying to discover where a particular version is stored).

aronatkins (approved):

R/dependencies.R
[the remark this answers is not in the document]
akgold: Just to be clear -- this is a possible flag for the future, but not something you're suggesting changing now, right?

R/dependencies.R
aronatkins: R version filtering feels appropriate any time we might encounter repositories that do not contain the entire release history for some package. That probably includes the BioC repositories.

R/dependencies.R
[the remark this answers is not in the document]
aronatkins: Correct. Partial repositories are trouble, but let's wait until we see additional examples.

R/dependencies.R
akgold: Done.

akgold:
I'm not sure what standard is for this repo -- should I merge? (maybe @aronatkins?)

PR #461, opened 2020-10-16
Shiny Rmd takes precedence over server.R

aronatkins:
Requires changes from https://github.com/rstudio/rmarkdown/pull/1926 to test deploys.

```console
R -e 'remotes::install_github("rstudio/rmarkdown", ref="feature/shinyrmd-server-r")'
```

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -444,9 +444,21 @@ inferAppMode <- function(appDir, appPrimaryDoc, files) {
     return("shiny")
   }
 
-  # shiny directory
-  shinyFiles <- grep("^(server|app).r$", files, ignore.case = TRUE, perl = TRUE)
-  if (length(shinyFiles) > 0) {
+  # Shiny application or ShinyRmd using server.R
+  serverR <- grep("^server.r$", files, ignore.case = TRUE, perl = TRUE)
+  if (length(serverR) > 0) {
+    # Shiny server.R and UI.Rmd
+    uiRmd <- grep("^ui.rmd$", files, ignore.case = TRUE, perl = TRUE)
+    if (length(uiRmd) > 0) {
+      return("rmd-shiny")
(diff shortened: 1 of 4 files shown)


R/bundle.R
```diff
@@ -444,9 +444,21 @@ inferAppMode <- function(appDir, appPrimaryDoc, files) {
     return("shiny")
   }
 
-  # shiny directory
-  shinyFiles <- grep("^(server|app).r$", files, ignore.case = TRUE, perl = TRUE)
-  if (length(shinyFiles) > 0) {
+  # Shiny application or ShinyRmd using server.R
```
jjallaire, line 447: It might be a bit more complicated that that. ui.Rmd is really just a convention (as is index.Rmd). You could also have foo.Rmd with a server.R. The real heuristic for rmd-shiny is that there is at least one .Rmd file in the directory with `runtime: shiny_prerendered` or `runtime: shinyrmd`. If that condition is met it's an "rmd-shiny" (no matter whether it has a server.R and no matter what the name of the .Rmd file containing the requisite `runtime` is).

aronatkins: Ok. Thanks. This is helpful. That suggests a different precedence than we have today, which prefers running the directory as a Shiny application. Will need to do a little more digging. Looking at the runtime field is easily done here. The heuristic in Connect doesn't read the YAML today.

aronatkins:
Still need more testing, but the proposed precedence rules:

1. Plumber APIs via plumber.R or endpoint.R
2. Shiny application via app.R
3. Shiny Rmd (Rmd with shiny* runtime); optionally paired with a server.R
4. Shiny application with server.R
5. Static (rendered) Rmd
6. TF model
7. Static content

christierney (commented):
👍🏻

PR #462, opened 2020-10-16
trim activation from the renv profile

### LINKED ISSUE #439: Issue with renv
Currently, when `rsconnect` builds a tar file to upload content to Connect, it inspects the `.Rprofile` file (if present) for the presence of `packrat`.  If it's there, it comments it out so that it does not conflict with Connect's use of `packrat`.

This same process needs to be done for `renv`.  If `renv` is sourced in `.Rprofile` then you'll end up with competing package builds (Connect's use of `packrat`) and the content's use of `renv`.

The code should look for:

```r
source("renv/activate.R")
```

in `.Rprofile` and comment it out if found.

`packrat` is handled here: https://github.com/rstudio/rsconnect/blob/40c162bf3c4cd9b10deed82e1df9ec1b0c107c1f/R/bundle.R#L29-L39

aronatkins:
fixes #439

Rewrites `.Rprofile` during deployment.

```
# Modified by rsconnect package 0.8.16.9002 on 2020-10-16 16:40:44:
# renv initialization disabled in published application
# source("renv/activate.R")
```

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -25,16 +25,24 @@ bundleAppDir <- function(appDir, appFiles, appPrimaryDoc = NULL, verbose = FALSE
       dir.create(dirname(to), recursive = TRUE)
     file.copy(from, to)
 
-    # ensure .Rprofile doesn't call packrat/init.R
+    # ensure .Rprofile doesn't call packrat/init.R or renv/activate.R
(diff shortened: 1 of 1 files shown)


kevinushey (approved):
LGTM!

PR #467, opened 2020-11-20
don't use 'R_version' filter when querying available packages

kevinushey:
This PR fixes an issue where `rsconnect::writeManfiest()` could fail if:

1. The current version of R is older (e.g. R 3.5), but
2. The application uses a package whose current version on CRAN requires a new version.

An example of this is the `foreign` package; it currently requires (R >= 4.0.0). This implies that manifest generation will fail, with an error of the form:

```
> rsconnect::writeManifest(appDir = "usesforeign/")
Warning: Unable to determine the repository for package foreign
Error: 
* Unable to determine the location for some packages. Packages must come from a package
   repository like CRAN or a source control system. Check that options('repos') refers to a
   package repository containing the needed package versions.
```

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -73,22 +73,36 @@ snapshotDependencies <- function(appDir, implicit_dependencies=c()) {
     sapply(repos, "[[", 1)
   )
 
+  # read available.packages filters (allow user to override if necessary;
+  # this is primarily to allow debugging)
+  #
+  # note that we explicitly exclude the 'R_version' filter as we want to ensure
+  # that packages which require newer versions of R than the one currently
+  # in use can still be marked as available on CRAN -- for example, currently
+  # the package 'foreign' requires 'R (>= 4.0.0)' but older versions of R
+  # can still successfully install older versions from the CRAN archive
+  filters <- getOption(
+    "available_packages_filters",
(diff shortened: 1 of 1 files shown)


kevinushey:
Adding @akgold + @aronatkins just to double-check whether this was intentional or not.

aronatkins:
@kevinushey - I think https://github.com/rstudio/rsconnect/pull/451 by @akgold was trying to find a "better" match when different repos had different package versions available. Unfortunately, that probably causes us to miss things when the latest package version (and the only one defined) publicizes needing a different R version.

I'd like to hear @akgold's take.

akgold:
That's right -- in #451 the issue was that in line 112, `package.contrib` could have 2 entries (e.g. `Matrix` package), with the match for `repo.lookup$contrib.url` in the second and therefore not found in line 112.

Filtering by `R_version` was a suggested addition during the review process. It may be the case that `R_version` isn't necessary combined with changes for 112.

jmcphers (approved):
LGTM!

aronatkins:
I've confirmed that a Shiny application using `foreign` and R 3.6.x cannot compute its manifest without this change. With this change, it calculates the correct dependencies (successfully deploying to RSC).

PR #472, opened 2020-12-14
exclude packrat/* files from the manifest.json file listing

aronatkins:
When bundling for writeManifest, remove the packrat tree before building
the set of files. Useful for writeManifest callers, who do not the
packrat directory we create while performing a bundle snapshot.

The manifest.json lacks packrat entries _only_ when creating a
standalone manifest, but in the future, we could suppress the packrat
entries when deploying to shinyapps.io, as well.


meatballhat (approved):
:tada:

PR #475, opened 2020-12-18
Use pip list --format-freeze instead of pip freeze

mmarchetti:
### Description

In conda environments, `pip freeze` lists paths in the requirements.txt file which do not exist locally. This is caused by the PEP 610 metadata present for conda-installed packages, and is exposed by a change in pip 20.1b1.

### Testing Notes / Validation Steps

Publish from virtualenvs and conda envs, without an existing requirements.txt file.


scottmmjackson (approved):

inst/resources/environment.py
```diff
@@ -59,46 +100,59 @@ def get_conda(conda=None):
...
     v = sys.version_info
     return "%d.%d.%d" % (v[0], v[1], v[2])
 
 
 def get_conda_version(conda):
     try:
-        args = [conda, '-V']
-        proc = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
+        args = [conda, "-V"]
+        proc = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True,)
```
kevinushey, line 126: Is the trailing comma here intentional?

inst/resources/environment.py
```diff
@@ -141,28 +194,32 @@ def pip_freeze():
     """
     try:
         proc = subprocess.Popen(
-            [sys.executable, '-m', 'pip', 'freeze'],
-            stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
+            [sys.executable, "-m", "pip", "list", "--format=freeze"],
```
kevinushey, line 197: Do we know when the `--format=freeze` argument was added? Do we need to worry about older versions of `pip` that don't support this flag?

kevinushey (commented):
Overall looks good to me, but I wonder if we need to handle the case where the user has an older version of `pip` that doesn't support the `--format=freeze` argument.

inst/resources/environment.py
mmarchetti: No, I'm guess this was multiline formatted (with the comma) and `black` decided to put it on one line.

inst/resources/environment.py
mmarchetti: It is present in pip 9.0 which was released in November 2016. We're probably OK?

kevinushey (approved):

kevinushey:
LGTM; thanks!

PR #476, opened 2021-01-11
Increase the default rsconnect timeout

MariaSemple:
If 5 seconds isn't enough, users can get stuck and it's not straightforward for someone who isn't familiar with the internal workings of `rsconnect` how to fix the issue.


kevinushey (approved):
LGTM!

PR #477, opened 2021-01-27
warn on long user/group file ownership

aronatkins:
The R tar implementation does not produce valid archives when file ownership uses long user/group names. Emit a warning about this problem. It appears that all R through 4.0.4 have this issue.

Follow-up to earlier work on #446, where we added support for an alternate tar in #447

Here is some example output when I force a low length limit and run it against the `rsconnect` repository.

```
rsconnect:::detectLongNames(".", lengthLimit = 3)
Warning message:
In rsconnect:::detectLongNames(".", lengthLimit = 3) :
  The bundle contains files with user/group names having more than 3 characters: _pkgdown.yml is owned by aron:staff. Long user and group names cause the internal R tar implementation to produce invalid archives. Set the RSCONNECT_TAR environment variable to use an external tar command.
```

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -270,8 +270,36 @@ writeBundle <- function(bundleDir, bundlePath, verbose = FALSE) {
 
   tarImplementation <- Sys.getenv("RSCONNECT_TAR", "internal")
   logger(sprintf("Using tar: %s", tarImplementation))
+
+  if (tarImplementation == "internal") {
+    detectLongNames(bundleDir)
+  }
+
   utils::tar(bundlePath, files = NULL, compression = "gzip", tar = tarImplementation)
 }
+
+# Scan the bundle directory looking for long user/group names.
+#
+# Warn that the internal tar implementation may produce invalid archives.
+# https://github.com/rstudio/rsconnect/issues/446
+# https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17871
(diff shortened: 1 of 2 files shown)


R/bundle.R
```diff
@@ -270,9 +270,37 @@ writeBundle <- function(bundleDir, bundlePath, verbose = FALSE) {
...
+#
+# Warn that the internal tar implementation may produce invalid archives.
+# https://github.com/rstudio/rsconnect/issues/446
+# https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17871
+detectLongNames <- function(bundleDir, lengthLimit = 32) {
+  files <- list.files(bundleDir, recursive = TRUE, all.files = TRUE,
+                      include.dirs = TRUE, no.. = TRUE, full.names = FALSE)
+  for (f in files) {
+    info <- file.info(file.path(bundleDir,f))
+    cat(sprintf("f=%s; u=%s (%d); g=%s, (%d)\n", f, info$uname, nchar(info$uname), info$grname, nchar(info$grname)))
```
kevinushey, line 291: Do we always want to emit this line, or only if we detect a long user / group name?

aronatkins: No. This was a mistake. It's gone.

kevinushey (approved):
LGTM!

aronatkins:
CC @scottmmjackson fyi.

PR #490, opened 2021-03-23
fix "Error in readLines(contentFile) : 'con' is not a connection" error when JSON tracing enabled

kevinushey:
The code was erroneously checking the null-ity of `file` rather than `contentFile`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ## 0.8.17 (in development)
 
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
LGTM, want to add a NEWS entry or do you think this is too minor to warrant one?

kevinushey:
Thanks for the nudge; I've updated the NEWS file as well.

PR #492, opened 2021-03-25
GitLab is a valid SCM

aronatkins:


aronatkins:
Abandoning in favor of #493.

PR #496, opened 2021-03-29
size is missing/null in some applications entries

### LINKED ISSUE #495: rsconnect::applications() fails because shinyapps API may not return consistent output format
I called rsconnect::applications() and got an error:
```
Error in rbind.data.frame(..., stringsAsFactors = FALSE) : 
  numbers of columns of arguments do not match
```

The error is triggered at https://github.com/rstudio/rsconnect/blob/master/R/applications.R#L111 because the data coming back from /applications/ endpoint does not include the `size` attribute. This may be symptomatic of a more general problem and it may be an issue with the API rather than rsconnect but wanted to mention the issue regardless for someone to take further. 

Here is the data structure parsed from the endpoint for my application (I can share a non-redacted version with an rstudio employee if needed):

```
$id
[1] {REDACTED}

$name
[1] "kelchVis"

$url
[1] "https://{REDACTED}.shinyapps.io/kelchVis/"

$status
[1] "pending"

$created_time
[1] "2020-10-20T07:04:25"

$updated_time
[1] "2020-10-20T07:04:25"

$instances
[1] NA

$guid
[1] NA```

aronatkins:
fixes #495

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,7 @@
 * Require `packrat` 0.5 or later (#434)
 * Fix error when handling empty application / content lists (#417, #395)
 * Calls to `writeManifest()` no longer reference `packrat` files in the generated `manifest.json`. The `packrat` entries were transient and only existed while computing dependencies. (#472)
+* Fix `applications` when ShinyApps does not return `size` details (#496)
 
 ## 0.8.16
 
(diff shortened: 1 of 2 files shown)


scottmmjackson (approved):

jfunction:
Confirmed working. Thanks!

PR #504, opened 2021-04-16
lint different-case markdown links

### LINKED ISSUE #388: linting: identify different-case filenames on case-preserving, case-insensitive systems
Would it be possible to offer feedback when using a case-preserving, case-insensitive file system and referencing on-disk files using different-case names? For example, we may have an on-disk file named `images/RStudio.svg` and then use it from an R Markdown document with:

```markdown
![](images/rstudio.svg)
```

This renders fine on a macOS system, for example, but will probably err if that same content is rendered on a Linux system. The image reference is non-portable.

File-system differences will cause problems when migrating code from RStudio desktop to RStudio server, for example, or when deploying code to shinyapps.io or RStudio Connect.

Could we attempt to identify these situations?

Note: We sometimes avoid this problem when deploying content, as the filename for detected dependencies uses the case from the Rmd source - and that case is preserved in the deployment bundle.
image

The on-disk case is used when dependencies are manually included in a deployment.

aronatkins:
Identifies text within ](...) and check for case problems against files in the project. Covers links and images.

fixes #388

Tested with:

```
rsconnect::deployApp(appDir = getwd(), appFiles = c("index.Rmd", "RStudio.svg"), ...)
```

This gives the lint complaint:

image

Note: When an Rmd references the same image with two different names, `rsconnect::deployDoc()` and the IDE publish button attempt to self-resolve these case problems and create a bundle having both names:

image

The actual bundle only contains one of those two filenames, which still causes server-side rendering problems.

### DIFF
--- a/R/linters.R
+++ b/R/linters.R
@@ -41,10 +41,8 @@ addLinter("filepath.capitalization", linter(
 
     content <- stripComments(content)
 
-    # Inferred files within source documents (really, we just
-    # extract everything between two quotes)
-    regexes <- c("(?!\\\\)\'", "(?!\\\\)\"")
-    inferredFiles <- lapply(regexes, function(regex) {
+    # Extract references between bounding (unescaped) quotes.
+    extractQuoted <- function(regex) {
       matches <- gregexpr(regex, content, perl = TRUE)
       results <- vector("list", length(content))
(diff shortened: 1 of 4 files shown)


jmcphers (approved):
LGTM!

PR #540, opened 2021-09-23
preserve file timestamps when copying files into bundle staging

aronatkins:
This change makes preserves the timestamps from the source files when copying
into the temporary bundle staging directory. As a result, the tar.gz archive
will use those timestamps rather than the timestamps from the time the copy was
performed.

Having accurate timestamps within a bundle makes it possible for deployments to
make decisions based on the staleness of the deployed files. A pre-rendered R
Markdown document, for example, will have a newer timestamp when the included
HTML is stale. Prior to this change, the HTML and Rmd would almost always be
written in the same second and rmarkdown::run is unable to detect staleness.


aronatkins:
Should we consider local timestamps as trustworthy? They could be wildly wrong, but are the timestamps within a bundle likely to be internally consistent? Is that helpful? It may make it less likely to see out-of-date HTML for pre-rendered content.

In my testing, I have found that this change lets accurate timestamps make their way to shinyapps.io deployments, but RStudio Connect drops incoming timestamps. RSC would need a similar change.

CC @kippandrew

PR #544, opened 2021-10-15
Change curl to use -T -X POST instead of data-binary.

telnoratti:
When uploading a large bundle (many GB) to RStudio Connect we have to set `options(rsconnect.http = "curl")` to use the curl uploader. The code currently uses `--data-binary` which tells curl to load the entire file into memory at once before uploading. Obviously this can cause problems if the size of the bundle is larger than memory, so instead we should use the `-T <file>` and `-X POST` options.

I haven't done extensive testing, but from my reading of the curl man page these options should be equivalent to `--data-binary`. As far as portability goes, both options exist on curl as old as 7.15.5.

### DIFF
--- a/R/http-curl.R
+++ b/R/http-curl.R
@@ -45,8 +45,9 @@ httpCurl <- function(protocol,
 
   if (!is.null(contentFile)) {
     command <- paste(command,
-                     "--data-binary",
-                     shQuote(paste("@", contentFile, sep="")),
+                     "-T",
+                     shQuote(contentFile),
+                     "-X", "POST",
                      "--header", paste('"' ,"Content-Type: ",contentType, '"', sep=""),
                      "--header", paste('"', "Content-Length: ", fileLength, '"', sep=""))
   }


aronatkins:
@telnoratti - would you mind adding an entry to NEWS.md?
@jmcphers - any concerns taking this change?

jmcphers (approved):

aronatkins:
Thanks @telnoratti. I'm going to coordinate some testing of this change and will get it merged shortly.

telnoratti:
@aronatkins Great! Let me know if there's anything I can do to help with testing. If you'd like to test bundles larger than memory size you'll probably also need to set the `rsconnect.max.bundle.size`.

On a related note we've also patched our version of the code to use `compression_level = 1` in the `utils::tar` function, but that's because we're uploading pre-compressed data. Our bundles can be up to 250GB in size and using higher compression levels can take a very long time. It might be worth to allow setting that as an option, but I'll defer to your opinion there. I'm not sure how generally useful it would be.

PR #548, opened 2021-11-08
let packrat use renv for dependency analysis

aronatkins:
requires https://github.com/rstudio/packrat/pull/647

Automatically enables the `packrat.dependency.discovery.renv` option, requesting that packrat use renv for dependency analysis. An existing value for `packrat.dependency.discovery.renv` is left alone.


kevinushey (approved):
LGTM!

aronatkins:
@kevinushey Any concerns with including this in rsconnect ahead of a possible packrat release?

kevinushey (approved):
LGTM!

PR #554, opened 2021-11-16
hook-up GH actions

aronatkins:

PR #555, opened 2021-11-16
hook-up GH actions (do not merge)

aronatkins:
Branch with changes from #554 but before #551 landed. Should help us prove that the GH actions CI would have seen the CRAN failure.


aronatkins:
The experimentation in this PR indicates that we are not retaining files beneath `~/Library/Application Support/R` after our `rsconnect` test runs. It is possible that CRAN noticed the files left behind by old package versions, or as a side-effect of tests by consumers of the rsconnect package.

hadley:
Can this be closed?

aronatkins:
Experiment over.

PR #559, opened 2021-11-22
New shinyapps deployments can be immediately private

### LINKED ISSUE #403: No option to set visibility (private/public) when deploying to shinyapps.io
I think it would be really useful to have an option in rsconnect::deployApp() that enabled setting of public/private visibility. At the moment publishing to shinyapps.io defaults to public and if you want it to be private you have to then remember to set it in the dashboard manually afterwards while your application is already open to the public. Found 2015 post on the subject by @tareefk [here](https://groups.google.com/forum/#!topic/shinyapps-users/--CE88p_IxU) (shinyapps.io google group).

mbaynton:
Fixes #403

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -115,7 +115,7 @@ deployApp <- function(appDir = getwd(),
                       python = NULL,
                       on.failure = NULL,
                       forceGeneratePythonEnvironment = FALSE,
-                      shinyAppsVisibility = "public"
+                      appVisibility= "public"
                       ) {
 
   condaMode <- FALSE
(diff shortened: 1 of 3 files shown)


R/deployApp.R
```diff
@@ -351,6 +353,20 @@ deployApp <- function(appDir = getwd(),
     application <- applicationForTarget(client, accountDetails, target, forceUpdate)
   })
 
+  # Shinyapps defaults to public visibility.
+  # Other values should be set before data is deployed.
+  currentVisibility <- application$deployment$properties$application.visibility
+  if (identical(currentVisibility, NULL)) {
```
aronatkins, line 359: I think you want `is.null(currentVisibility)`

R/deployApp.R
```diff
@@ -351,6 +353,20 @@ deployApp <- function(appDir = getwd(),
     application <- applicationForTarget(client, accountDetails, target, forceUpdate)
   })
 
+  # Shinyapps defaults to public visibility.
+  # Other values should be set before data is deployed.
+  currentVisibility <- application$deployment$properties$application.visibility
```
aronatkins, line 358: Move computation of `currentVisibility` into your `isShinyapps` block. Only shinyapps has that data structure.

aronatkins (commented):
Please update NEWS.md as well.

R/deployApp.R
```diff
@@ -114,7 +114,9 @@ deployApp <- function(appDir = getwd(),
                       forceUpdate = getOption("rsconnect.force.update.apps", FALSE),
                       python = NULL,
                       on.failure = NULL,
-                      forceGeneratePythonEnvironment = FALSE) {
+                      forceGeneratePythonEnvironment = FALSE,
+                      shinyAppsVisibility = "public"
```
kippandrew, line 118: Seems a bit odd to include `shinyapps` in the parameter name. What about `appVisibility`? Obviously, this is a shinyapps.io specific flag, but we've never explicitly called out RSC specific flags.

R/deployApp.R
```diff
@@ -68,6 +68,9 @@
 #' @param forceGeneratePythonEnvironment Optional. If an existing
 #'   `requirements.txt` file is found, it will be overwritten when this argument
 #'   is `TRUE`.
+#' @param appVisibility One of `"private"`, `"public"`, or `"hidden"`; indicates
```
kippandrew, line 71: Let's omit the "hidden" parameter here, it's not relevant for shinyapps.io.

kippandrew (approved):

aronatkins (approved):

PR #564, opened 2022-01-27
Add a SHA-256 checksum to requests via the HTTP Digest header

atheriel:
MD5 checksums are not all that secure, and cause problems on FIPS-compliant systems in particular, so we'd ideally move to SHA-256. This is tracked already in #363.

In order to someday deprecate the use of MD5 while also continuing to work with older versions of Connect that may require it, we have two broad options:

1. Determine whether the upstream server supports another checksum mechanism, likely by checking its version; or

2. Keep the existing checksum support in place but send an additional non-MD5 checksum for use by servers that support it.

This commit implements the second approach: we send an additional header with a SHA-256 checksum in addition to the existing MD5 one. Rather than use another custom header, I've opted to use the `Digest` header specified in RFC 3230, which is consistent with the semantics we're after anyway.

Unit tests are included but I have not done any end-to-end testing, hence the draft status.

### DIFF
--- a/R/http.R
+++ b/R/http.R
@@ -409,6 +409,7 @@ signatureHeaders <- function(authInfo, method, path, file) {
   headers$`X-Auth-Token` <- authInfo$token
   headers$`X-Auth-Signature` <- signature
   headers$`X-Content-Checksum` <- md5
+  headers$Digest <- fileDigestHeader(file)
   headers
 }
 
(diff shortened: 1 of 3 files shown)


hadley:
Given that `signatureHeaders()` mostly exists to work around the absence of https support in very old versions of R, maybe we can gradually move to ditch this auth mechanism altogether?

PR #567, opened 2022-03-03
Bump the minimum `openssl` package version in the DESCRIPTION

### LINKED ISSUE #452: Implement alternative to OpenSSL RSA keygen
A customer has an issue with this code:
https://github.com/rstudio/rsconnect/blob/1306eed851a01506a81a3f80bf7a92106f7e5a8b/R/rsa.R#L39-L42

The call to `openssl::rsa_keygen` is producing a FIPS error:

```
> rsconnect::connectUser(server='mylocaldeployserver')
Error: OpenSSL error in EVP_DigestInit_ex: disabled for fips
```

This issue is closely related to #378, which hit FIPS errors while using MD5.

(Internal: See support ticket 51376).

atheriel:
The 2.0.0 release contains a change that fixes publishing on FIPS systems, so requiring this minimum version should insulate users from this issue on new installs. See #452 for discussion.

Fixes #452 and #489.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,7 +17,7 @@ Imports:
     curl,
     digest,
     jsonlite,
-    openssl,
+    openssl (>= 2.0.0),
     packrat (>= 0.6),
     rstudioapi (>= 0.5),
     tools,
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 ## 0.8.26 (in development)
 
 * Allow ShinyApps deployments to be private at creation time (#403)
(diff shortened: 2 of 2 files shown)


aronatkins (approved):
Should this change be announced in NEWS?

atheriel:
@aronatkins I'm on the fence about whether it's appropriate. But I'm happy to add an entry if you think it makes sense.

aronatkins:
I'm leaning towards mentioning this in NEWS just because it avoids a particular class of FIPS issues that a number of folks have hit; it's more about the consequential change in behavior rather than the required version update.

atheriel:
Makes sense. Added a NEWS entry.

PR #569, opened 2022-03-10
Check RETICULATE_PYTHON_FALLBACK and use in manifest

jgutman:
....when `RETICULATE_PYTHON` unset

Addresses [rstudio#10736](https://github.com/rstudio/rstudio/issues/10736) and [connect#20879](https://github.com/rstudio/connect/issues/20879)

Tested by deploying site after rebuilding `rsconnect` package.

Deploy logs show: 
```
[Connect] 2022/03/10 22:00:10.777366461 [rsc-quarto] Using Python binary: /opt/Python/3.9.7/bin/python3.9
[Connect] 2022/03/10 22:00:10.777405223 [rsc-quarto] Configuring PATH from /opt/Python/3.9.7/bin/python3.9: /opt/Python/3.9.7/bin:/opt/R/3.6.3/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
[Connect] 2022/03/10 22:00:10.777411676 [rsc-quarto] Using bundle Python binary: /opt/rstudio-connect/mnt/app/python/env/bin/python
[Connect] 2022/03/10 22:00:10.777779626 [rsc-quarto] Configuring PATH from /opt/rstudio-connect/mnt/app/python/env/bin/python: /opt/rstudio-connect/mnt/app/python/env/bin:/opt/Python/3.9.7/bin:/opt/R/3.6.3/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
```

And manifest.json now includes: 
```
"python": {
    "version": "3.9.10",
    "package_manager": {
      "name": "pip",
      "version": "21.3.1",
      "package_file": "requirements.txt"
    }
  },
  ```

PR #570, opened 2022-03-10
When getting Python executable, check `RETICULATE_PYTHON_FALLBACK` if `RETICULATE_PYTHON` is empty

toph-allen:
### Intent

This resolves the IDE issue https://github.com/rstudio/rstudio/issues/10736.

A change to the IDE in [src/cpp/session/modules/SessionRSConnect.cpp](https://github.com/rstudio/rstudio/blob/e1e8a2dd70868a625348355e634631414795efa9/src/cpp/session/modules/SessionRSConnect.cpp) changed the state of the environment that `rsconnect::deployApp()` runs in during push-button deployment to Connect, if `RETICULATE_PYTHON` is not set in the user's environment.

As I understand it: Previously, whatever Python was discovered would be set to `RETICULATE_PYTHON` when that function ran during push-button deployment. The recent change meant that if `RETICULATE_PYTHON` was not set by the user, the executable path would be located in the `RETICULATE_PYTHON_FALLBACK` variable instead.

The `rsconnect:::getPython()` function would only check `RETICULATE_PYTHON`, which led to deployments with manifests that didn't contain a `python` object. This resulted in different outcomes in different environments. In my local dev environment, deployment failed. On beta.rstudioconnect.com, code would run using the wrong Python.

### Approach

`rsconnect:::getPython()` will now check `RETICULATE_PYTHON_FALLBACK` if `RETICULATE_PYTHON` is empty. If `RETICULATE_PYTHON_FALLBACK` is also empty, the function returns NULL, as before.

I added one new test and modified another to account for this new behavior.

With this change, the reproducible example I attached to the IDE issue succeeds — the `manifest.json` contains the expected Python info.

### Notes

I also committed a change to `rsconnect.Rproj`. Whenever I hit "Install" in the IDE's Build tab, the IDE will modify the Rproj file to include the state of the "Clean before install" checkbox in the "Build Tools" pane of Project Options. It seems to default to "Yes", so that's what I recorded.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -501,6 +501,9 @@ getPython <- function(path) {
   if (is.null(path)) {
     path <- Sys.getenv("RETICULATE_PYTHON")
     if (path == "") {
+      path <- Sys.getenv("RETICULATE_PYTHON_FALLBACK")
+    }
+    if (path == "") {
       return(NULL)
     }
   }
--- a/rsconnect.Rproj
+++ b/rsconnect.Rproj
@@ -17,6 +17,7 @@ StripTrailingWhitespace: Yes
 
 BuildType: Package
 PackageUseDevtools: Yes
+PackageCleanBeforeInstall: Yes
 PackageInstallArgs: --no-multiarch --with-keep.source
 PackageCheckArgs: --as-cran
 PackageRoxygenize: rd,collate,namespace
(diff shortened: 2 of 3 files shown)


toph-allen:
Oops, I definitely didn't see @jgutman’s PR (https://github.com/rstudio/rsconnect/pull/569) before I wrote this up 🤦🏻

toph-allen:
@jgutman also filed a PR solving this that I didn't see earlier: https://github.com/rstudio/rsconnect/pull/569

PR #581, opened 2022-05-02
[shinyapps.io] Add ability to resend invitations 

### LINKED ISSUE #543: Get pending invitation links from the library
We can currently invite users to an app using the the library, but, AFAIK, it is not possible to retrieve the invitation link after that. In case a user is already invited, we are forced to go to the web console manually to get the invitation link.

It would be REALLY appreciated to be allowed to retrieve invitation links from the library.

kippandrew:
Resolves #543 

This PR adds a new function `resendInvitation` that will resend an invitation email (and optionally regenerate the invitation code). An invites expiration status has also been added to the dataframe returned by `showInvited`.


R/auth.R
```diff
@@ -220,6 +221,56 @@ showInvited <- function(appDir=getwd(), appName=NULL, account = NULL,
...
+                             appDir=getwd(), appName=NULL,
+                             account = NULL, server=NULL) {
+
+  # resolve account
+  accountDetails <- accountInfo(resolveAccount(account, server), server)
+
+  # get invitations
+  invited <- showInvited(appDir, appName, account, server)
+
+  if (is.numeric(invite)) {
```
aronatkins, line 249: Is there any point to handling a vector of invitations?

aronatkins (approved):

R/auth.R
kippandrew: Potentially useful, but it would be inconsistent with the other APIs. 🤷‍♂️

aronatkins: kk. as-is it is, then! let me know once you're happy with testing and we'll get this merged.

kippandrew: Any idea why Travis is complaining? The output looks like CRAN check non-sense to me. Are these transient failures?

aronatkins: No clue, sadly, and other recent PRs passed.

aronatkins: When in doubt, regenereate! 🎉

kippandrew: I think the travis output is being truncated, but I found this running it locally.

PR #582, opened 2022-05-04
Allow selecting a target `image` in the bundle manifest

dbkegley:
### Approach

Updates `writeManifest` and `deployApp` with an optional `image` parameter which is used to set the `environment.image` field of the generated content manifest.

### Automated Tests

Added a new `writeManifest` testcase in `test-bundle.R` 

### QA Notes

1. `rsconnect::writeManifest(..., image = "test-image")` should generate a manifest in the bundle directory with an `environment.image` field.

```
{
  "version": 1,
  "locale": "en_US",
  "platform": "4.1.2",
  "metadata": {
    "appmode": "shiny",
    "primary_rmd": null,
    "primary_html": null,
    "content_category": null,
    "has_parameters": false
  },
  "environment": {
    "image": "test-image"
  },
...
```

Omitting the `image` parameter should generate a `manifest.json` with no `environment` field.

2. Push button deploy from the IDE should work for all R application types but will not allow you to select a target image.

3. The following functions should now accept an `image` argument to use during build/execution on the server:
`rsconnect::deployApp(..., image = "test-image")` 
`rsconnect::deployAPI(..., image = "test-image")`
`rsconnect::deployDoc(..., image = "test-image")`
`rsconnect::deploySite(..., image = "test-image")`

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -934,7 +934,7 @@ createAppManifest <- function(appDir, appMode, contentCategory, hasParameters,
 
   # if there is a target image, attach it to the environment
   if (!is.null(image)) {
-    manifest$environment$image <- image
+    manifest$environment <- list(image = image)
   }
 
   # indicate whether this is a quarto app/doc
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -222,7 +222,7 @@ enforceBundleLimits <- function(appDir, totalSize, totalFiles) {
 bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
                       contentCategory, verbose = FALSE, python = NULL,
                       condaMode = FALSE, forceGenerate = FALSE, quarto = NULL,
-                      isShinyApps = FALSE, metadata = list()) {
+                      isShinyApps = FALSE, metadata = list(), image = NULL) {
   logger <- verboseLogger(verbose)
 
   quartoInfo <- inferQuartoInfo(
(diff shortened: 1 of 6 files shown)


R/bundle.R
```diff
@@ -924,6 +932,11 @@ createAppManifest <- function(appDir, appMode, contentCategory, hasParameters,
   # add metadata
   manifest$metadata <- metadata
 
+  # if there is a target image, attach it to the environment
+  if (!is.null(image)) {
+    manifest$environment$image <- image
```
aronatkins, line 937: Do you need something like the following to make sure that `manifest$environment` exists?

```r
if (!is.null(image)) {
    manifest$environment <- list(image = image)
}
```

aronatkins (changes requested):

aronatkins (approved):

dbkegley:
Hm, Travis is failing but I'm not sure why. Do I need to re-generate the Rmds in `man` since I changed 2 public APIs?

toph-allen (approved):
Looks good to me. I'm guessing that the nested structure of `manifest -> environment -> image` is an existing convention from outside of `rsconnect`? I was worried that Aron's suggestion for assigning `list(image = image)` to `environment` would overwrite anything already there, but it doesn't look like there's anything else that writes to that property.

dbkegley:
`environment` is a new field in the Connect manifest that is being added so that we can select an image for a given piece of content when we are using off host execution. There's nothing else using that field today though

PR #583, opened 2022-05-09
showLogs takes a server parameter

### LINKED ISSUE #57: showLogs does not seem to take the "server" paramater

aronatkins:
fixes #57

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
   and websites by supplying the path to a Quarto executable
 * Added support for deploying Quarto content that uses only the `jupyter` runtime
 * Added support for selecting a target `image` in the bundle manifest
+* The `showLogs` function takes a `server` parameter. (#57)
 
 ## 0.8.25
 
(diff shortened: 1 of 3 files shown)


toph-allen (approved):
I'm fine merging this as-is. However, I think that JJ's comment on the issue [here](https://github.com/rstudio/rsconnect/issues/57#issuecomment-112513349) makes sense — since this function is shinyapps.io-only, perhaps the `server` parameter in the `accountInfo()` call on line 212 of `applications.R` should be hardcoded to "shinyapps.io"?

aronatkins:
@toph-allen - not constraining the server parameter gives us the ability to point to test infrastructure.

toph-allen:
> @toph-allen - not constraining the server parameter gives us the ability to point to test infrastructure.

That makes sense. I was going to suggest having it default to "shinyapps.io", but really, it's not a big deal, as users will only have to provide that parameter if they share an account name on that site and on another server.

PR #584, opened 2022-05-12
support launch.browser argument in connectUser()

daler:
When trying to connect to an RStudio Connect instance on a system with no browser available (e.g., command-line only on a cluster), the URL to visit is printed. But then `connectUser()` immediately exits when `browseURL` fails rather than waiting for `getAuthedUser()`. The result is that the user vists the URL and authenticates, making it seem like everything's ok, but the connection is not actually made.

This simply adds the ability to specify `launch.browser = FALSE` like in other functions to avoid exiting the function early.


daler:
Updated; for the record, this is a fix for ticket \#75690 to RStudio Support.

aronatkins (approved):

PR #585, opened 2022-05-12
Add cloud support.

ssinnott:
This adds the wiring that we have been using for cloud publishing. I think this is kinda foundational - it's very likely to change in the future. But the general approach is that,

1) Using shinyapps.io credentials you call `setAccountInfo(..., cloud = TRUE)`.
2) Then whenever you use this account it uses the `cloudClient` instead of the `lucidClient` - which handles the cloud workflow.
3) The biggest change is that the `application` resource is not created directly it is the `source` of an `output` resource. This has a couple parts,
  a. Whenever we get the "application" we actually want the output's source.
  b. We don't want to redirect to the application url after deployment - we want to redirect to the outputs url (which ends up in the cloud chrome).
  c. An output can have a relationship to a cloud project. When you are authoring code in cloud there's an environment variable `LUCID_APPLICATION_ID` - which is application ID related to the project that you are current in. This is used to establish a relationship between the `project` and `output` after it is created.
  d. An `output` ends up with a type that is derived from largely `appMode`.

Any suggestions in terms of the best way to integrate this into the client library would be welcome!


R/lucid.R
```diff
@@ -259,6 +259,292 @@ lucidClient <- function(service, authInfo) {
...
+      handleResponse(GET(service, authInfo, path, query))
+    },
+
+    createApplication = function(name, title, template, accountId) {
+      json <- list()
+      json$name <- name
+      # the title field is only used on connect
+      # the template and account id fields are not user by cloud.
+      # TODO: do we derive output type.
+      json$output_type <- "shiny"
```
ssinnott, line 364: I left this as a todo - but ultimately this is largely derived from `appMode` in the manifest. Which I think is derived here,

https://github.com/rstudio/rsconnect/blob/6801917d55d9d4035fbf643a6f4be3a6f795c764/R/bundle.R#L236

But wiring that through seems a bit messy - so I was wondering if there's a cleaner way to get that here that I'm missing?

R/accounts.R
```diff
@@ -235,7 +235,11 @@ setAccountInfo <- function(name, token, secret) {
     stop(stringParamErrorMessage("secret"))
 
   # create connect client
-  serverInfo <- shinyappsServerInfo()
+  if (cloud) {
+    serverInfo <- cloudServerInfo()
+  } else {
+    serverInfo <- shinyappsServerInfo()
```
aronatkins, line 241: Should we create a stricter separation between shinyapps.io and cloud? Just below this change, for example, is still attempting to use `lucidClient` even when `cloud = TRUE`.

R/accounts.R
```diff
@@ -468,7 +472,7 @@ resolveAccount <- function(account, server = NULL) {
 }
 
 isShinyapps <- function(server) {
-  identical(server, "shinyapps.io")
+  identical(server, "shinyapps.io") | identical(server, "rstudio.cloud")
```
aronatkins, line 475: How does this and `setAccountInfo` interact with some of the other hostnames that might be in use? I'm thinking about the staging stack.

R/lucid.R
aronatkins: Hm. Connect gets its content type indirectly from the uploaded bundle. It's not provided when the content is created. Content starts off with an "unknown" type, which receives its type with the first bundle upload and is thereafter prevented from changing.

How is the "template" argument used (currently always given the value "shiny")? Connect doesn't have that notion.

aronatkins (approved):
It's a pity that the server information doesn't identify the kind of thing we're deploying to; it leaves us with string comparisons and other, similar things.

I guess my only recommendation is to think about having more separation between shinyapps.io and the rstudio.cloud options rather than trying to scatter if statements everywhere.

aronatkins:
CI is probably failing because docs need updating.

R/lucid.R
ssinnott: Yeah - we followed a path more similar to how connect handles this. Which removes this as a parameter. This is now determined at build time from the bundle rather then as the bundle is uploaded.

R/accounts.R
ssinnott: So for our other environments we override the `rsconnect.shinyapps_url` option which changes the url. So like this for localhost:

`options(rsconnect.shinyapps_url="http://localhost:7100/v1")`

R/accounts.R
ssinnott: I took another pass at this and turned `lucidClient` into `lucidClientForAccount`. This ends up returning a client that is either configured to talk to shinyapps.io or one that is targeting rstudio.cloud depending on the account passed in which I think ends up cleaning up the references in other areas of the code base.

R/servers.R
```diff
@@ -84,6 +84,14 @@ shinyappsServerInfo <- function() {
                                "https://api.shinyapps.io/v1"))
 }
 
+cloudServerInfo <- function() {
+  info <- list(name = "rstudio.cloud",
+               certificate = inferCertificateContents(
+                 system.file("cert/shinyapps.io.pem", package = "rsconnect")),
+               url = getOption("rsconnect.shinyapps_url",
```
aronatkins, line 91: Is this reuse of the shinyapps.io cert & URL intentional?

aronatkins: Update: It is; different API calls, same API host.

R/lucid.R
```diff
@@ -1,8 +1,24 @@
+lucidClientForAccount <- function(account) {
+  authInfo <- account
 
-# return a list of functions that can be used to interact with lucid
-lucidClient <- function(service, authInfo) {
-  service <- parseHttpUrl(service)
+  # determine appropriate server information for account
+  if (account$server == cloudServerInfo()$name) {
```
aronatkins, line 5: Does this check allow for staging/prod testing?

aronatkins (approved):

R/lucid.R
ssinnott: It does because you can change the `rsconnect.shinyapps_url` options which targets a new URL - but doesn't influence the name.

ChaitaC:
Deployed a couple of apps to shinyapps.io and rsc.radixu.com to verify deployment works when using rsconnect branch `add-cloud-support`

PR #586, opened 2022-05-18
Better support for BioConductor

michaelmayer2:
Many customers have issues when publishing artefacts (mainly shiny apps) that use BioConductor packages to Connect. This comes about due to the different structure of the repository and the different commands used to install packages (e.g. `BiocManager::install()`) 

`renv` that is used by a number of users support BioConductor quite well, also when publishing to Connect. 

Users that are not using `renv` however struggle with using `rsconnect`. They have to use various workarounds, including manually setting up repositories, e.g. `options(repos=BiocManager::repositories())`, before publishing. 

This PR is bringing more native support for BioConductor repositories to Connect. 

The approach is:

If `BiocManager` is installed and `Bioc*` repositories are not set manually, it will add the repositories automatically. 

The subsequent algorithm of extracting repository and source information of packages is becoming much simpler as a consequence.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -96,10 +96,7 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
     paste0("x",grep('BioC', names(options("repos")), perl = TRUE, value = TRUE)) == "x" ) {
     
     repos <- BiocManager::repositories() 
-    } 
-
-#    biocRepos <- biocRepos[startsWith(names(biocRepos),"BioC")]    
-  
+  } 
 
   repo.packages <- available.packages(
     contriburl = contrib.url(repos, type = "source"),
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -93,7 +93,7 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
   #   * the non-existence of any bioconductor repo in the CRAN like 
   #            repo definitions 
   if (requireNamespace("BiocManager", quietly = TRUE) &&
-    !any(grepl("BioC", names(getOption("repos"))) 
+    !any(grepl("BioC", names(getOption("repos"))))) 
   {
     repos <- BiocManager::repositories() 
   } 
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -93,8 +93,8 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
   #   * the non-existence of any bioconductor repo in the CRAN like 
   #            repo definitions 
   if (requireNamespace("BiocManager", quietly = TRUE) &&
-    !any(grepl("BioC", names(getOption("repos"))) {
-    
+    !any(grepl("BioC", names(getOption("repos"))) 
+  {
     repos <- BiocManager::repositories() 
   } 
 
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -93,7 +93,7 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
   #   * the non-existence of any bioconductor repo in the CRAN like 
   #            repo definitions 
   if ("BiocManager" %in% installed.packages() && 
-    paste0("x",grep('BioC', names(options("repos")), perl = TRUE, value = TRUE)) == "x" ) {
+    !any(grepl("BioC", names(getOption("repos"))) {
     
     repos <- BiocManager::repositories() 
   } 
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -87,17 +87,20 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
     default = c("duplicates")
   )
 
-  # get Bioconductor repos if any
-  biocRepos <- repos[grep('BioC', names(repos), perl = TRUE, value = TRUE)]
-  biocPackages <- if (length(biocRepos) > 0) {
-    available.packages(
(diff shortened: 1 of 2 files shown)


R/dependencies.R
```diff
@@ -87,17 +87,17 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
...
-      type = "source",
-      filters = filters
-    )
-  }
+  # get Bioconductor repos if aniy
+  # We detect the presence of Bioconductor reopos by 
+  #   * existence of BiocManager amongst the installed packages
+  #   * the non-existence of any bioconductor repo in the CRAN like 
+  #            repo definitions 
+  if ("BiocManager" %in% installed.packages() &&
```
kevinushey, line 95:
```suggestion
  if (requireNamespace("BiocManager", quietly = TRUE) &&
```

R/dependencies.R
```diff
@@ -87,17 +87,17 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
...
-      filters = filters
-    )
-  }
+  # get Bioconductor repos if aniy
+  # We detect the presence of Bioconductor reopos by 
+  #   * existence of BiocManager amongst the installed packages
+  #   * the non-existence of any bioconductor repo in the CRAN like 
+  #            repo definitions 
+  if ("BiocManager" %in% installed.packages() && 
+    paste0("x",grep('BioC', names(options("repos")), perl = TRUE, value = TRUE)) == "x" ) {
```
kevinushey, line 96:
```suggestion
    !any(grepl("BioC", names(getOption("repos")))
```

R/dependencies.R
```diff
@@ -87,17 +87,17 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
...
+    
+    repos <- BiocManager::repositories()
```
kevinushey, line 98: Does this preserve the user's existing repositories if they've already customized them? What if you already have something like:

```
options(repos = c(RSPM = "https://rspm.company.org/repos/cran"))
```

Is that repository preserved?

kevinushey (commented):
For what it's worth, I think it's fairly likely that someone might have BiocManager installed, but could be deploying an application that doesn't use Bioconductor. But it's also probably fine to include Bioconductor repositories in the list of repositories anyhow.

Do we have access to the packages actually required by the deployment? If so, we could probably detect if packages like Biobase or other "base"-level Bioconductor packages are used.

R/dependencies.R
michaelmayer2: Yes, it preserves any other repo they may have defined, see: 

```
options()$repos; options(repos = c(RSPM = "https://rspm.company.org/repos/cran", RSPM2 = "https://rspm2.company.org/repos/cran")); paste("BiocManager"); BiocManager::repositories()
                                  RSPM                                  RSPM2 
 "https://rspm.company.org/repos/cran" "https://rspm2.company.org/repos/cran" 
[1] "BiocManager"
                                                BioCsoft                                                  BioCann 
           "https://bioconductor.org/packages/3.15/bioc" "https://bioconductor.org/packages/3.15/data/annotation" 
                                                 BioCexp                                            BioCworkflows 
"https://bioconductor.org/packages/3.15/data/experiment"       "https://bioconductor.org/packages/3.15/workflows" 
                                               BioCbooks                                                     RSPM 
          "https://bioconductor.org/packages/3.15/books"                    "https://rspm.company.org/repos/cran" 
                                                   RSPM2 
                  "https://rspm2.company.org/repos/cran" 
```

kevinushey: Great, thanks for confirming!

kevinushey (approved):
LGTM, but I still think it's worth considering whether we could infer a dependency on Bioconductor based on whether one of the packages that is part of the deployment is a Bioconductor package; presumedly that's possible to infer via the presence of a `biocViews` field in the DESCRIPTION file (if we're not already doing this)

R/dependencies.R
```diff
@@ -120,14 +120,7 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
     pkg <- records[i, "Package"]
     source <- records[i, "Source"]
     repository <- NA
-    # capture Bioconcutor repository
-    if (identical(source, "Bioconductor")) {
-      if (pkg %in% biocPackages) {
-        repository <- biocPackages[pkg, 'Repository']
-      }
-    } else if (isSCMSource(source)) {
```
aronatkins, line 128: What happened to the treatment of SCM sources?

michaelmayer2: Great question: As per my underatanding previously nothing happened in `R/dependencies.R`  if `isSCMSource(source)` was true. By removing this reference altogether, SCM sources are still respected. We may possibly be able to remove the function altogether (and the reference in `R/bundle.R` as well).

As a sanity check I pushed the faithful geyser app (with shiny as of latest in github) into my local RSC and it installed the same version of shiny from github. 

```
options(repos = c(CRAN = "https://packagemanager.rstudio.com/cran/latest"))
install.packages("remotes")
remotes::install_github("rstudio/shiny")
remotes::install_github("michaelmayer2/rsconnect")
```

michaelmayer2:
Hi @aronatkins & @kevinushey - the PR had some issues with going through Travis/CI - I now fixed the issues. 

Due to the use of `BiocManager` in `rsconnect` I had to add it as a `Suggests` in the `DESCRIPTION` file. Is that ok ? 

Would appreciate your feedback before pressing the "merge pull request" button.

michaelmayer2:
Closing this PR as code base has diverged too much - opening a new one with a consolidated list of changes.

PR #587, opened 2022-05-19
Add option to select `tar` implementation

toph-allen:
### Intent

Document `RSCONNECT_TAR` more comprehensively.

Fixes https://github.com/rstudio/connect/issues/21307

### Approach

This value could only be set via an environment variable. I added an option, `rsconnect.tar`, which brings it in line with other similar values, like `rsconnect.http` and `rsconnect.ca.bundle`. I don't think this will eliminate *all* possible confusion, because not all users will know about options, or will know how to get the path to a `tar` implementation. I'm not sure that we need to solve for that here.

The approach in detail:

- Move the `tar` lookup into its own function, `getTarImplementation()`. This function first checks the environment variable, since this was the previous only supported option. If that is empty, it checks for an `rsconnect.tar` option. It returns the first non-empty result found, and returns `"internal"` if both are empty.
- Adds a test of the new function.
- Alters the error message displayed if a long filename is detected to mention the option as well.
- Adds an entry to the `options.Rd` file about this option that also mentions the environment variable.
- Updated `NEWS.md`.

### Automated Tests

Added a test to `test-bundle.R` to confirm that all permutations of settings behave as expected.

### QA Notes

Note: #589 is based on this branch, and its verification just requires a simple deployment, so you can probably do this verification on that branch to get both at the same time.

After installing this package, you can verify that the new option works by setting it to a `tar` implementation that does not exist. After this, trying to deploy any content will fail.

```r
> library(rsconnect)
> options("rsconnect.tar" = "does/not/exist")
> rsconnect::deployApp(appDir = "tests/testthat/shinyapp-simple", server = "localhost")
Preparing to deploy application...DONE
Uploading bundle for application: 251...sh: does/not/exist: No such file or directory
Error: The file does not exist: /var/folders/q7/1b7h_nx931q6mc70r3w8k_2c0000gn/T//Rtmp41ZFwC/rsconnect-bundle6e817e8bcdf2.tar.gz
In addition: Warning message:
In system(cmd) :
 Error: The file does not exist: /var/folders/q7/1b7h_nx931q6mc70r3w8k_2c0000gn/T//Rtmp41ZFwC/rsconnect-bundle6e817e8bcdf2.tar.gz
```

If you also set an environment variable, it will not be used until you unset the option.

```r
> Sys.setenv("RSCONNECT_TAR" = "does/not/exist/envvar")
> rsconnect::deployApp(appDir = "tests/testthat/shinyapp-simple", server = "localhost")
Preparing to deploy application...DONE
Uploading bundle for application: 251...sh: does/not/exist: No such file or directory
Error: The file does not exist: /var/folders/q7/1b7h_nx931q6mc70r3w8k_2c0000gn/T//Rtmp41ZFwC/rsconnect-bundle6e8140dde813.tar.gz
In addition: Warning message:
In system(cmd) :
 Error: The file does not exist: /var/folders/q7/1b7h_nx931q6mc70r3w8k_2c0000gn/T//Rtmp41ZFwC/rsconnect-bundle6e8140dde813.tar.gz 
> options("rsconnect.tar" = NULL)
> rsconnect::deployApp(appDir = "tests/testthat/shinyapp-simple", server = "localhost")
Preparing to deploy application...DONE
Uploading bundle for application: 251...sh: does/not/exist/envvar: No such file or directory
Error: The file does not exist: /var/folders/q7/1b7h_nx931q6mc70r3w8k_2c0000gn/T//Rtmp41ZFwC/rsconnect-bundle6e813ed0ef63.tar.gz
In addition: Warning message:
In system(cmd) :
 Error: The file does not exist: /var/folders/q7/1b7h_nx931q6mc70r3w8k_2c0000gn/T//Rtmp41ZFwC/rsconnect-bundle6e813ed0ef63.tar.gz
```

If you unset both, deployment should succeed.

```r
> Sys.unsetenv("RSCONNECT_TAR")
> rsconnect::deployApp(appDir = "tests/testthat/shinyapp-simple", server = "localhost")
Preparing to deploy application...DONE
Uploading bundle for application: 251...DONE
Deploying bundle: 331 for application: 251 ...
[Connect] Building Shiny application...
[Connect] Bundle created with R version 3.6.3 is compatible with environment Local with R version 3.6.3 from /opt/R/3.6.3/bin/R 
```

If, like in these examples, you deployed using something in the `tests/testthat` folder, you delete the folder that `rsconnect` leaves. In R:

```r
> unlink("tests/testthat/shinyapp-simple/rsconnect", recursive = TRUE)
```

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -320,11 +320,11 @@ writeBundle <- function(bundleDir, bundlePath, verbose = FALSE) {
 getTarImplementation <- function() {
   # Check the RSCONNECT_TAR environment variable first. If that is unset, check
   # the rsconnect.tar option. If neither are set, use "internal"
-  tarImplementation <- Sys.getenv("RSCONNECT_TAR", unset = NA)
-  if (is.na(tarImplementation)) {
-    tarImplementation <- getOption("rsconnect.tar", default = NA)
+  tarImplementation <- getOption("rsconnect.tar", default = NA)
+  if (is.na(tarImplementation) || !nzchar(tarImplementation)) {
+    tarImplementation <- Sys.getenv("RSCONNECT_TAR", unset = NA)
   }
-  if (is.na(tarImplementation)) {
+  if (is.na(tarImplementation) || !nzchar(tarImplementation)) {
     tarImplementation <- "internal"
   }
   return(tarImplementation)
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -356,8 +356,8 @@ detectLongNames <- function(bundleDir, lengthLimit = 32) {
       warning("The bundle contains files with user/group names having more than ", lengthLimit,
               " characters: ", f, " is owned by ", info$uname, ":", info$grname, ". ",
               "Long user and group names cause the internal R tar implementation to produce invalid archives. ",
-              "Set the rsconnect.tar option or the RSCONNECT_TAR environment variable ",
-              "to use an external tar command.")
+              "Set the rsconnect.tar option or the RSCONNECT_TAR environment variable to the path to ",
+              "a tar executable to use that implementation.")
       return(invisible(TRUE))
     }
   }
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -318,8 +318,8 @@ writeBundle <- function(bundleDir, bundlePath, verbose = FALSE) {
 
 
 getTarImplementation <- function() {
-  # Check the RSCONNECT_TAR environment variable first. If that is unset, check
-  # the rsconnect.tar option. If neither are set, use "internal"
+  # Check the rsconnect.tar option first. If that is unset, check the
+  # RSCONNECT_TAR environment var. If neither are set, use "internal".
   tarImplementation <- getOption("rsconnect.tar", default = NA)
   if (is.na(tarImplementation) || !nzchar(tarImplementation)) {
     tarImplementation <- Sys.getenv("RSCONNECT_TAR", unset = NA)
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -306,7 +306,7 @@ writeBundle <- function(bundleDir, bundlePath, verbose = FALSE) {
   prevDir <- setwd(bundleDir)
   on.exit(setwd(prevDir), add = TRUE)
 
-  tarImplementation <- Sys.getenv("RSCONNECT_TAR", "internal")
+  tarImplementation <- getTarImplementation()
   logger(sprintf("Using tar: %s", tarImplementation))
 
   if (tarImplementation == "internal") {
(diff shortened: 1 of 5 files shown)


R/bundle.R
```diff
@@ -316,6 +316,21 @@ writeBundle <- function(bundleDir, bundlePath, verbose = FALSE) {
   utils::tar(bundlePath, files = NULL, compression = "gzip", tar = tarImplementation)
 }
 
+
+getTarImplementation <- function() {
+  # Check the RSCONNECT_TAR environment variable first. If that is unset, check
+  # the rsconnect.tar option. If neither are set, use "internal"
+  tarImplementation <- Sys.getenv("RSCONNECT_TAR", unset = NA)
```
aronatkins, line 323: The `rsconnect.ca.bundle` / `RSCONNECT_CA_BUNDLE` are evaluated in the opposite order - the option is used first.

Given that `rsconnect.tar` is new with this work, why not follow the same evaluation order?

Option takes precedence over the environment variable.

aronatkins (commented):
I would prefer consistency between option and environment variable precedence within the package. Are there existing conventions we should follow?

R/bundle.R
toph-allen: Yeah, I'll make this change. I was unsure whether it'd be better to be consistent or to let the older option take precedence, but thinking about it now, I completely agree with you.

R/bundle.R
```diff
@@ -316,6 +316,21 @@ writeBundle <- function(bundleDir, bundlePath, verbose = FALSE) {
   utils::tar(bundlePath, files = NULL, compression = "gzip", tar = tarImplementation)
 }
 
+
+getTarImplementation <- function() {
+  # Check the RSCONNECT_TAR environment variable first. If that is unset, check
```
aronatkins, line 321: The comment is inconsistent with the implementation.

tests/testthat/test-bundle.R
```diff
@@ -569,3 +571,29 @@ test_that("writeManifest: Sets environment.image in the manifest if one is provi
   manifest <- makeManifest(appDir, appPrimaryDoc = NULL)
   expect_null(manifest$environment)
 })
+
+test_that("tarImplementation: checks environment variable and option before using default", {
+  # Environment variable only set should use environment varaible
+  Sys.setenv("RSCONNECT_TAR" = "envvar")
```
aronatkins, line 577: Is there any point in restoring any incoming `RSCONNECT_TAR` or `rsconnect.tar` values on the way out of the test?

aronatkins (approved):

R/bundle.R
toph-allen: Fixed

tests/testthat/test-bundle.R
toph-allen: Adjusted so that this happens.

PR #588, opened 2022-05-20
Reorganize functions into different files (no actual changes) [DRAFT]

toph-allen:
This PR doesn't actually change any code, but it moves things around into different files, in an effort to organize things a little more logically. I'm not sure if it succeeds at that, and am happy to close it if it doesn't.

Principles I tried to follow:
- User-facing functions that aren't also supporting functions for multiple functions go in their own files.
  - Functions that are called only by one function go in the same file
- Functions that are called by many functions go in `[area]Utils.R` files.

List of the changes:
- `appDependencies.R`: Renamed from `dependencies.R`, now only contains `appDependencies()`.
- `writeManifest.R` only contains `writeManifest()`.
- `deployApp.R` only contains `deployApp()`, `bundleApp()`, and other code that it uses exclusively.

UPDATE: I'm going to redo my re-organizing. VS Code had "format on paste" enabled globally, and I don't want that turned on for this task.


toph-allen:
Superseded by https://github.com/rstudio/rsconnect/pull/589.

PR #590, opened 2022-05-24
Use GitHub Actions for CI

toph-allen:
This PR removes Travis CI from the `rsconnect` repo and adds GitHub Actions.

Fixes https://github.com/rstudio/connect/issues/21525

There are two actions:

- `R-CMD-check.yaml`: Checks the package. Runs on a variety of operating systems and R versions. Mostly follows [`packrat`'s implementation](https://github.com/rstudio/packrat/blob/f16d36ec1f5dc643fbddc13f5ffc7420f2f6445f/.github/workflows/R-CMD-check.yaml). The template provided by `usethis` still uses the `v1` actions.
- `pkgdown.yaml`: Publishes the docs to GitHub Pages. This is an amalgamation of the [`v1`](https://github.com/r-lib/actions/blob/v1/examples/pkgdown.yaml) and [`v2`](https://github.com/r-lib/actions/blob/v2-branch/examples/pkgdown.yaml) examples: it uses the setup from `v2` but the deployment step from `v1`, which uses a `pkgdown` function rather than a GitHub action step to publish.
  - Because this package releases infrequently, I thought it made more sense to only run this on releases and manually, not on pushes to `main`.

Open questions:

- Any changes to the R version matrix in `R-CMD-check`?
- Is the approach I took with `pkgdown` reasonable? I.e. 1. Should I just use the entirety of the `v2` template? 2. Should our GH Pages track `main`?

### Removing Travis

I removed `.travis.yml`. Looking more into Travis, I don't think any other actions are required.

Travis currently has access to "All repositories". The only way to _remove_ `rsconnect` from Travis is to change this to "Only select repositories", which I feel like would break a lot of things.

Screen Shot 2022-05-24 at 03 45 04 PM@2x

The previous CI runs done in Travis exist in Travis. Future CI runs done with GH Actions will exist there.

### QA Notes

- No QA needed; this is a change to CI only.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -26,8 +26,13 @@ jobs:
           extra-packages: any::pkgdown, local::.
           needs: website
 
-      - name: Deploy package
-        run: |
-          git config --local user.name "GitHub Actions"
-          git config --local user.email "actions@users.noreply.github.com"
-          Rscript -e 'pkgdown::deploy_to_branch(new_process = FALSE)'
+      - name: Build site
+        run: pkgdown::build_site_github_pages(new_process = FALSE, install = FALSE)
+        shell: Rscript {0}
+
+      - name: Deploy to GitHub pages 🚀
+        uses: JamesIves/github-pages-deploy-action@4.3.3
+        with:
+          clean: false
+          branch: gh-pages
+          folder: docs


aronatkins (approved):
We may want to follow the v2 actions recommendation and separate the pkgdown build from deploy, but that's only a "maybe" suggestion.

PR #595, opened 2022-06-14
Disallow deploying `.qmd` files with `rmd-*` app modes

### LINKED ISSUE #594: rsconnect: disallow deployment of qmd files as R Markdown
Currently, `.qmd` files will be deployed as R Markdown content if no `quarto` argument is passed to `deployApp()`, and an R Markdown manifest will be created if no `quarto` is passed to `writeManifest()`.

This is because Quarto files and R Markdown files are detected [with the same regular expression](https://github.com/rstudio/rsconnect/blob/27c50cb1be28a05ad2a62682e867515580ab1691/R/bundle.R#L410), and the `quarto` argument determines the app mode assigned.

https://github.com/rstudio/rsconnect/blob/27c50cb1be28a05ad2a62682e867515580ab1691/R/bundle.R#L413-L420

https://github.com/rstudio/rsconnect/blob/27c50cb1be28a05ad2a62682e867515580ab1691/R/bundle.R#L430-L437

To change this, we should add a separate regex to detect `qmdFiles`. If `rmdFiles > `, the deployment will receive an `rmd`-related app mode. Otherwise, it will receive a `qmd`-related app mode.

Our reasoning in leaving this the other way was to make the selection of Quarto-related app modes manual, by passing in the `quarto` argument, which is a path to a Quarto binary. If the user does not pass that in, we won’t deploy as Quarto.

It almost seems silly to ask them to pass in a Quarto argument explicitly when we could just run `quarto inspect`. But I guess, they could get the same behavior by passing in `”quarto”` to that argument, because I don’t think we require a full path.

toph-allen:
### Intent

Previously, Quarto documents and projects would deploy as `rmd-static` or `rmd-shiny` if they were deployed without Quarto info available (either via running `quarto inspect` or via the `metadata` argument).

With this PR, `rsconnect` will stop with an error if that condition occurs, and advise the user to pass the path to a Quarto binary to the `quarto` argument.

Fixes #594 

### Approach

Previously, we would infer one set of `rmdFiles` that would contain both Quarto and R Markdown files, and decide between `rmd-*` and `quarto-*` app modes based on the presence or absence of `quartoInfo`, which is the inferred Quarto info from either running Quarto Inspect or unpacking metadata.

Now, we find `.Rmd` and `.qmd` files separately. The list of file names is stored in `rmdFiles` and `qmdFiles` separately, and a logical vector of whether they're `runtime: shiny` is stored in `shinyRmdFiles` and `shinyQmdFiles`. We also look for `^_quarto.y(a)?ml$`.

A deployment requires Quarto if… I was going to write out the condition, but the code is more readable than English.

```r
  hasQuartoSupportedFiles <- any(length(qmdFiles) > 0, length(rmdFiles > 0))
  requiresQuarto <- (hasQuartoSupportedFiles && hasQuartoYaml) || length(qmdFiles) > 0
```

*Any* `.qmd` files make us require Quarto info. or, if we have Quarto-compatible files and a `_quarto.yml`, we need Quarto info. In those cases, if we don't have it, we raise an error.

After this point, we take the presence of the `quartoInfo` variable to determine whether we use `quarto` or `rmd` app modes, in each locations where we assign `*-shiny` and `*-static`. This still will let you deploy R Markdown files as Quarto content, but prevents you from deploying Quarto content as non-Quarto.

The RStudio IDE was adding the line `PackageCleanBeforeInstall: Yes` to `.Rproj` files for a while. Now it apparently removes that line. I committed the change, because it happened with every install of a package.

### Automated Tests

Three tests were added:
- Attempt to create a manifest for `quarto-website-r`, leaving the `quarto` argument `NULL`, and expect the error message.
- Attempt to create a manifest for `quarto-doc-none`, leaving the `quarto` argument `NULL`, and expect the error message.
- Attempt to create a manifest for `test-reds/simple.Rmd`, passing in `quarto`. Ensure that the manifest has the `quarto-static` app mode and `knitr` engine.

### QA Notes

Install this branch of `rsconnect` from the RStudio IDE.

Quarto content in the `tests/testthat` directory should deploy correctly:

```r
rsconnect::deployApp(appDir = "tests/testthat/quarto-website-r", server = "localhost", quarto = quarto::quarto_path())
```

Without passing the `quarto` argument, you should see an error:

```r
rsconnect::deployApp(appDir = "tests/testthat/quarto-website-r", server = "localhost")
```

After testing, you'll want to go into the `tests/testthat/quarto-website-r` folder and remove the `rsconnect` folder that was created.

Deploying an R Markdown document from the IDE with this version installed should still deploy with an R Markdown app mode. (I tested this myself, but it's good to verify in a different environment.)

### Checklist

- [x] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md`

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -406,14 +406,32 @@ inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo) {
     return("shiny")
   }
 
-  # Determine if we have Rmd and if they are (optionally) need the Shiny runtime.
-  rmdFiles <- grep("^[^/\\\\]+\\.[rq]md$", files, ignore.case = TRUE, perl = TRUE, value = TRUE)
+  # Determine if we have Rmd files, and if they use the Shiny runtime.
+  rmdFiles <- grep("^[^/\\\\]+\\.rmd$", files, ignore.case = TRUE, perl = TRUE, value = TRUE)
   shinyRmdFiles <- sapply(file.path(appDir, rmdFiles), isShinyRmd)
 
-  # An Rmd file with a Shiny runtime uses rmarkdown::run.
+  # Determine if we have qmd files, and if they use the Shiny runtime
+  qmdFiles <- grep("^[^/\\\\]+\\.qmd$", files, ignore.case = TRUE, perl = TRUE, value = TRUE)
+  shinyQmdFiles <- sapply(file.path(appDir, qmdFiles), isShinyRmd)
+
+  # Trying to deploy Quarto and R Markdown files simultaneously is an error.
+  if (length(rmdFiles) > 0 && length(qmdFiles) > 0) {
+    stop("Cannot infer app mode because there are both .qmd and .Rmd files in deployment files.")
+  }
+
+  # To deploy Quarto content, we need to have received or inferred Quarto metadata.
+  missingQuartoInfoErrorText <- paste(
+    "Attempting to deploy Quarto project without successfully running 'quarto inspect'.",
+    "Please provide the path to a quarto binary to the 'quarto'."
+  )
+
+  # Shiny or Quarto documents with "server: shiny" in their YAML front matter
+  # are rmd-shiny or quarto-shiny.
   if (any(shinyRmdFiles)) {
+    return("rmd-shiny")
+  } else if (any(shinyQmdFiles)) {
     if (is.null(quartoInfo)) {
-      return("rmd-shiny")
+      stop(missingQuartoInfoErrorText)
     } else {
       return("quarto-shiny")
     }
@@ -427,10 +445,13 @@ inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo) {
     return("shiny")
   }
 
-  # Any non-Shiny R Markdown documents are rendered content (rmd-static).
+  # Any non-Shiny R Markdown or Quarto documents are rendered content and get
+  # rmd-static or quarto-static.
   if (length(rmdFiles) > 0) {
+    return("rmd-static")
+  } else if (length(qmdFiles) > 0) {
     if (is.null(quartoInfo)) {
-      return("rmd-static")
+      stop(missingQuartoInfoErrorText)
     } else {
       return("quarto-static")
     }
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -422,7 +422,7 @@ inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo) {
   # To deploy Quarto content, we need to have received or inferred Quarto metadata.
   missingQuartoInfoErrorText <- paste(
     "Attempting to deploy Quarto project without successfully running 'quarto inspect'.",
-    "Please provide the path to a quarto binary to the 'quarto'."
+    "Please provide the path to a quarto binary to the 'quarto' argument."
   )
 
   # Shiny or Quarto documents with "server: shiny" in their YAML front matter
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -421,7 +421,7 @@ inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo) {
 
   # To deploy Quarto content, we need to have received or inferred Quarto metadata.
   missingQuartoInfoErrorText <- paste(
-    "Attempting to deploy Quarto project without successfully running 'quarto inspect'.",
+    "Attempting to deploy Quarto project without Quarto metadata.",
     "Please provide the path to a quarto binary to the 'quarto' argument."
   )
 
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -413,27 +413,27 @@ inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo) {
   # Determine if we have qmd files, and if they use the Shiny runtime
   qmdFiles <- grep("^[^/\\\\]+\\.qmd$", files, ignore.case = TRUE, perl = TRUE, value = TRUE)
   shinyQmdFiles <- sapply(file.path(appDir, qmdFiles), isShinyRmd)
+  hasQuartoYaml <- any(grepl("^_quarto.y(a)?ml$", x = files, ignore.case = TRUE, perl = TRUE))
 
-  # Trying to deploy Quarto and R Markdown files simultaneously is an error.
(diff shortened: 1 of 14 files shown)


R/bundle.R
```diff
@@ -406,14 +406,32 @@ inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo) {
...
+  rmdFiles <- grep("^[^/\\\\]+\\.rmd$", files, ignore.case = TRUE, perl = TRUE, value = TRUE)
   shinyRmdFiles <- sapply(file.path(appDir, rmdFiles), isShinyRmd)
 
-  # An Rmd file with a Shiny runtime uses rmarkdown::run.
+  # Determine if we have qmd files, and if they use the Shiny runtime
+  qmdFiles <- grep("^[^/\\\\]+\\.qmd$", files, ignore.case = TRUE, perl = TRUE, value = TRUE)
+  shinyQmdFiles <- sapply(file.path(appDir, qmdFiles), isShinyRmd)
+
+  # Trying to deploy Quarto and R Markdown files simultaneously is an error.
+  if (length(rmdFiles) > 0 && length(qmdFiles) > 0) {
```
aronatkins, line 418: We should think about this error a bit more. It is possible to have a Quarto project that uses a `document.Rmd` file. Similarly, it is possible to have a Quarto project with a `notebook.ipynb` file.

It might be helpful if we captured an enumeration of file name collections and explained which combinations lead to different app modes, especially in the presence / non-presence of `quartoInfo`.

Here's one example:

* `_quarto.yml`
* `analysis.Rmd`
* `playground.ipynb`

What content type should be used when this code is deployed to Connect? Why?

Similarly, what about:

* `report.Rmd` with non-nil `quartoInfo`

Is this information more a signal to use Quarto or to use R Markdown?

aronatkins (commented):
I'm unconvinced about the error when we have both Rmd and Qmd files.

R/bundle.R
toph-allen: Yeah, you raise a good point.

The app mode is inferred _after_ we have put Quarto info into the format we desire, so by this point it's not clear whether it comes from running `quarto inspect` or as metadata passed in from the Quarto package.

I'm not sure what Connect would do with the first example you provide.

We had been using `quartoInfo`'s presence as an indicator of user intent to deploy Quarto. If continue to rely on that signal, we could do something like the following.

- with `quartoInfo`:
    - App mode is `quarto-static`, or `quarto-shiny` if any `shinyRmdFiles` or `shinyQmdFiles`.
- without `quartoInfo`:
    - App mode is `rmd-static` or `rmd-shiny`, as would have been otherwise detected
    - Any `qmdFiles` or `shinyQmdFiles` results in an error?

(Thinking about this more. When Quarto renders R Markdown files, their chunk arguments are parsed as R Markdown would, I'm guessing.)

toph-allen:
From a discussion today, the behavior of this code is going to change to the following:

1. Content will require Quarto metadata (`quartoInfo`) if we find any `.qmd`s or a `_quarto.yml` in our deployment `files`.
2. If content requires `quartoInfo`, missing `quartoInfo` info will stop with an error.
3. If we have `quartoInfo`, we will use `quarto-shiny` and `quarto-static` app modes instead of `rmd-shiny` and `rmd-static` app modes.

This will allow users to deploy non-`.qmd` files with Quarto, but stop them from deploying `.qmd` files without Quarto.

toph-allen:
@aronatkins Discovered another bug in this while working on Sprint Review slides.

If you try to publish the "finished document" via the IDE, `_quarto.yaml` might be included by default, and if it is, we will detect that file and the error will occur.

aronatkins (approved):

toph-allen:
The Quarto app mode is not correctly assigned when an `appPrimaryDoc` isn't present, it seems.

I'm duplicating [this comment](https://github.com/rstudio/rsconnect/issues/594#issuecomment-1168996800) here so that discussion related to the PR itself can occur on the PR.

When I deployed the Rmd file, it deployed with a Quarto app mode.

```
> rsconnect::deployApp(".", appPrimaryDoc = "rmdfile.Rmd", quarto = quarto::quarto_path(), server = "rsc.radixu.com")
Preparing to deploy document...DONE
Uploading bundle for document: 5672...DONE
Deploying bundle: 13125 for document: 5672 ...
[Connect] Building Quarto document...
```

https://rsc.radixu.com/connect/#/apps/6cc07579-932b-430e-9399-fa3a3b67905b/info/5454

When I deployed without passing in `appPrimaryDoc`, it deployed with an R Markdown app mode.

jthomasmock:
> When I deployed the Rmd file, it deployed with a Quarto app mode.
>
> ```r
>  rsconnect::deployApp(".", appPrimaryDoc = "rmdfile.Rmd",
>    quarto = quarto::quarto_path(), server = "rsc.radixu.com")
>
> Preparing to deploy document...DONE
> Uploading bundle for document: 5672...DONE
> Deploying bundle: 13125 for document: 5672 ...
> [Connect] Building Quarto document...
> ```

In my mind, specifying should always publish Quarto
- `format: html` in the YAML even if a `.Rmd`
- `quarto = quarto::quarto_path()` in `rsconnect`

 It'd be very surprising as a user to specify a Quarto path and end up with a RMarkdown type document.

jthomasmock:
> If I understand correctly, in this case, passing in the `quarto` argument would fail unless `quarto inspect` runs successfully, rather than falling back to other app modes? In this case, we would need new error messages for those failures.

Yes, I think we'd rather throw an error on failure to treat as Quarto rather than silently publish as Quarto.

> I don't understand the first bullet point. Are you suggesting that we should force Quarto for R Markdown docs that output HTML, even if no Quarto arg is given? As the code is written right now, we can't do that without an explicit Quarto path, and we've tried to be careful about not changing the package's behavior in response to consistent inputs.

Sorry for lack of context!

In the YAML header:

`format: html` indicates Quarto
`output: html_document` indicates RMarkdown

aronatkins:
We could expand `inferQuartoInfo` to take the primary doc and the files list.
1. inspect the directory and use its result when successful.
2. inspect the primary doc when non-NULL.
3. inspect the first doc in the files list when there is no explicit primary doc.

toph-allen:
Per our discussion in today's sync, I'm going to move the current discussion to a new GitHub issue, and this PR has solved the problem it set out to solve.

PR #598, opened 2022-06-24
Add ragg as a dependency to shiny apps (if it's installed locally)

cpsievert:
For more context, see https://github.com/rstudio/shiny/pull/3654

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -500,13 +500,13 @@ inferRPackageDependencies <- function(appMode, hasParameters, documentsHavePytho
   }
   if (appMode == "quarto-shiny") {
     # Quarto Shiny documents are executed with rmarkdown::run
-    deps <- c(deps, "rmarkdown", "shiny")
+    deps <- c(deps, "rmarkdown", shinyDeps())
   }
   if (appMode == "rmd-shiny") {
-    deps <- c(deps, "rmarkdown", "shiny")
+    deps <- c(deps, "rmarkdown", shinyDeps())
   }
   if (appMode == "shiny") {
-    deps <- c(deps, "shiny")
+    deps <- c(deps, shinyDeps())
   }
   if (appMode == "api") {
     deps <- c(deps, "plumber")
@@ -515,6 +515,14 @@ inferRPackageDependencies <- function(appMode, hasParameters, documentsHavePytho
     deps <- c(deps, "reticulate")
   }
   unique(deps)
+}
+
+# shiny::renderPlot() now uses ragg for png rendering if it's installed,
+# so include it if it's installed locally (this way users will get consistent
+# PNG results without needing to add library(ragg) to the app).
+# https://github.com/rstudio/shiny/pull/3654
+shinyDeps <- function() {
+  c("shiny", if (system.file(package = "ragg") != "") "ragg")
 }
 
 isWindows <- function() {
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -546,9 +546,8 @@ inferRPackageDependencies <- function(appMode, hasParameters, documentsHavePytho
 # will get consistent PNG results without needing to add library(ragg) to the
 # app). https://github.com/rstudio/shiny/pull/3654
 shinyDeps <- function() {
-  has_ragg <- system.file(package = "ragg") != ""
-  new_shiny <- packageVersion("shiny") >= "1.7.2"
-  c("shiny", if (has_ragg && new_shiny) "ragg")
+  include_ragg <- isAvailable("ragg") && isAvailable("shiny", "1.7.2")
+  c("shiny", if (include_ragg) "ragg")
 }
 
 isWindows <- function() {
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -517,12 +517,14 @@ inferRPackageDependencies <- function(appMode, hasParameters, documentsHavePytho
   unique(deps)
 }
 
-# With shiny v1.7.2 or higher, renderPlot() will default to ragg for png
-# rendering (if it's installed), so include it if it's installed locally (this
-# way users will get consistent PNG results without needing to add library(ragg)
-# to the app). https://github.com/rstudio/shiny/pull/3654
+# With shiny v1.7.2 or higher, renderPlot() defaults to ragg for png rendering
+# (if it's installed), so include it if it's installed locally (this way users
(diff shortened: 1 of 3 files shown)


R/bundle.R
```diff
@@ -517,6 +517,14 @@ inferRPackageDependencies <- function(appMode, hasParameters, documentsHavePytho
   unique(deps)
 }
 
+# shiny::renderPlot() now uses ragg for png rendering if it's installed,
+# so include it if it's installed locally (this way users will get consistent
+# PNG results without needing to add library(ragg) to the app).
+# https://github.com/rstudio/shiny/pull/3654
+shinyDeps <- function() {
+  c("shiny", if (system.file(package = "ragg") != "") "ragg")
```
aronatkins, line 525: Is there any merit to narrowing this further? For example, add `ragg` only when the Shiny version is >= 1.7.1 (or whatever)?

Is there a plan to make ragg the default (and therefore an explicit dependency)?

R/bundle.R
```diff
@@ -517,6 +517,14 @@ inferRPackageDependencies <- function(appMode, hasParameters, documentsHavePytho
...
+}
+
```
aronatkins, line 527: Would you mind adding a NEWS entry?

R/bundle.R
aronatkins: With this explanation, I'm OK without the version check.

Do you think there are cases where folks will want to opt-out of `ragg`? In other words: Are there systems that will not be able to install `ragg`? This change is adding it as an explicit requirement to the manifest / packrat lock -- which means that deploys will err if those systems cannot build / install `ragg`.

wch: Hm, yes, it is possible that systems won't be able to install ragg. [ragg](https://cran.r-project.org/web/packages/ragg/) imports [systemfonts](https://cran.r-project.org/web/packages/systemfonts/) and [textshaping](https://cran.r-project.org/web/packages/textshaping/). Here are the system requirements for each:

* ragg: C++11, freetype2, libpng, libtiff, libjpeg
* systemfonts: C++11, fontconfig, freetype2
* textshaping: C++11, freetype2, harfbuzz, fribidi

Up until this point, I thought that we could assume that the deployment targets could build/install these packages. Is that not the case?

IIRC, we used to install Cairo by default on shinyapps.io because of plot rendering issues. Do you remember if that's the case? It would be nice if we could do something similar today with ragg (but of course it wouldn't just be on shinyapps.io).

cpsievert: If we're worried about some environments not being able to install `{ragg}` (which seems like a valid concern since IIRC these packages have had issue in the past with older R versions), maybe it is worth adding the shiny version check.

aronatkins (approved):
Thanks for that last round of adjustment. I'll coordinate some testing from this end and handle merging.

R/bundle.R
```diff
@@ -541,6 +541,15 @@ inferRPackageDependencies <- function(appMode, hasParameters, documentsHavePytho
   unique(deps)
 }
 
+# With shiny v1.7.2 or higher, renderPlot() defaults to ragg for png rendering
+# (if it's installed), so include it if it's installed locally (this way users
+# will get consistent PNG results without needing to add library(ragg) to the
+# app). https://github.com/rstudio/shiny/pull/3654
+shinyDeps <- function() {
+  include_ragg <- isAvailable("ragg") && isAvailable("shiny", "1.7.2")
```
aronatkins, line 549: As written, it does not appear as if we can exercise this path, since the current Shiny on GitHub is identified as `1.7.1.9003`.

```r
rsconnect:::shinyDeps()
# => [1] "shiny"
```

Should we be checking that the Shiny version is greater than 1.7.1?

aronatkins: Addressing this issue from here, as we are actively testing.

R/utils.R
```diff
@@ -237,3 +237,11 @@ fileMD5.as.string <- function(path) {
 md5.as.string <- function(md5) {
   paste(md5, collapse = "")
 }
+
+isAvailable <- function(package, version = NULL) {
+  installed <- nzchar(system.file(package = package))
+  if (is.null(version)) {
+    return(installed)
+  }
+  installed && isTRUE(utils::packageVersion(package) >= version)
```
aronatkins, line 246: This comparison is incorrect, as we want version comparison via `compareVersion`, not string comparisons.

Addressing myself, as we are actively testing.

aronatkins: Wrong about this; local testing was relying on strings when testing version comparisons, not the object returned by `packageVersion`.

R/bundle.R
aronatkins: @cpsievert - changed this to check for 1.7.1.9000 and higher.

R/bundle.R
hafen: Just to echo the concern of `ragg` having some not-so-standard system dependencies (`textshaping` requires harfbuzz and fribidi), I just tried to re-deploy an app to a connect server with nontrivial updates to the app (some changes in text) and the deploy failed (did not fail a few days prior) because harfbuzz and fribidi weren't available on the server.

It took a lot of head scratching and digging to figure out that the only thing that had changed between app deployments is that `ragg` was not installed locally before the failed deployment and at some point got installed (not because of an app dependency) before the failed deployment. Because `rsconnect` found `ragg` locally it then added this as a dependency, which resulted in the failed deployment. I had a hard time tracking down why `ragg` and `textshaping` were in my packrat snapshot.

Anyway, that's a longwinded way to say that it might be good to at least provide a message when extra dependencies are added that are not inferred directly from the app code to help people understand why these dependencies are there.

shearerpmm: Echoing hafen. We had exactly the same issue with a mysterious failed deployment. We control our dev environment very carefully with an renv lockfile, and we could not understand why rsconnect had randomly decided that we needed to install textshaping when we hadn't asked it to do so. I spent an hour playing detective in the log files and the google search bar until I came to the rsconnect NEWS.md item about the new implicit behavior in rsconnect 0.8.27.

aronatkins: The addition of `ragg` has been reverted with https://github.com/rstudio/rsconnect/pull/625, which should be released to CRAN soon.

shearerpmm: Thank you - I am fighting for RSConnect/Shiny to be considered a legitimate production environment at my company, and I think this attention to engineering quality will reassure everyone that we won't regret the decision.

PR #605, opened 2022-07-05
add better way to deal with BioConductor

michaelmayer2:
Users that do not manually define BioConductor Repositories in their Rprofiles are not able to use BioConductor packages with RStudio Connect. 

This patch will allow `rsconnect`
* to behave as before when BioConductor repositories are pre-specified 
* auto-detect if BioConductor packages are present (installed via `BiocManager::install()`) and add them to the `manifest.json`
* set the bioconductor version via an `rsconnect` option`rsconnect.biocVersion`. 

This is a successor to https://github.com/rstudio/rsconnect/pull/586

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -1015,8 +1015,27 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
     default = c("duplicates")
   )
 
+  # get Bioconductor version if specified
+  biocVersion <- getOption("rsconnect.biocVersion")
   # get Bioconductor repos if any
-  biocRepos <- repos[grep('BioC', names(repos), perl = TRUE, value = TRUE)]
+  biocRepos <- if (any(startsWith(names(repos),"BioC"))) {
+        # User already has specified BioConductor repositories
+        # ==> only need to extract the BioConductor repos from the repo list
+	repos[grep('BioC', names(repos), perl = TRUE, value = TRUE)]
+  } else if (requireNamespace("BiocManager", quietly = TRUE)) {
+        # User has used BiocManager::install() only
+        # ==> need to set/define repositories 
+        tmpRepos <- if (is.null(biocVersion)) {
+                 # Use default Bioconductor version
+		 tmpRepos=BiocManager::repositories() 
+        } else if (!is.na(as.numeric(biocVersion))) {
+		 # Use specified Bioconductor version 
+	         tmpRepos=BiocManager::repositories(version=biocVersion) 
+        } else {
+  	         cat(sprintf("BioConductor Version %s specified but not a number.\n", biocVersion)) 
+        }
+        tmpRepos[startsWith(names(tmpRepos),"BioC")] 
+  }
   biocPackages <- if (length(biocRepos) > 0) {
     available.packages(
       contriburl = contrib.url(biocRepos, type = "source"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,6 +23,7 @@ Imports:
     tools,
     yaml (>= 2.1.5)
 Suggests:
+    BiocManager,
     RCurl,
     callr,
     httpuv,


R/bundle.R
```diff
@@ -1015,8 +1015,27 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
     default = c("duplicates")
   )
 
+  # get Bioconductor version if specified
+  biocVersion <- getOption("rsconnect.biocVersion")
   # get Bioconductor repos if any
-  biocRepos <- repos[grep('BioC', names(repos), perl = TRUE, value = TRUE)]
+  biocRepos <- if (any(startsWith(names(repos),"BioC"))) {
```
kevinushey, line 1021: nit: `startsWith()` requires R 3.3.0; does rsconnect still support R 3.2.x?

kevinushey, line 1021:
```suggestion
  biocRepos <- if (any(startsWith(names(repos), "BioC"))) {
```

R/bundle.R
```diff
@@ -1015,8 +1015,27 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
...
+        # User already has specified BioConductor repositories
+        # ==> only need to extract the BioConductor repos from the repo list
+	repos[grep('BioC', names(repos), perl = TRUE, value = TRUE)]
```
kevinushey, line 1024:
```suggestion
	repos[startsWith(names(repos), "BioC")]
```

That is, might as well make sure this condition matches the version above?

R/bundle.R
```diff
@@ -1015,8 +1015,27 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
...
+  } else if (requireNamespace("BiocManager", quietly = TRUE)) {
+        # User has used BiocManager::install() only
+        # ==> need to set/define repositories 
+        tmpRepos <- if (is.null(biocVersion)) {
+                 # Use default Bioconductor version
+		 tmpRepos=BiocManager::repositories()
```
kevinushey, line 1030:
```suggestion
		 BiocManager::repositories() 
```

I don't think you need the assignment here if you're already assigning the result of the `if ()` expression.

R/bundle.R
```diff
@@ -1015,8 +1015,27 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
...
+        # User already has specified BioConductor repositories
+        # ==> only need to extract the BioConductor repos from the repo list
+	repos[grep('BioC', names(repos), perl = TRUE, value = TRUE)]
+  } else if (requireNamespace("BiocManager", quietly = TRUE)) {
+        # User has used BiocManager::install() only
+        # ==> need to set/define repositories 
+        tmpRepos <- if (is.null(biocVersion)) {
+                 # Use default Bioconductor version
+		 tmpRepos=BiocManager::repositories() 
+        } else if (!is.na(as.numeric(biocVersion))) {
```
kevinushey, line 1031: I don't think this is a great check; it would fail if Bioconductor ever switched to a 3-component version system, e.g.

```
> as.numeric("3.10.1")
Warning: NAs introduced by coercion
[1] NA
```

If you need some kind of validation, you could check that `numeric_version(biocVersion)` gives a valid version.

R/bundle.R
```diff
@@ -1015,8 +1015,27 @@ snapshotRDependencies <- function(appDir, implicit_dependencies=c(), verbose = F
...
+		 # Use specified Bioconductor version 
+	         tmpRepos=BiocManager::repositories(version=biocVersion) 
+        } else {
+  	         cat(sprintf("BioConductor Version %s specified but not a number.\n", biocVersion)) 
+        }
+        tmpRepos[startsWith(names(tmpRepos),"BioC")]
```
kevinushey, line 1037: Is this code still going to work in the "invalid bioc version" case? (What happens if `tmpRepos` is `NULL`?)

kevinushey:
Also, super nit: the project name is Bioconductor (no capitalized C); however, the shortened version is usually stylized as "BioC".

hadley:
Closing this because this code has been almost completely rewritten. @michaelmayer2 if this is still important, I'm happy to work with you on a fix.

PR #608, opened 2022-07-12
Add Quarto to CI so that Quarto tests are not skipped.

toph-allen:
### Intent

Quarto tests are skipped if no Quarto installation is detected. This PR adds a Quarto installation to our CI workflows so that Quarto tests run in CI.

This surfaced a bug with `quartoInspect()` on Windows with non-Quarto content, so this PR solves that too. The bug prevents the successful deployment of non-Quarto content when a Quarto path is passed to the `quarto` argument of `writeManifest()`, `deployApp()`, etc. In those situations, we want deployment to continue with a non-Quarto app mode, but it would cause deployment to stop.

### Approach

To install Quarto, I used the [`quarto-setup`](https://github.com/quarto-dev/quarto-actions/tree/main/setup) action from [quarto-dev/quarto-actions](https://github.com/quarto-dev/quarto-actions).

The Quarto-related bug on windows stems from seemingly different behavior of `system2()` on Windows. I'm not 100% sure about this, but I'm trying to narrow it down. Here's my current understanding:

Running `quartoInspect()` when the `appDir` and `appPrimaryDoc` passed are not Quarto-compatible should return `NULL`; on Windows it failed.

This worked by calling `quarto inspect` like so:

```r
args <- c("inspect", path.expand(path))
inspect <- suppressWarnings(system2(quarto, args, stdout = TRUE, stderr = FALSE))
```

When `quarto inspect` is called on a non-Quarto path, it outputs on `stderr`, and in this case, `system2()` returns `NULL` because `stderr` is FALSE. We would only attempt to parse non-`NULL` results with `jsonlite::fromJSON(inspect)`, with the expectation that all non-`NULL` results are successful results. (l

However, on Windows, it returns `character(0)`, which causes an error `jsonlite::fromJSON(inspect)` to err. Screenshots demonstrating this behavior in a Windows VM are attached at the bottom of this note.

I solved this by wrapping the `inspect` and `jsonlite::fromJSON()` calls in `tryCatch()`. Strictly, only the latter of those needs to be treated like this. At the time I wrote that, I wasn't 100% confident in localizing the error to `jsonlite::fromJSON`, and it *does*, I think, makes sense to capture any errors that `system2` would produce and ignore them.

### Automated Tests

This PR enables more automated tests to run in CI on various operating systems, surfacing errors like this that I wouldn't have seen developing on macOS.

### QA Notes

My laptop is on low battery so these might be a bit skimpy. This PR does touch Quarto-related deployment code, so we should make sure the following work:

- [x] Deploy Quarto-compatible content to a Connect instance:

```r
rsconnect::deployApp("tests/testthat/quarto-proj-r-shiny", quarto = quarto::quarto_path())
rsconnect::deployApp("tests/testthat/quarto-doc-none", quarto = quarto::quarto_path())

```

- [ ] Deploy Quarto non-content to a Connect instance:

```r
rsconnect::deployApp("tests/testthat/shinyapp-simple", quarto = quarto::quarto_path())
```

### Other

- I updated NEWS with this bugfix.

### Screenshots of the issue on Windows

Running `quarto inspect` on Quarto content

Screen Shot 2022-07-13 at 04 39 32 PM@2x

The same, but parsing the output:

Screen Shot 2022-07-13 at 04 41 48 PM@2x

Running `quarto inspect` on non-Quarto content. Note it returns `character(0)` instead of `NULL` as it does on Unix-alikes.

Screen Shot 2022-07-13 at 04 40 27 PM@2x

The same, but parsing the output. This is the bug we were seeing. This would be skipped on Unix in the previous build because `system2` returns `NULL`, which we checked for.

Screen Shot 2022-07-13 at 04 42 21 PM@2x

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -925,17 +925,23 @@ explodeFiles <- function(dir, files) {
 quartoInspect <- function(appDir = NULL, appPrimaryDoc = NULL, quarto = NULL) {
   if (!is.null(quarto)) {
     inspect <- NULL
+    print("Running quarto inspect")
+    print(paste("appDir:", appDir))
+    print(paste("appPrimaryDoc:", appPrimaryDoc))
     # If "quarto inspect appDir" fails, we will try "quarto inspect
     # appPrimaryDoc", so that we can support single files as well as projects.
     primaryDocPath <- file.path(appDir, appPrimaryDoc) # prior art: appHasParameters()
     for (path in c(appDir, primaryDocPath)) {
       args <- c("inspect", path.expand(path))
       inspect <- suppressWarnings(system2(quarto, args, stdout = TRUE, stderr = FALSE))
+      cat(inspect)
       if (is.null(attr(inspect, "status"))) {
+        print("Returning inspect")
         return(jsonlite::fromJSON(inspect))
       }
     }
   }
+  print("Returning NULL")
   return(NULL)
 }
 
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -936,11 +936,12 @@ quartoInspect <- function(appDir = NULL, appPrimaryDoc = NULL, quarto = NULL) {
       {
         inspectOutput <- suppressWarnings(system2(quarto, args, stdout = TRUE, stderr = TRUE))
         inspect <- jsonlite::fromJSON(inspectOutput)
-        return(inspect)
       },
       error = function(e) e
     )
+    if (!is.null(inspect)) break
   }
+  return(inspect)
 }
 
 # Attempt to gather Quarto version and engines, first from quarto inspect if a
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -923,26 +923,24 @@ explodeFiles <- function(dir, files) {
 
 # Run "quarto inspect" on the target and returns its output as a parsed object.
 quartoInspect <- function(appDir = NULL, appPrimaryDoc = NULL, quarto = NULL) {
-  if (!is.null(quarto)) {
-    inspect <- NULL
-    print("Running quarto inspect")
-    print(paste("appDir:", appDir))
-    print(paste("appPrimaryDoc:", appPrimaryDoc))
-    # If "quarto inspect appDir" fails, we will try "quarto inspect
-    # appPrimaryDoc", so that we can support single files as well as projects.
-    primaryDocPath <- file.path(appDir, appPrimaryDoc) # prior art: appHasParameters()
-    for (path in c(appDir, primaryDocPath)) {
-      args <- c("inspect", path.expand(path))
-      inspect <- suppressWarnings(system2(quarto, args, stdout = TRUE, stderr = FALSE))
-      cat(inspect)
-      if (is.null(attr(inspect, "status"))) {
-        print("Returning inspect")
-        return(jsonlite::fromJSON(inspect))
-      }
-    }
+  if (is.null(quarto)) {
+    return(NULL)
   }
-  print("Returning NULL")
-  return(NULL)
+  inspect <- NULL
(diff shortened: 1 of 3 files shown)


R/bundle.R
```diff
@@ -923,20 +923,24 @@ explodeFiles <- function(dir, files) {
...
+  primaryDocPath <- file.path(appDir, appPrimaryDoc) # prior art: appHasParameters()
+  for (path in c(appDir, primaryDocPath)) {
+    args <- c("inspect", path.expand(path))
+    tryCatch(
+      {
+        inspectOutput <- suppressWarnings(system2(quarto, args, stdout = TRUE, stderr = TRUE))
+        inspect <- jsonlite::fromJSON(inspectOutput)
+        return(inspect)
+      },
+      error = function(e) e
```
aronatkins, line 941: What value is returned by this function when both inspect attempts err?

aronatkins (commented):
Still a draft?

toph-allen:
> Still a draft?

Dangit, I *always* forget that.

R/bundle.R
toph-allen: It should be `NULL`.

The function was originally written with the pattern of (pseudocode):

```
inspect <- NULL
inspect <- quarto inspect
return(inspect)
```

Checking the `appDir` and then the `appPrimaryDoc` in sequence, I added `return(inspect)` inside the loop so that we didn't call `quarto inspect` when we didn't need to, because it's expensive. I think I should change the main portion of the function to either (1) continue to return `inspect` within the loop, and return `NULL` at the end, or (2) assign `inspect <- NULL` at the start, break out of the loop if `inspect` is found, return `inspect` at the end.

Which do you think would be better? Or do you think there's a clearer approach I haven't considered?

toph-allen: From looking around the codebase, I think (2).

aronatkins (approved):

ChaitaC:
Installed test branch 
```
devtools::install_github("rstudio/rsconnect","toph-quarto-in-ci")```
rsconnect::deployApp("/Users/chaitamacpro/Desktop/quartoPUPROJECT", server="test_box", account= "chaitatest", quarto = quarto::quarto_path())
```

Deployed without any issue 
Screen Shot 2022-08-05 at 9 56 17 AM

ChaitaC:
The project has app.R 

```
> rsconnect::deployApp("/Users/chaitamacpro/Desktop/21825final", server="test_box", account= "chaitatest", quarto = quarto::quarto_path())
Error in lint(appDir, appFiles, appPrimaryDoc) : 
  Cancelling deployment: invalid project layout.
The project should have one of the following layouts:
1. 'server.R' and 'ui.R' in the application base directory,
2. 'server.R' and 'www/index.html' in the application base directory,
3. 'app.R' or a single-file Shiny .R file,
4. An R Markdown (.Rmd) or Quarto (.qmd) document,
5. A static HTML (.html) or PDF (.pdf) document.
6. 'plumber.R' API description .R file
7. 'entrypoint.R' plumber startup script
8. A tensorflow saved model
```

ChaitaC:
The document was in a sub-directory so IDE threw the above error: 
It passed once the correct path was populated
```
rsconnect::deployApp("/Users/chaitamacpro/Desktop/shiny", appName= "quartormd", server="test_box", account= "chaitatest", quarto = "/usr/local/bin/quarto")
```


We identified a possible enhancement ...when trying to deploy RMD as a quarto app, but the path is incorrect (`quarto = "/usr/local/bin/quarto123"`) then we ignore the `quarto` function and deploy the app as RMD instead of `quarto document` in this specific case we can give a better error message. 
We have a ticket on board 
Screen Shot 2022-08-05 at 1 21 23 PM

PR #612, opened 2022-08-26
Adds support for directly opening jumpstart examples from R

sellorm:
The jumpstart examples in RStudio Connect currently only offer the option to download to the user's desktop via their web browser.

This makes life difficult for users or RStudio Workbench as they must then upload the zip file through the RSW UI and figure out how to unpack it either from R or from the terminal before they can open the project.

This PR adds a helper function that downloads, unpacks and opens a jumpstart example, all from a single function call in R.

RSW users could use this function to more easily obtain the jumpstart examples, which in turn may lead to faster user onboarding. It also helps Connect play better with RSW.

If this PR, or something like it, is accepted I can do something similar on the Python side.

Once both of those pieces are in place, the Connect UI could be updated to point folks to these helper functions, or even just pre-populate code snippets on the download page for the specific jumpstart example that users could copy and paste. 

Thanks for your consideration!


aronatkins (commented):
I like this idea a lot, but find myself wishing that we were recommending these steps in the jumpstart workflow.

Suggesting, for example:

From the R console:

```r
curl::curl_download(
  "https://connect.company.com/__api__/v1/experimental/examples/stock-report/zip", 
  "stock-report.zip")
utils::unzip("stock-report.zip")
rstudioapi::openProject("stock-report", newSession = TRUE)
```

From a terminal:

```bash
curl -JO https://connect.company.com/__api__/v1/experimental/examples/stock-report/zip
unzip stock-report.zip
# Open the stock-report directory.
```

Or download the linked archive (using today's instructions).

We should (within Connect) shift the examples API to be a public, supported API. I would prefer if that work would happen before we start building upon it.

CC @kellobri

sellorm:
Thanks for the feedback, @aronatkins - there are definitely things I missed, or things that could be greatly improved upon here.

My main motivation was really to try and move the needle on improving the integration between Connect and Workbench. I'm not at all precious about this idea specifically and would be more than happy with tweaks to the Connect UI instead.

I'm also happy to work on improving this PR, if the team decide it's worth pursuing.

hadley:
Turning into an issue for now; this would still be useful to have, but we're unlikely to work on it for this release.

PR #614, opened 2022-09-01
Rebranding

whimsicallyson:
Updates references to RStudio and RStudio Connect for the rebranding effort.

Links that still need to be updated:
- https://www.rstudio.com/products/connect/

(updated link list, only base rstudio URLs are being migrated at launch)

Issue #204, opened 2017-09-13
deployAPI happily deploys to shinyapps.io despite lack of support


jspiewak:
#182 added `deployAPI` but doesn't prevent deploying to shinyapps.io which does not support this type of application.

jspiewak:
@scottmmjackson @aronatkins

scottmmjackson:
Here's what I get: (takes 5 min or so), can be reproduced with my silly plumber test app.

```
> rsconnect::deployAPI('~/myApi', account="scottmmjackson")
Preparing to deploy api...DONE
Uploading bundle for api: 214370...DONE
Deploying bundle: 980413 for api: 214370 ...
Waiting for task: 488228839
  building: Processing bundle: 980413
  building: Parsing manifest
  building: Building image: 977515
  building: Fetching packages
  building: Building package: rlang
  building: Building package: tibble
  building: Building package: plumber
  building: Building package: dplyr
  building: Building package: tidyr
  building: Installing files
  building: Pushing image: 977515
  deploying: Starting instances
################################ Begin Task Log ################################ 
<snip>
################################# End Task Log ################################# 
 Show Traceback
 
 Rerun with Debug
 Error: Unhandled Exception: Child Task 488229131 error: Unhandled Exception: Timeout waiting for '_check_container_alive' after 0:00:30 
```

PR #621, opened 2022-10-07
More cloud support changes

omar-rs:
This PR contains several small changes to better support publishing to cloud targets (shinyapps.io and rstudio.cloud).  

* Renamed `isShinyapps` to `isCloudServer` to better reflect the meaning of cloud targets
* Changed `inferAppMode` to return `rmd-shiny` instead of `rmd-static` for cloud targets.  This coercion is intended to be in place until we have a better support for static content in cloud.
* Changed `getPythonForTarget` to have `pythonEnabled` be `true` by default, except for the case of shinyapps.io.
* Changed `deployApp` to allow `contentCategory=api` to work for rstudio.cloud, but not shinyapps.io

### DIFF
--- a/R/accounts.R
+++ b/R/accounts.R
@@ -484,8 +484,12 @@ resolveAccount <- function(account, server = NULL) {
   }
 }
 
-isShinyapps <- function(server) {
+isCloudServer <- function(server) {
   identical(server, "shinyapps.io") | identical(server, "rstudio.cloud")
+}
+
+isShinyappsServer <- function(server) {
+  identical(server, "shinyapps.io")
 }
 
 isRPubs <- function(server) {
@@ -494,7 +498,7 @@ isRPubs <- function(server) {
 
 isConnectInfo <- function(accountInfo = NULL, server = NULL) {
   host <- if (is.null(accountInfo)) server else accountInfo$server
-  !isShinyapps(host) && !isRPubs(host)
+  !isCloudServer(host) && !isRPubs(host)
 }
 
 stopWithNoAccount <- function() {
--- a/R/accounts.R
+++ b/R/accounts.R
@@ -485,7 +485,7 @@ resolveAccount <- function(account, server = NULL) {
 }
 
 isCloudServer <- function(server) {
-  identical(server, "shinyapps.io") | identical(server, "rstudio.cloud")
+  identical(server, "shinyapps.io") || identical(server, "rstudio.cloud")
 }
 
 isShinyappsServer <- function(server) {
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -387,7 +387,7 @@ isShinyRmd <- function(filename) {
 
 # infer the mode of the application from its layout
 # unless we're an API, in which case, we're API mode.
-inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo) {
+inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo, isCloudServer = FALSE) {
   # plumber API
   plumberFiles <- grep("^(plumber|entrypoint).r$", files, ignore.case = TRUE, perl = TRUE)
   if (length(plumberFiles) > 0) {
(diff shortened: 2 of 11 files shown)


R/bundle.R
```diff
@@ -456,6 +456,12 @@ inferAppMode <- function(appDir, appPrimaryDoc, files, quartoInfo) {
     if (!is.null(quartoInfo)) {
       return("quarto-static")
     } else {
+      # For Shinyapps and rstudio.cloud, treat "rmd-static" app mode as "rmd-shiny" so that
+      # they can be served from a shiny process in Connect until we have better support of
+      # rmarkdown static content
+      if (isCloudServer) {
+        return("rmd-shiny")
```
mbaynton, line 463: Actually tested by running this locally and getting an `rmd-static` output to deploy successfully?

mbaynton (approved):
Nice! :rocket:

R/accounts.R
```diff
@@ -484,17 +484,21 @@ resolveAccount <- function(account, server = NULL) {
   }
 }
 
-isShinyapps <- function(server) {
+isCloudServer <- function(server) {
   identical(server, "shinyapps.io") | identical(server, "rstudio.cloud")
```
kippandrew, line 488: Should we prefetch the rebranding work and include "posit.cloud" here?

R/bundle.R
omar-rs: Yes, tested and verified

R/accounts.R
omar-rs: There is a separate PR for rebranding: https://github.com/rstudio/rsconnect/pull/614

ssinnott:
I think there's another location of of `inferAppMode` over here,

https://github.com/rstudio/rsconnect/blob/a0d843fa06cb0286b3f7538ee051c3ba1ebdaf7f/R/writeManifest.R#L67-L71

R/deployApp.R
```diff
@@ -336,12 +336,13 @@ deployApp <- function(appDir = getwd(),
   accountDetails <- accountInfo(target$account, target$server)
 
   # test for compatibility between account type and publish intent
-  if (isShinyapps(accountDetails$server)) {
-    # ensure we aren't trying to publish an API to shinyapps.io; this will not
-    # currently end well
-    if (identical(contentCategory, "api")) {
-      stop("Plumber APIs are not currently supported on shinyapps.io; they ",
-           "can only be published to RStudio Connect.")
+  if (isCloudServer(accountDetails$server)) {
```
ssinnott, line 339: This can probably just check shinyapps directly?

```
if (isShinyappsServer(accountDetails$server)) {
  if (identical(contentCategory, "api")) {
    stop("Plumber APIs are not currently supported on shinyapps.io; they ",
         "can only be published to RStudio Connect or rstudio.cloud.")
  }
}
```

ssinnott:
Have a couple tweaks but overall I think this looks about right.

ssinnott (approved):

R/accounts.R
aronatkins: Nit: You wanted `||` here, rather than `|`. The two-character form always returns a length-one vector.
https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/Logic

R/deployApp.R
aronatkins: We should never be relying a `contentCategory` of `api`. I thought we had ditched that.

It was removed with https://github.com/rstudio/rsconnect/pull/304

Unfortunately, it returned with the fix to https://github.com/rstudio/rsconnect/issues/204 (https://github.com/rstudio/rsconnect/commit/9e768e87c9a63885f175ccbc4ed69eb6c503f0f2; no PR?)

~This code is after we call `inferAppMode` and can use `appMode` rather than `contentCategory`.~

We do not want `deployAPI` to set `contentCategory`.

Correction: This check is happening ahead of the call to `bundleApp`, which means that we have not yet determined the `appMode`. That's unfortunate.

aronatkins (approved):
Could we adjust the error handling around APIs to stop setting `contentCategory` in `deployAPI`? That can happen in a separate PR.

omar-rs:
> I think there's another location of of `inferAppMode` over here,

It looked like `writeManifest` is a top level API that is not relevant for the cloud publishing use case, so I left it alone.

omar-rs:
> Correction: This check is happening ahead of the call to bundleApp, which means that we have not yet determined the appMode. That's unfortunate.

@aronatkins looks like this is going to be a larger refactor to address your concern here, so I can create a separate issue to track that.

aronatkins:
> so I can create a separate issue to track that.

@omar-rs - thanks. i agree that the problem with content category is best addressed outside this PR.

toph-allen (approved):

toph-allen:
@omar-rs @aronatkins Yeah, the order of operations for gathering info in `bundleApp` and `writeManifest` is kind of a Gordian knot right now.

PR #622, opened 2022-10-11
escape an account name before performing recursive directory search

### LINKED ISSUE #620: publishing credentials files with a + in the name are not usable
If i created a publishing credential with a `+` in the filename, the file cannot be found when trying to use it.

```
> rsconnect::deployApp("myapp")
Error in accountInfo(accounts) : 
  account named 'myemail+test@rstudio.com' does not exist
```

Pretty sure its because of this code:

https://github.com/rstudio/rsconnect/blob/8c98105575c4823606db93a5aa23944765529dcc/R/accounts.R#L441-L451

`list-files` treats `pattern` as a regular expression where + has special meaning. If the + is double escaped, then it would find the file.

aronatkins:
fixes #620

@samperman - would you mind trying this?

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -253,3 +253,11 @@ isAvailable <- function(package, min_version = NULL) {
   # the package has at least that version.
   return(utils::packageVersion(package) >= min_version)
 }
+
+# Escape characters that have special meaning for extended regular expressions as supported by
+# list.files.
+#
+# Similar to rex::escape
+escapeRegex <- function(name) {
+  gsub("([.|()\\^{}+$*?\\[\\]])", "\\\\\\1", name, perl = TRUE)
+}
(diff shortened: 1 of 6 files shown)


samperman (approved):
LGTM. i tested two projects, one using rsconnect from `master` and one using this branch. Each had a single `.dcf` file with a `+` in it. The project using `master` failed while the project using this branched worked great!

R/utils.R
```diff
@@ -253,3 +253,8 @@ isAvailable <- function(package, min_version = NULL) {
   # the package has at least that version.
   return(utils::packageVersion(package) >= min_version)
 }
+
+# similar to rex::escape
```
toph-allen, line 257: Maybe this comment would be better if it described what `escapeRegex` does in a sentence?

toph-allen (approved):
Looks good to me. Didn't install & test because I haven't rebuilt my R environments.

PR #625, opened 2022-10-14
ragg is no longer an implicit dependency of Shiny content

aronatkins:
Reverts #598


toph-allen (approved):
LGTM

wch (approved):

aronatkins:
Confirmed that `ragg` is no longer automatically included in the generated set of dependencies, but present with an explicit `library(ragg)` statement.

PR #627, opened 2022-10-18
Add "type" query filter when listing apps on cloud or shinyapps

omar-rs:
This fixes an issue with cloud support whereby publishing a Shiny App to rstudio.cloud with the same name as an existing app on shinyapps.io results in the app getting published to shinyapps.io again.

This is because the call to `listApplications` on the `cloudClient` does not currently limit the apps to cloud applications (`type=connect`), so it returns the matching shiny application as the first result. 

The change is to include a `type` filter in the `listApplications` function.  The type value used is `connect` or `shiny`, depending on the client.

### DIFF
--- a/R/lucid.R
+++ b/R/lucid.R
@@ -74,8 +74,8 @@ shinyAppsClient <- function(service, authInfo) {
    listApplications = function(accountId, filters = list()) {
       path <- "/applications/"
       query <- paste(filterQuery(
-        c("account_id", names(filters)),
-        c(accountId, unname(filters))
+        c("account_id", "type", names(filters)),
+        c(accountId, "shiny", unname(filters))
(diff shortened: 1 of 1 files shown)


omar-rs:
### Steps to verify
1. Publish an app on shinyapps.io, name it "shinytest"
2. Create a shiny app in a project in Cloud and push the Publish button
3. Give the app the same name ("shinytest") as the one you published earlier to shinyapps.io
4. Verify that the app is deployed correctly to Cloud

### Other verifications
* deploy application to shinyapps.io via publish button and via rsconnect::deployApp
* update application in shinyapps.io via publish button and via rsconnect::deployApp

aronatkins (approved):

PR #628, opened 2022-10-18
separate the HTTP tests into per-protocol checks with shared helpers

aronatkins:


toph-allen (approved):
At the very least we should be able to track stuff down by line number!

PR #629, opened 2022-10-18
Include rstudio.cloud in servers() list

mbaynton:
Required for IDE validation when re-deploying an already deployed output.

### DIFF
--- a/R/servers.R
+++ b/R/servers.R
@@ -11,8 +11,7 @@
 #' The `servers` and `serverInfo` functions are provided for viewing
 #' previously registered servers.
 #'
-#' There is always at least one server registered (the `shinyapps.io`
-#' server)
+#' Servers for `shinyapps.io` and `rstudio.cloud` are always registered.
 #'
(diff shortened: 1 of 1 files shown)


R/servers.R
```diff
@@ -63,7 +63,8 @@ servers <- function(local = FALSE) {
   if (local) {
     locals
   } else {
-    rbind(locals, as.data.frame(shinyappsServerInfo(), stringsAsFactors = FALSE))
+    locals <- rbind(locals, as.data.frame(shinyappsServerInfo(), stringsAsFactors = FALSE))
```
aronatkins, line 66: I think you also could have:

```r
rbind(
    locals,
    as.data.frame(shinyappsServerInfo(), stringsAsFactors = FALSE),
    as.data.frame(cloudServerInfo(), stringsAsFactors = FALSE))
```

aronatkins, line 66: Please update the docs for `servers`, which indicates that there is always a single shinyapps.io entry; we should do something similar for cloud.

aronatkins (approved):

PR #634, opened 2022-11-07
quick rebranding pass (RStudio => Posit)

aronatkins:
Replaces #614

At least one minor change was because we had stale docs on `main`.


toph-allen (approved):

whimsicallyson (approved):
yup, you got all the spots I saw!

PR #637, opened 2022-11-29
Rebrand rstudio.cloud -> posit.cloud

### LINKED ISSUE #636: [posit cloud] rebranding
Update the built-in server info for `rstudio.cloud` to `posit.cloud`.

Ensure that outputs published to posit cloud using existing versions of rsconnect, and accounts configured on local machines outside of cloud using the `rstudio.cloud` server identifier, continue to work.

mbaynton:
Fixes #636 

This is a bit more than find/replace in order to support accounts and applications that were already published under the `rstudio.cloud` name; lots of logic wants the server name on the account or publication history to match this. We also need to simultaneously support `setAccountInfo()` using both product names for a while to give this release time to be adopted. At a future point we'll update the cloud UI's code snippet to use `server = posit.cloud`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## 0.8.29 (in development)
 
-* Update company and product name for rebranding to Posit.
+* Update company and product names for rebranding to Posit.
 
 ## 0.8.28
 
--- a/R/auth.R
+++ b/R/auth.R
@@ -147,7 +147,7 @@ showUsers <- function(appDir=getwd(), appName=NULL, account = NULL,
   accountDetails <- accountInfo(resolveAccount(account, server), server)
 
   if (!isCloudServer(accountDetails$server)) {
-    stop("This method only works for ShinyApps or rstudio.cloud servers.")
+    stop("This method only works for ShinyApps or posit.cloud servers.")
   }
 
   # resolve application
--- a/R/accounts.R
+++ b/R/accounts.R
@@ -246,8 +246,8 @@ setAccountInfo <- function(name, token, secret,
     stop(stringParamErrorMessage("secret"))
 
   # create connect client
-  if (identical(server, cloudServerInfo()$name)) {
-    serverInfo <- cloudServerInfo()
+  if (identical(server, cloudServerInfo(server)$name)) {
+    serverInfo <- cloudServerInfo(server)
   } else {
     serverInfo <- shinyappsServerInfo()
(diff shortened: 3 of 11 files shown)


R/auth.R
```diff
@@ -147,7 +147,7 @@ showUsers <- function(appDir=getwd(), appName=NULL, account = NULL,
   accountDetails <- accountInfo(resolveAccount(account, server), server)
 
   if (!isCloudServer(accountDetails$server)) {
-    stop("This method only works for ShinyApps or rstudio.cloud servers.")
+    stop("This method only works for ShinyApps or posit.cloud servers.")
```
omar-rs, line 150: should we use `shinyapps.io` instead of `ShinyApps`?  we use `shinyapps.io` in other places

omar-rs (approved):

NEWS.md
```diff
@@ -2,7 +2,7 @@
 
 ## 0.8.29 (in development)
 
-* Update company and product name for rebranding to Posit.
+* Update company and product names for rebranding to Posit.
```
aronatkins, line 5: Let's add an additional NEWS entry for this PR, since we're rewriting hostnames. That's a functional change, and different from some of the earlier work.

aronatkins (commented):
Are deployment records being rewritten? Should they be rewritten? Could we end up with two deployment records (using old and new hosts) which really should be the same?

R/auth.R
mbaynton: Good point, these names are incorrect and not consistent with each other. I'll tweak.

mbaynton:
> Are deployment records being rewritten?

No

> Should they be rewritten?

Debatable, but I don't think it's important, because

> Could we end up with two deployment records (using old and new hosts) which really should be the same?

No. The approach this change takes is to try to stick to calling the server `rstudio.cloud` if the user previously set themselves up with that name. Given very low adoption of cloud publishing / very few real-world users who might have done this, I think this is a good enough solution.

R/auth.R
mbaynton: I'm just going to revert [87f4f63](https://github.com/rstudio/rsconnect/pull/637/commits/87f4f63facef9c3eb24d11e20022d059c7e81194) to keep this PR focused. We can answer the above question elsewhere.

aronatkins (approved):
Merge once your team has done verification.

PR #657, opened 2023-01-26
Standardize Posit name

jonthegeek:
I scraped CRAN to find companies that fund R packages, and noticed that Posit is credited 6 different ways (9 if you include RStudio/RStudio PBC/RStudio, PBC). I'm submitting PRs with the official "Posit Software, PBC" name on the few Posit varieties to nip it in the bud.


whimsicallyson (approved):
Yes this is the legal name of the company, I imagine that's what we'll want throughout

PR #660, opened 2023-02-09
Simplify temporary options setting

hadley:
`options()` invisibly returns a list of the previous values, which you can use to reset to the previous state.


toph-allen (approved):
Much better. Thanks.

PR #661, opened 2023-02-09
Update to testthat 3e

hadley:
I tried to keep the tests as similar as possible but I needed to do some minor tweaking to get some of the tests to run at the command line so I could diagnose problems.

### DIFF
--- a/tests/testthat/test-http.R
+++ b/tests/testthat/test-http.R
@@ -1,5 +1,3 @@
-context("http")
-
 # Test HTTP server doesn't work on Solaris
 skip_on_os("solaris")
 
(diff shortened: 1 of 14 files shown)


tests/testthat/test-http.R
```diff
@@ -62,7 +60,7 @@ setup({
   Sys.sleep(2)
 })
 
-teardown({
+withr::defer({
```
aronatkins, line 63: When is this `withr::defer` triggered? Does `testthat` have an environment for each test file, meaning the `defer` runs as the file completes?

aronatkins (approved):

tests/testthat/test-http.R
hadley: At the end of the file, and yes 😄

PR #662, opened 2023-02-12
Prepare for refactoring on `deploymentTarget()`

hadley:
I've identified that `deploymentTarget()` is a large source of frustration for me, because when I misspecify the function arguments, it never gives me _quite_ enough information to immediately resolve the problem. This PR does not improve the errors, but prepares for a future PR that will do so. In this PR:

* I've moved `deploymentTarget()` to its own file, so it can have it's own test
* Extracted out `userNameFromDeployment()` and `createDeployment()` target into standalone functions
* Extract `findAppId()` out of `createDeploymentTarget()`.
* Added some mocking infrastructure
* Got to 100% test coverage of the main branches of `deploymentTarget()` (I have not yet tested the `findAppId()` path since that felt a bit peripheral to this PR)
* Added a bunch of TODOs that I would like to resolve in future PRs.

I've taken care to not change any behaviour in this PR, so any review can be more about the clarity of the tests and the mocking infrastructure. (Writing these tests have revealed a few annoyances in our mocking tools, which I'll tackle in the weeks to come.)


tests/testthat/test-deploymentTarget.R
```diff
@@ -0,0 +1,159 @@
...
+  )
+
+  target <- deploymentTarget(app_dir, appId = "123")
+  expect_equal(target$appId, "123")
+  expect_equal(target$appTitle, "my title")
+
+  target <- deploymentTarget(app_dir, appName = "test", appId = "123")
+  expect_equal(target$appId, "123")
+  # TODO(HW): why doesn't this equal my title?
+  # expect_equal(target$appTitle, "my title")
```
aronatkins, line 139: I think this clause might be the cause (version before your PR):
https://github.com/rstudio/rsconnect/blob/ce49da6c57599137b8fe1c41c609ed4ad5e82e56/R/deployApp.R#L694-L699

We have a target deployment, but aren't using its stored data.

aronatkins (approved):

tests/testthat/test-deploymentTarget.R
hadley: Yeah, looks like I added a matching TODO here: https://github.com/rstudio/rsconnect/pull/662/files#diff-7284d883e7820ca83c32ae955a374305739a8841be7d9ec6d5860cbe2000e903R82

aronatkins: It appears that we don't have any way to perform a title "update", so yes -- the stored title overrides any incoming user-provided title. The `connect.R` and `lucid.R` only define `createApplication` and don't offer a way to update `title`.

There are probably lurking inconsistencies, however. The `application` returned by `applicationForTarget` is conditionally fetched from the server, and I cannot see where its `title` (Connect-only) is written back to the deployment record.

hadley: When I refactor this, I'll have a go at extracting out the "find existing application" in it's own logic to make it all more consistent.

PR #663, opened 2023-02-12
Use verboseLogger() consistently

hadley:
And fix up some minor spacing issues


aronatkins (approved):
Thanks! I added `verboseLogger` ages ago and should have returned to update the other "when verbose" messages.

PR #664, opened 2023-02-12
Prepare `bundlePackages()` for refactoring

hadley:
Similar idea to #662, but I think this needs some discussion before I proceed further because I don't understand the logic.


tests/testthat/_snaps/bundlePackage.md
```diff
@@ -0,0 +1,11 @@
+# errors if dependencies aren't installed
+
+    Code
+      bundlePackages(app_dir, appMode = "rmd-static", assetTypeName = "asset")
+    Warning <simpleWarning>
+      Package 'doesntexist1' not available in repository or locally
+      Package 'doesntexist2' not available in repository or locally
+    Error <simpleError>
+      Unable to retrieve package records for the following packages:
+      - 'doesntexist1', 'doesntexist2'
```
aronatkins, line 10: It looks as if the Windows test failures are due to differences in output quoting.

This output comes from Packrat and is sensitive to `shQuote`:

https://github.com/rstudio/packrat/blob/9917985132f2dce44ed1ef9d3a5aa7679a7c411c/R/pkg.R#L259-L263

aronatkins (approved):
I'm OK with this once CI passes.

R/bundlePackage.R
```diff
@@ -0,0 +1,157 @@
...
+
+    # get package description; note that we need to remove the
+    # packageDescription S3 class from the object or jsonlite will refuse to
+    # serialize it when building the manifest JSON
+    # TODO: should we get description from packrat/desc folder?
+    info$description <- suppressWarnings(unclass(utils::packageDescription(name)))
+
+    # TODO(HW): should this code be removed? I don't see how it is ever reached
+    # if description is NA, application dependency may not be installed
+    if (is.na(info$description[1])) {
```
hadley, line 51: Do you understand how we hit this code? Is it if there was a previous snapshot and those packages are no longer installed?

tests/testthat/_snaps/bundlePackage.md
hadley: Ok, that's an easy fix.

R/bundlePackage.R
hadley: Ah got it, I'll change my test.

R/bundlePackage.R
```diff
@@ -0,0 +1,157 @@
...
+}
+
+## check for extra dependencies congruent to application mode
+inferRPackageDependencies <- function(appMode,
+                                      hasParameters,
+                                      documentsHavePython,
+                                      quartoInfo) {
+
+  # TODO(HW): appUsesR(quartoInfo) is always TRUE because this function is
+  # only called for apps that use R
```
toph-allen, line 110: I think it's perhaps still better to leave that call on line 120 than to remove it (assuming I understand what you're talking about). Although it will always be `TRUE` where the function is called within `bundlePackages`, I'd rather have that unnecessary call than have this function return incorrect R dependencies for Quarto content that doesn't use R.

toph-allen (approved):

R/bundlePackage.R
hadley: My feeling was that the name of the function already gives a pretty strong signal — why would you be calling `inferRPackageDependencies` for content that doesn't use R? But if you want this to be robust against future incorrect usage, I could error if `!appUsesR.

aronatkins: The call to this function is now guarded by `appUsesR`; I'm OK making that an expectation, and leave it to the caller to understand what information needs to be known before calling. By combining the call to `inferredRDependencies` and `shapshotRDependencies`, we need to look at `appUsesR` much less.

R/bundlePackage.R
hadley: I've failed to create this situation in a test, so I'll just do some mocking to simulate it.

PR #665, opened 2023-02-13
Use rlang to check if required package versions are installed

hadley:
Obviously, taking a dependency on rlang is not worth for just this case, but using rlang will allow me to simplify and clarify a bunch of other code in future PRs.


R/deployDoc.R
```diff
@@ -23,15 +23,16 @@
 #' @family Deployment functions
 #' @export
 deployDoc <- function(doc, ...) {
+  check_installed(
```
aronatkins, line 26: This is a functional change, since it will prompt interactive users to install the packages rather than immediately stopping; should we announce it in NEWS.md?

aronatkins (approved):

R/deployDoc.R
hadley: Ooops, missed this. I'll add momentarily.

PR #666, opened 2023-02-13
Simply `deploymentTarget()` logic

hadley:


R/deploymentTarget.R
```diff
@@ -1,210 +1,66 @@
...
+      appName <- generateAppName(appTitle, appPath, account, unique = FALSE)
     }
-  }
-
-  # read existing deployments
-  appDeployments <- deployments(appPath = appPath)
-
-
-  # if appTitle specified but not appName, generate name from title
-  if (is.null(appName) && !is.null(appTitle) && nzchar(appTitle)) {
```
hadley, line 33: This is now only done on first deploy as documented. And I'm pretty sure the previous behaviour was incorrect as it meant that setting a new `appTitle` would generally result in publishing a new app, rather than updating an existing app.

R/deployments.R
```diff
@@ -11,28 +11,6 @@ saveDeployment <- function(appPath,
                            url,
                            metadata = list()) {
 
-  # TODO(HW): move this logic into deploymentTarget
-  # if there's no new title specified, load the existing deployment record, if
-  # any, to preserve the old title
-  if (is.null(title) || is.na(title) || length(title) == 0 ||
```
hadley, line 17: `saveDeployment()` is only called `deployApp()` where `title` is the `appTitle` returned by `deploymentTitle()`, so I moved the logic there.

R/deployApp.R
```diff
@@ -4,6 +4,20 @@
 #' [RMarkdown][rmarkdown::rmarkdown-package] document, a plumber API, or HTML
 #' content to a server.
 #'
+#' ## Updating existing apps
+#'
+#' If you have previously deployed an app, `deployApp()` will do its best to
+#' update the existing deployment. In the simple case where you have only on
```
aronatkins, line 10: "only one", not "only on"

R/deploymentTarget.R
aronatkins: Could you check what values are provided by the IDE? Its conventions have changed over time. In particular, what happens when going through the "Other Destination" workflow? Is that different from a "republish" workflow?

I don't recall what of name/title is given by the IDE. It would be good to capture that expectation.

We'll probably want to retest a handful of "new deployment" versus "overwrite deployment" workflows. CC @ChaitaC

R/import-standalone-obj-type.R
```diff
@@ -0,0 +1,348 @@
+# Standalone file: do not edit by hand
```
aronatkins, line 1: One of your other PRs is pulling in `rlang`; does this file go away with an `rlang` dependency?

aronatkins (approved):

R/deploymentTarget.R
hadley: Here's what I see in the "Other destination" dialog — it like the UI allows you to choose the `appTitle`, which appears to be auto generated from the doc name:

Screenshot 2023-02-13 at 16 34 47

It's not clear what it users for the `appName`. I'll need to do some more exploration.

The republish dialog only allows me to change which files are published.

R/import-standalone-obj-type.R
hadley: We haven't yet standardised this interface, so we're currently vendoring it into every package where we use it. I think it's worth the cost, because it makes input validation _so_ much more pleasant. At some point once we provide an exported interface, we'll be able to remove from this and other packages.

R/deploymentTarget.R
hadley: Note that the id also uses the output of `applications()` to find the `appId` based on the `appName`, so that even if `deployments()` is empty, it'll still attempt to update an existing bundle.

toph-allen (approved):
Upon a read-through, the changes look good. We should probably make sure we put these changes through QA, but if there's more coming down the pike, maybe it makes sense to bundle them and have @ChaitaC look at them together.

R/acccount-find.R
```diff
@@ -0,0 +1,82 @@
+findAccount <- function(accountName = NULL, server = NULL, error_call = caller_env()) {
```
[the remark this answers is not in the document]
hadley: Oops, and I made that typo in a few other places.

R/accounts.R
```diff
@@ -294,16 +294,10 @@ setAccountInfo <- function(name, token, secret,
...
 
-  if (!isStringParam(name))
-    stop(stringParamErrorMessage("name"))
-
-  configFile <- accountConfigFile(name, server)
-  if (length(configFile) > 1)
-    stopWithMultipleAccounts(name)
-  if (length(configFile) == 0 || !file.exists(configFile))
-    stop(missingAccountErrorMessage(name))
+  fullAccount <- findAccount(name, server)
```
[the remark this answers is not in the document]
hadley: Me too! It also always bugged me that I had to supply both `name` and `server`  even though all the servers I publish to have the same username.

R/tasks.R
```diff
@@ -1,9 +1,6 @@
 #' List Tasks
 #'
-#' @param account Account name. If a single account is registered on the system
-#'   then this parameter can be omitted.
-#' @param server Server name. Required only if you use the same account name on
-#'   multiple servers (see [servers()])
+#' @inheritParams deployApp
```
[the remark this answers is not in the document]
hadley: Doc issue at https://github.com/r-lib/roxygen2/issues/1465, and a better long form explanation at https://r-pkgs.org/man.html#inheriting-arguments

R/deploymentTarget.R
hadley: I've made an issue https://github.com/rstudio/rsconnect/issues/669 to track logging more information so it's easier to verify this. I'll do that in a separate PR because this one is already big.

PR #670, opened 2023-02-14
Refactor `bundlePackages()`

### LINKED ISSUE #659: Warn if package versions are newer than those on CRAN
Fixes the problem I have 

Add something to https://github.com/rstudio/rsconnect/blob/9c93300249327012e3635fd8c339d8f85c5e50bc/R/bundle.R#L694-L695

hadley:
This turns the main strategy "inside out", so that we find all uninstalled packages, then all packages with missing repo, then collect the package metadata into a list.

I've switched the error and warning to use cli and spent a little time polishing the language. But I'm not sure it's that great yet, so I'd love your feedback :)

I also rewrote `inferRPackageDependencies()` to use a much simpler strategy since there can only be one `appMode`.

Fixes #659


R/bundlePackage.R
```diff
@@ -25,135 +25,59 @@ bundlePackages <- function(appDir,
...
-      )
-      next
-    }
-
-    # validate package source (returns a message if there is a problem)
-    packageMessages <- c(packageMessages, validatePackageSource(deps[i, ]))
-
-    # good to go
-    packages[[name]] <- info
+  not_installed <- !vapply(deps$Package, is_installed, logical(1))
```
aronatkins, line 28: It's going to take me time to grok that things like `is_installed` are coming from `rlang` rather than part of this package. All other external references use `packagename::XX` names.

R/bundlePackage.R
```diff
@@ -25,135 +25,59 @@ bundlePackages <- function(appDir,
...
+  if (any(not_installed)) {
+    pkgs <- deps$Package[not_installed]
+    cli::cli_abort(c(
+      "All packages used by {assetTypeName} must be installed.",
```
aronatkins, line 32: I think `assetTypeName` is "site", "content", "document", or "application", which gives "All packages used by application must be installed". Maybe "used by _the_ application" instead?

R/bundlePackage.R
```diff
@@ -25,135 +25,59 @@ bundlePackages <- function(appDir,
...
-        "Unable to determine the source location for some packages. ",
-        "Packages should be installed from a package repository like ",
-        "CRAN or a version control system. Check that ",
-        "options('repos') refers to a package repository containing ",
-        "the needed package versions."
+  missing_repo <- !isSCMSource(deps$Source) & is.na(deps$Repository)
+  if (any(missing_repo)) {
+    pkgs <- deps$Package[missing_repo]
+    cli::cli_warn(c(
+      "Packages require a known repository for install on remote system.",
```
aronatkins, line 41: Perhaps: "for installation on a remote system"?

R/bundlePackage.R
```diff
@@ -25,135 +25,59 @@ bundlePackages <- function(appDir,
...
+      x = "Packages with unknown repository: {.pkg {pkgs}}.",
+      i = paste0(
+        "Automatic source detection relies on packages being installed from ",
+        "a standard repository like CRAN or BioConductor, or from a version ",
+        "control system like GitHub or GitLab."
```
aronatkins, line 46: I don't know what "Automatic source detection" means. Maybe:

> Packages must be installed from a standard repository like CRAN or BioConductor, or from a version control system like GitHub or GitLab.

It's still not great. We're trying to communicate that the _local_ package installs need to come from a supported location so that the same version of that package can be installed by the _remote_ system. These errors often throw folks off because they forget that there are client-side requirements that need to be met before the remote system is ever involved. We occasionally have folks try to install packages on the remote system rather than adjusting their client library.

Should we enumerate the most common mistakes?

* Installing packages from source.
* No repository (`options('repos')`) containing the target package.

aronatkins (approved):
Wordsmithing is unending.

R/bundlePackage.R
```diff
@@ -25,135 +25,59 @@ bundlePackages <- function(appDir,
...
-    # cause.
-    packageMessages <- c(
-      packageMessages,
-      paste0(
-        "Unable to determine the source location for some packages. ",
-        "Packages should be installed from a package repository like ",
-        "CRAN or a version control system. Check that ",
-        "options('repos') refers to a package repository containing ",
-        "the needed package versions."
+  missing_repo <- !isSCMSource(deps$Source) & is.na(deps$Repository)
```
toph-allen, line 37: Is it worth warning about `missing_repo` packages before aborting due to `not_installed` packages? And is there a reason this is a `warn` rather than an `abort`?

R/bundlePackage.R
toph-allen: Yeah, "source" is ambiguous here, e.g. "the source from which a package was installed" vs short-hand for "source code". This is tricky to communicate succinctly.

Maybe something that gives the user more of a clear instruction for how to fix the problem would read better.

> Cannot determine the installation source for these packages. Try reinstalling from a standard…

toph-allen (approved):

R/bundlePackage.R
hadley: Yeah, our philosophy is that rlang is the sort of missing stdlib, that we've decided that we can't live with out and use so frequently that `::` is high friction.

R/bundlePackage.R
hadley: The reason for the warn, not abort (from @aronatkins), is it's not uncommon for CRAN packages to be installed in such a way that this information is lost (e.g. installed locally or from source). But maybe I can tackle this in pursuit of #659 — I could check the package name against a list of CRAN packages and error if it's not found.

R/bundlePackage.R
```diff
@@ -25,135 +25,59 @@ bundlePackages <- function(appDir,
...
-
-  # construct package list from dependencies
-  for (i in seq_len(nrow(deps))) {
-    name <- deps[i, "Package"]
-
-    # get package info
-    info <- as.list(deps[i, c("Source", "Repository")])
-
-    # include github package info
-    info <- c(info, as.list(deps[i, grep("Github", colnames(deps), perl = TRUE, value = TRUE)]))
```
hadley, line 40: Does this imply that rsconnect doesn't actually support packages installed from GitLab and Bitbucket? Or does connect read the needed metadata from the serialised DESCRIPTION and these extra columns can be dropped?

R/bundlePackage.R
```diff
@@ -24,136 +25,66 @@ bundlePackages <- function(appDir,
...
-      packageMessages,
-      paste0(
-        "Unable to determine the source location for some packages. ",
-        "Packages should be installed from a package repository like ",
-        "CRAN or a version control system. Check that ",
-        "options('repos') refers to a package repository containing ",
-        "the needed package versions."
-      )
+checkBundlePackages <- function(deps, call = caller_env()) {
+  not_installed <- !vapply(deps$Package, is_installed, logical(1))
```
hadley, line 46: This logic still isn't quite right because these errors will occur before `deployApp()`  has called `saveDeployment()`, so if this fails and you're publishing with the IDE, you'll need to click through all of the dialogs again. Not yet sure how to resolve this.

R/bundlePackage.R
hadley: It doesn't seem possible to give enough detail in this error message, so I've gone for a pointer to the docs:

      ! Local packages must be installed from a supported source.
      x Unsupported packages: shiny.
      i Supported sources are CRAN and CRAN-like repositories, BioConductor, GitHub, GitLab, and Bitbucket.
      i See `rsconnect::appDependencies()` for more details.

I'll keep working on the language but this feels like a more fruitful approach.

R/bundlePackage.R
hadley: Can you point me to the code in connect that parses the packages info from the manifest?

R/bundlePackage.R
aronatkins: @toph-allen The additional context is from #511 / #508, which relaxed the decisions in #412

https://github.com/rstudio/rsconnect/blob/3bf50f095e80fc58ea3b966cd40d925e111fdbbc/R/bundlePackage.R#L69-L81

R/bundlePackage.R
aronatkins: I believe that this is consumed by shinyapps.io, not Connect. I'll try to uncover the reference and send it separately.

hadley: Oh, what does connect use? Just the lockfile?

aronatkins: Connect prefers the packrat/packrat.lock. When only a manifest is provided, a lockfile is constructed from the manifest before handing off to packrat for package installation.

hadley: Ok, good to know that this is all for shinyapps.io. I'll definitely add some comments.

aronatkins: At one point in time, there were two packages. https://github.com/rstudio/shinyapps

hadley: And ShinyApps.io only supports GitHub packages, hence the filtering.

toph-allen:
Just a heads up that Tests are failing on R 3.6.3 and R 3.5.3. Not immediately obvious to me from the GH Actions output what's incompatible.

R/bundlePackage.R
hadley: Discussion now captured in comments.

hadley:
Ugh, that was a dumb and rather hard to track down mistake.

hadley:
@aronatkins I think this is worth another review. In summary:

* We better handle my personal motivation: development versions of packages that I've installed locally (#659). We do this by only matching CRAN if the locally-installed package name matches a CRAN package name *and* the local version is not ahead of the CRAN version.
* This gives me more confidence that we are detecting true failure states, so the new `checkBundlePackages()` function errors for both uninstalled packages and packages we don't know how to reinstall. If this causes the initial deployment to fail, you'll lose your work defining the destination, but I plan to fix that in #677.
* Since the failure mode is hard to fully explain in an error message, I have fleshed out the docs in `appDependencies()` and point there from the error.

There is some mild risk that I have introduced bugs that are not caught by the unit tests because this code is hard to test. I think this is somewhat mitigated by the fact that I will be using the dev version of rsconnect, have a mix of CRAN + local + GitHub packages, and will be actively deploying projects over the next few weeks.

toph-allen:
> Ugh, that was a dumb and rather hard to track down mistake.

`stringsAsFactors` strikes again!

R/bundle.R
```diff
@@ -933,50 +937,58 @@ snapshotRDependencies <- function(appDir, implicit_dependencies = c(), verbose =
...
-    stringsAsFactors = FALSE)
-
-  # get packages records defined in the lockfile
-  records <- utils::tail(df, -1)
+    stringsAsFactors = FALSE
+  )
 
-  # if the package is in a named CRAN-like repository capture it
+  # Sources are created by packrat:
+  # https://github.com/rstudio/packrat/blob/main/R/pkg.R#L328
```
aronatkins, line 944: nit: source references against the main branch grow stale as the code evolves; should we link to specific commits or tags?

R/bundlePackage.R
```diff
@@ -24,136 +25,68 @@ bundlePackages <- function(appDir,
...
-    packageMessages <- c(packageMessages, validatePackageSource(deps[i, ]))
+    unclass(utils::packageDescription(nm))
+  })
 
-    # good to go
-    packages[[name]] <- info
-  }
+  # Connect will generally use the packrat/packrat.lock file; the data
+  # in the manifest is used primarily by shinyapps.io, which only supports
+  # Github remotes, not Bitbucket or Gitlab.
```
aronatkins, line 38: shinyapps.io wants Github information in the outer package record and does not consider the description sub-record. Connect generally prefers the packrat.lock, but when it only has a manifest, it uses the GitHub/Remotes fields from the description sub-record.

R/bundle.R
```diff
@@ -933,50 +937,58 @@ snapshotRDependencies <- function(appDir, implicit_dependencies = c(), verbose =
...
-      #
-      # shinyapps.io isn't going to use the manifest-provided CRAN URL,
-      # but other consumers (Connect) will.
-      if (tolower(source) != "cran") {
-        source <- package.repo$name
+      # If the name of a source-installed package matches a CRAN package,
+      # packrat automatically fills in the repository information. But that
+      # causes problems if you're working with a developement version that's
+      # newer than CRAN.
+      repo_version <- package_version(repo.packages[pkg, "Version"])
```
aronatkins, line 961: What defines `package_version`? Could you explain how this will cope with archived packages (e.g. `Kmisc`, from https://github.com/rstudio/rsconnect/issues/508)?

aronatkins (approved):
OK, with testing concerns about archived packages.

R/bundle.R
hadley: `package_version()` is the base R function that takes a string and turns it into a version object so that `1.2.3` compares correctly to `1.10.4`.

Good question about archived packages. I will investigate.

aronatkins: Thanks (on both). I'm more used to seeing `packageVersion` and forget about `package_version`.

hadley: Ok, I think I've fixed it but I need to figure out how to test it and that feels like a post-weekend problem 😄

hadley:
I reverted the changes to the `findPackageRepoAndSource()` to preserve the existing algorithm. I plan to merge this shortly and then work on some more refactoring before re-tackling the motivating problem.

PR #672, opened 2023-02-15
Refactor test helpers to use withr patterns

hadley:


tests/testthat/test-cert.R
```diff
@@ -1,149 +1,149 @@
...
+  addServer(
+    url = "https://localhost:4567/",
+    name = "cert_test_b",
+    quiet = TRUE
+  )
+
+  # add the certificate
+  addServerCertificate(
+    name = "cert_test_b",
+    certificate = "certs/sample.crt",
```
aronatkins, line 35: Another opportunity for `test_path`? (multiple examples)

aronatkins (approved):
There's a similar cleanup opportunity in `test-http.R`; the `datafile`, for example, is test-local (not shared state between the server and tests).

The `input` and `output` are different, unfortunately, and server-managed state. They can be cleaned up after each test, but also should be removed as the server shuts down.

hadley:
I have my eye on `test-http.R` too, but I haven't quite figured out the right approach yet.

tests/testthat/test-cert.R
hadley: Good point.

PR #673, opened 2023-02-16
Mild refactoring of app metadata

hadley:
This PR a new function `appMetadata()` that collects the results of `inferQuartoInfo()`, `inferAppMode()`, `inferAppPrimaryDoc()`, `appHasParameters()`, and `detectPythonInDocuments()` into a single object. These functions (and their helpers) now all live a new, `appMetadata.R`, and I've increased the test coverage. I've done some small refactorings along the way to make the code a little clearer, but I've tried to stick to low-risk changes that are unlikely to impact the operation of the code.

This PR is mostly in preparation for some bigger refactorings that I'll make in future PRs:

* Reorder the functions within `appMetadata()` to simplify the logic. I've left comments in the code to remind me what needs to happen.
* Also collect `python`, `forceGeneratePythonEnvironment`, and `condaMode` mode in the metadata object.
* Refactor `createAppManifest()` and friends to take a single metadata object instead of many small arguments. (i.e. an "Introduce Parameter Object" refactoring).


R/appMetadata.R
```diff
@@ -0,0 +1,378 @@
+appMetadata <- function(appDir,
+                        appFiles = NULL,
+                        appPrimaryDoc = NULL,
+                        quarto = NULL,
+                        contentCategory = NULL,
+                        isCloudServer = FALSE,
+                        metadata = list()) {
+
+  if (is.null(appFiles)) {
```
aronatkins, line 9: Discovery of `appFiles` via `bundleFiles` feels like an operation that should be done by the caller, especially as its result is discarded.

There are separate calls to `appFiles` both here and in `inferAppMode`.

R/appMetadata.R
```diff
@@ -0,0 +1,378 @@
...
+                         appFiles = NULL,
+                         quartoInfo = NULL,
+                         isCloudServer = FALSE) {
+
+  if (is.null(appFiles)) {
+    appFiles <- bundleFiles(appDir)
+  }
+
+  # TODO(HW): detection rules only apply to files in top-level directory
+  # Should make that explicit in the code
```
aronatkins, line 74: On the runtime side, we effectively use things like `shiny::runApp()` using the directory defined by the bundle as the working directory. The content can reference resources in sub-directories (as allowed by `shiny` or the in-use framework), but those packages all expect that the primary "thing" is at that top-level.

The one exception to this: If folks should happen to accidentally create wrapping directories when constructing the `bundle.tar.gz`, those get unwound and discarded (e.g. `./wolfgang/amadeus/mozart/app.R` in the archive becomes ./app.R` assuming no other entries at those intermediate levels).

aronatkins: Do you think that it would be helpful to document these content-directory expectations in terms of how the code effectively gets run by Connnect (e.g. the `runApp` calls)? I can help enumerate those things, if you'd like.

R/appMetadata.R
```diff
@@ -0,0 +1,378 @@
...
+      # rmarkdown static content
+      if (isCloudServer) {
+        return("rmd-shiny")
+      }
+      return("rmd-static")
+    }
+  }
+
+  # We don't have an RMarkdown, Shiny app, or Plumber API, but we have a saved model
+  if (length(grep("(saved_model.pb|saved_model.pbtxt)$", appFiles, ignore.case = TRUE, perl = TRUE)) > 0) {
```
aronatkins, line 158: FYI: I have outstanding work to remove support for the TF app-mode from `rsconnect`. We're letting the most recent Connect release (2023.01.1) age a little before making that change.

R/appMetadata.R
```diff
@@ -0,0 +1,378 @@
...
+  hasQuartoCompatibleFiles <- length(qmdFiles) > 0 || length(rmdFiles > 0)
+  requiresQuarto <- (hasQuartoCompatibleFiles && hasQuartoYaml) || length(qmdFiles) > 0
+
+  # We gate the deployment of content that appears to be Quarto behind the
+  # presence of Quarto metadata. Rmd files can still be deployed as Quarto
+  # content.
+  # TODO(HW): I think this logic should be moved into appMetdata() because this
+  # should also error if the user specified appMode = "quarto-shiny"
+  if (requiresQuarto && is.null(quartoInfo)) {
+    stop(paste(
```
aronatkins, line 117: How are you deciding when to `stop` and when to `cli::cli_abort`?

R/appMetadata.R
```diff
@@ -0,0 +1,378 @@
...
+  } else if (is.list(server) && identical(server[["type"]], "shiny")) {
+    TRUE
+  } else {
+    FALSE
+  }
+}
+
+# If deploying an R Markdown, Quarto, or static content, infer a primary
+# document if one is not already specified.
+  # Note: functionality in inferQuartoInfo() depends on primary doc inference
```
aronatkins, line 195: Inconsistent indentation.

R/appMetadata.R
```diff
@@ -0,0 +1,378 @@
...
+      cli::cli_abort(c(
+        "Failed to determine {.arg appPrimaryDoc}.",
+        x = "No files matching {.str {ext}}."
+      ))
+    }
+  }
+
+  if (sum(matching) > 1) {
+    # if we have multiple matches, pick the first
+    appFiles[matching][[1]]
```
aronatkins, line 226: This "alphabetically first" heuristic probably deserves to be explained, somewhere.

R/appMetadata.R
```diff
@@ -0,0 +1,378 @@
...
+    # If no appPrimaryDoc has been provided, we will use the file that will be
+    # chosen if this deployment ends up with an R Markdown or Quarto app mode.
+    # This works because:
+
+    # - App modes are only used to gate primary doc inference; the behavior does
+    #   not differ between app modes.
+    # - inferAppPrimaryDoc() returns appPrimaryDoc() if it is not null.
+
+    # TODO(HW): eliminate this code by running inferAppPrimaryDoc() before
+    # inferQuartoInfo()
```
aronatkins, line 312: @toph-allen - do you still have the context behind this two-pass workflow?

Callers like the `quarto` package and the IDE may provide Quarto metadata, but that's not guaranteed.

R/bundlePackage.R
```diff
@@ -157,3 +157,14 @@ validatePackageSource <- function(pkg) {
 
   return()
 }
+
+appUsesR <- function(quartoInfo) {
```
aronatkins, line 161: Should `appUsesR` and `appUsesPython` float together?

R/deployApp.R
```diff
@@ -520,37 +520,20 @@ bundleApp <- function(appName, appDir, appFiles, appPrimaryDoc, assetTypeName,
                       isCloudServer = FALSE, metadata = list(), image = NULL) {
   logger <- verboseLogger(verbose)
 
-  quartoInfo <- inferQuartoInfo(
+  logger("Inferring App mode and parameters")
+  metadata <- appMetadata(
```
aronatkins, line 524: `metadata` is a function argument; you're overwriting the incoming value.

tests/testthat/test-appMetadata.R
```diff
@@ -0,0 +1,291 @@
...
+  dir <- local_temp_app(list("foo.Rmd" = c("```{python}", "1+1", "````")))
+  expect_true(detectPythonInDocuments(dir))
+
+  dir <- local_temp_app(list("foo.qmd" = c("```{python}", "1+1", "````")))
+  expect_true(detectPythonInDocuments(dir))
+})
+
+# quarto ------------------------------------------------------------------
+
+fakeQuartoMetadata <- function(version, engines) {
```
aronatkins, line 178: This helper produces data for the existing `metadata` argument. I think the IDE can pass in other information through this optional argument. @toph-allen - do you have a reference to that?

aronatkins (changes requested):
It appears as if `metadata` is being overwritten.

R/appMetadata.R
hadley: I was meaning something a bit more narrowly scoped. Currently we have code like:

```R
rmdFiles <- grep("^[^/\\\\]+\\.rmd$", appFiles, ignore.case = TRUE, perl = TRUE, value = TRUE)
```

I was thinking we could make the regexps simpler and the logic clearer by doing:

```R
rootFiles <- appFiles[dirname(appFiles) = "."]
# ...
rmdFiles <- grep("\\.rmd$", rootFiles, ignore.case = TRUE, perl = TRUE, value = TRUE)
```

R/appMetadata.R
hadley: I left this as `stop()` because I'll rewrite it in the next round of changes.

R/bundlePackage.R
hadley: Maaaaaybe? But I think you could equally argue that they should live near where they are used. I suspect these will get refactored out at some point in the future (i.e. to become `usesR` and `usesPython` fields of `appMetadata`) so I'm not too worried.

R/deployApp.R
hadley: I'll call this `appMetadata` to be consistent with the function name.

R/appMetadata.R
hadley: I do it here because I think it's a reasonable default and it makes testing code much cleaner. I suspect this will end up looking cleaner as I continue to refactor.

hadley: Actually, one way to eliminate an argument would be to combine `appDir` with `appFiles`/`appPrimaryDoc` to create absolute paths in `appMetadata()`. That would eliminate one argument from each of `inferQuartoInfo()`, `inferAppMode()`, `inferAppPrimaryDoc()`, `appHasParameters()`, and `detectPythonInDocuments()`.

R/appMetadata.R
hadley: Good to know, because this regular expression is inconsistent with the others.

R/appMetadata.R
toph-allen: Yeah, the context is in [this PR](https://github.com/rstudio/rsconnect/pull/604).

If the user passes a Quarto path to the `quarto` argument, we want to deploy that app as Quarto if it's Quarto-compatible (even if it only has Rmds, not qmds.).

This leads to a silly circular dependency. This is a little hazy, because it was 8 months ago, but something like: 

- If the user hasn't provided a primary doc, we need to infer it, e.g. `deployApp(appDir = ".", quarto = "quarto")` where `"."` is not a Quarto project but a dir containing an Rmd.
- Primary doc inference depends on app mode.
- App mode inference depends on the result of `quarto inspect`.
- If a user doesn't explicitly pass a primary doc, we still want a primary doc to `quarto inspect`.

I can try to explain it in a different way.

I need to get offline now. Will review the full PR in the morning.

R/appMetadata.R
aronatkins: Ah. Gotcha. That's a very different TODO! :)

R/appMetadata.R
aronatkins: I'm OK keeping the `bundleFiles` call here, for now. Do we continue to need both calls?

If you make paths absolute, we'll need to make sure we retain the correct directory structure within the bundle.

R/writeManifest.R
```diff
@@ -56,56 +56,38 @@ writeManifest <- function(appDir = getwd(),
     appFiles <- explodeFiles(appDir, appFiles)
   }
 
-  quartoInfo <- inferQuartoInfo(
+  appMetadata <- appMetadata(
```
aronatkins, line 59: My preference would be to avoid variable shadowing like you've done here; is that common practice in other R code?

aronatkins (approved):
I'll let @toph-allen comment in the morning about the quarto-mode workflow, but my earlier concerns have been addressed.

tests/testthat/test-appMetadata.R
toph-allen: Here's where the Quarto `metadata` object comes from: https://github.com/quarto-dev/quarto-r/blob/08caf0f42504e74b7fc5382abed0820d5f415b3e/R/publish.R#L117-L121

R/writeManifest.R
hadley: You tell me: https://github.com/rstudio/rsconnect/blob/main/R/deployApp.R#L331-L333 😛 

But seriously, I'll keep thinking about this name as I continue to refactor this code.

R/appMetadata.R
hadley: I wonder if it shouldn't actually be an error

R/appMetadata.R
hadley: I think what's not obvious from that PR (because the functions were previously quite spread out), is that we could just call `inferAppPrimaryDoc()` first. I'm pretty sure this is a safe change, but we can discuss it more in the next PR.

R/appMetadata.R
hadley: I'll think about this more in a follow PR.

toph-allen (approved):
I like the direction of these changes. When adding incremental new features to a complex process like this, it feels like the best way to limit the complexity you're adding is to be strict about scoping things you add in self-contained functions, but it takes a more holistic refactor to actually *reduce* complexity.

R/appMetadata.R
aronatkins: This flow happens when trying to deploy an entire directory containing `a.Rmd` and `b.Rmd`. This choice is "reasonable", but it was not an explicit user choice.

This is effectively the behavior of:

```r
rmarkdown::run(file = NULL, dir = getwd())
```

That said, it would be better to encourage folks to make an explicit choice.

hadley: The other problem is I have no mental model of what "primary" doc actually means.

aronatkins: The simplest example: Assume you have a directory layout like:

```
project/
    primary.Rmd
    chunka.Rmd
    chunkb.Rmd
```

The `rmarkdown::render` needs to know its rendering target. It cannot be called against a directory. The primary document in this case is `primary.Rmd`.

Personally, I always give my projects meaningful names and then have `index.Rmd` within it, which makes the notion of "primary" pretty obvious. More complicated layouts are more common, however, and we often see folks with collections of Rmd files, and they all get deployed together, even though they often want only one document or a subset of the Rmds included.

Many other types of content do not require a primary document; this rendering case is the strongest example.

R/appMetadata.R
```diff
@@ -0,0 +1,378 @@
...
+    stop(paste(
+      "Attempting to deploy Quarto content without Quarto metadata.",
+      "Please provide the path to a quarto binary to the 'quarto' argument."
+    ))
+  }
+
+  # Shiny or Quarto documents with "server: shiny" in their YAML front matter
+  # are rmd-shiny or quarto-shiny.
+  if (any(shinyRmdFiles) || any(shinyQmdFiles)) {
+    if (!is.null(quartoInfo)) {
```
hadley, line 126: @toph-allen can you comment on why we use mode `quarto-shiny` when you have Rmd files and `quarto` set?

hadley: Don't worry, found #595

PR #675, opened 2023-02-16
Hide token/key from casual inspection

hadley:
This obviously isn't fool proof, but should be enough to avoid accidentally disclosing in a log file.


R/accounts.R
```diff
@@ -303,8 +325,13 @@ accountInfo <- function(name = NULL, server = NULL) {
   info <- as.list(accountDcf)
   # remove all whitespace from private key
   if (!is.null(info$private_key)) {
-    info$private_key <- gsub("[[:space:]]", "", info$private_key)
+    info$private_key <- secret(gsub("[[:space:]]", "", info$private_key))
```
aronatkins, line 328: Does `rsconnect::accounts` need the same adjustment?

This adjusts responses from the package API and should be included in NEWS.md.

aronatkins (commented):
NEWS and `accounts` adjustment.

R/accounts.R
hadley: `accounts()` just includes `name` and `server` — did you mean `servers()`?

aronatkins: Probably, yes!

hadley: I'm not sure that the `certificates` field in `servers()` is actually secret, but it is very long, so redacting it is a service to the user.

aronatkins (approved):

toph-allen (approved):

hadley:
@toph-allen `stringsAsFactors` strikes yet again.

PR #676, opened 2023-02-16
Use rlang helpers to do more argument checking

hadley:
* Used `check_string()` instead of `checkStringParam()`
* Introduced new `check_file()` and `check_directory()`
* Use `dirExists()` everywhere


R/deploySite.R
```diff
@@ -39,18 +39,14 @@ deploySite <- function(siteDir = getwd(),
     reason = "to deploy websites"
   )
 
+  # validate and normalize siteDir
+  check_directory(siteDir)
```
hadley, line 43: This needs to happen _before_ attempting to set the directory

R/auth.R
```diff
@@ -1,9 +1,7 @@
 cleanupPasswordFile <- function(appDir) {
 
-  # normalize appDir path and ensure it exists
-  appDir <- normalizePath(appDir, mustWork = FALSE)
-  if (!file.exists(appDir) || !file.info(appDir)$isdir)
-    stop(appDir, " is not a valid directory", call. = FALSE)
+  check_directory(appDir)
```
aronatkins, line 3: @kippandrew Should `rsconnect` still keep these references to `passwords.txt`?

R/deployApp.R
```diff
@@ -186,10 +185,8 @@ deployApp <- function(appDir = getwd(),
   }
 
   # normalize appDir path and ensure it exists
-  appDir <- normalizePath(appDir, mustWork = FALSE)
-  if (!file.exists(appDir)) {
-    stop(appDir, " does not exist")
-  }
+  check_directory(appDir)
```
aronatkins, line 188: Pre-existing nit, but this should be merged with the earlier `check_string(appDir)`. By this time, it has already been used in `runStartupScripts`.

aronatkins (approved):

PR #681, opened 2023-02-17
More `appMetadata()` refactoring

hadley:
I've made a bunch of small and steady changes to hopefully make `appMetadata()` easier to understand. Highlights:

* New `hasQuartoMetadata()` function makes allows us to figure out if quarto is needed earlier, allowing `inferQuartoInfo()` to be simpler and to come later in the process.
* I've pulled out the one special case of `inferAppMode()` that used the `appPrimaryDoc`, making it more clear that we can figure out the mode, and then figure out the primary doc.
* `inferAppMode()` now just gets a vector paths in the root directory of the app.

I'm reasonably confident that this hasn't change any behaviour and that the overall flow should now be easier to understand.

### DIFF
--- a/R/appMetadata.R
+++ b/R/appMetadata.R
@@ -12,13 +12,20 @@ appMetadata <- function(appDir,
   # https://github.com/quarto-dev/quarto-r/blob/08caf0f42504e7/R/publish.R#L117-L121
   hasQuarto <- !is.null(quarto) || hasQuartoMetadata(metadata)
 
-  appMode <- inferAppMode(
-    appDir = appDir,
-    appPrimaryDoc = appPrimaryDoc,
-    appFiles = appFiles,
-    hasQuarto = hasQuarto,
-    isCloudServer = isCloudServer
-  )
+  # Generally we want to infer appPrimaryDoc from appMode, but there's one
+  # special case
+  if (!is.null(appPrimaryDoc) &&
+      tolower(tools::file_ext(appPrimaryDoc)) == "r") {
+    appMode <- "shiny"
+  } else {
+    appMode <- inferAppMode(
+      appDir = appDir,
+      appFiles = appFiles,
+      hasQuarto = hasQuarto,
+      isCloudServer = isCloudServer
+    )
+  }
+
   appPrimaryDoc <- inferAppPrimaryDoc(
     appPrimaryDoc = appPrimaryDoc,
     appFiles = appFiles,
@@ -62,7 +69,6 @@ appMetadata <- function(appDir,
 # infer the mode of the application from its layout
 # unless we're an API, in which case, we're API mode.
 inferAppMode <- function(appDir,
-                         appPrimaryDoc = NULL,
                          appFiles = NULL,
                          hasQuarto = FALSE,
                          isCloudServer = FALSE) {
@@ -78,13 +84,6 @@ inferAppMode <- function(appDir,
   plumberFiles <- grep("^(plumber|entrypoint).r$", appFiles, ignore.case = TRUE, perl = TRUE)
   if (length(plumberFiles) > 0) {
     return("api")
-  }
-
-  # TODO(HW): I think this special case should live in appMetadata so that
-  # inferAppMode no longer needs the `appPrimaryDoc` argument
-  if (!is.null(appPrimaryDoc) &&
-      tolower(tools::file_ext(appPrimaryDoc)) == "r") {
-    return("shiny")
   }
 
   # Shiny application using single-file app.R style.
--- a/R/appMetadata.R
+++ b/R/appMetadata.R
@@ -18,9 +18,10 @@ appMetadata <- function(appDir,
       tolower(tools::file_ext(appPrimaryDoc)) == "r") {
     appMode <- "shiny"
   } else {
+    # Inference only uses top-level files
+    rootFiles <- appFiles[dirname(appFiles) == "."]
     appMode <- inferAppMode(
-      appDir = appDir,
-      appFiles = appFiles,
+      file.path(appDir, appFiles),
       hasQuarto = hasQuarto,
       isCloudServer = isCloudServer
     )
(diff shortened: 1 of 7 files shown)


R/appMetadata.R
```diff
@@ -6,32 +6,27 @@ appMetadata <- function(appDir,
...
+
+  # Generally we want to infer appPrimaryDoc from appMode, but there's one
+  # special case
+  if (!is.null(appPrimaryDoc) &&
+      tolower(tools::file_ext(appPrimaryDoc)) == "r") {
+    appMode <- "shiny"
   } else {
-    assetTypeName <- "application"
+    # Inference only uses top-level files
+    rootFiles <- appFiles[dirname(appFiles) == "."]
```
aronatkins, line 22: We should explain that `appFiles` should contain relative paths referencing files beneath `appDir`. Should `file.path(appDir, someAppFile)` not exist, it is silently dropped (see `rsconnect:::explodeFiles`).

I forgot this detail and needed to run a few `rsconnect:::standardizeAppFiles` examples to prove that it would only return relative paths.

R/appMetadata.R
```diff
@@ -6,32 +6,27 @@ appMetadata <- function(appDir,
...
+    appMode <- inferAppMode(
+      file.path(appDir, appFiles),
```
aronatkins, line 24: Does `inferAppMode` require absolute paths? Should we rename the argument to make it easier for the reader to remember that this instance has `absoluteAppFiles` rather than the relative paths present in `appFiles`?

R/appMetadata.R
```diff
@@ -58,72 +67,54 @@ appMetadata <- function(appDir,
...
+      "Can't deploy Quarto content when {.arg quarto} is {.code NULL}.",
+      i = "Please supply a path to a quarto binary in {.arg quarto}."
     ))
   }
 
-  # Shiny or Quarto documents with "server: shiny" in their YAML front matter
-  # are rmd-shiny or quarto-shiny.
+  # Documents with "server: shiny" in their YAML front matter need shiny too
+  shinyRmdFiles <- sapply(rmdFiles, isShinyRmd)
+  shinyQmdFiles <- sapply(qmdFiles, isShinyRmd)
```
aronatkins, line 115: Given the checks that happened earlier, we would have erred if `length(qmdFiles) > 0 && !hasQuarto`.

Taken by itself, though, this Quarto/Rmd shiny logic implies that `.qmd` files could result in a `rmd-shiny` mode.

Does this help explain the cases at all?

```r
if (any(shinyQmdFiles) { "quarto-shiny" }
else if (any(shinyRmdFiles)) { 
    if (hasQuarto) { "quarto-shiny" }
    else { "rmd-shiny" }
}
```

We have similar logic for the rendered (rmd-static/quarto-static) case. I'm not sure how to best model that to avoid confusing ourselves.

R/appMetadata.R
```diff
@@ -285,61 +270,40 @@ documentHasPythonChunk <- function(filename) {
...
 # Run "quarto inspect" on the target and returns its output as a parsed object.
-quartoInspect <- function(appDir = NULL, appPrimaryDoc = NULL, quarto = NULL) {
-  if (is.null(quarto)) {
-    return(NULL)
-  }
-
+quartoInspect <- function(quarto, appDir = NULL, appPrimaryDoc = NULL) {
   # If "quarto inspect appDir" fails, we will try "quarto inspect
   # appPrimaryDoc", so that we can support single files as well as projects.
   paths <- c(appDir, file.path(appDir, appPrimaryDoc))
```
aronatkins, line 309: nit: missed earlier - the `quarto_path` function is test-only and can get moved back into a test-only helper file.

R/bundle.R
```diff
@@ -850,6 +850,13 @@ standardizeAppFiles <- function(appDir, appFiles = NULL, appFileManifest = NULL)
     # file list provided directly
     appFiles <- explodeFiles(appDir, appFiles)
   }
+
+  if (length(appFiles) == 0) {
+    cli::cli_abort(
+      "No content to deploy; {.arg appFiles} is empty"
```
aronatkins, line 856: Could we adjust the message? This case could mean any of:

1. No files in the `appDir` directory to enumerate
2. No files in the manifest, or the files mentioned by the manifest do not exist
3. No files in `appFiles`, or the files included in `appFiles` do not exist.

R/deployApp.R
```diff
@@ -192,6 +192,9 @@ deployApp <- function(appDir = getwd(),
   appDir <- normalizePath(appDir)
 
   # create the full path that we'll deploy (append document if requested)
+  # TODO(HW): we use appPrimaryDoc here, but we have not inferred it yet
+  # so the appPath will different deneding on whether it's explicitly
+  # supplied or inferred from the files in the directory.
```
aronatkins, line 197: Related: `appPath` is also known as the variable `recordDir`, which is _NOT_ always a directory!

aronatkins (approved):
I like how these iterations are gradually uncovering some subtle assumptions, like expectations about relative and absolute paths at different layers.

R/deployApp.R
hadley: FWIW I really dislike the very nested directories that `recordDir()` uses and at some point I'll propose something more compact.

R/bundle.R
hadley: Makes sense. 

Since I'm in here, do you think I should make `appFiles` and `appFileManifest` mutually exclusive? It seems like supplying both is likely to be a mistake.

aronatkins: The implementation indicates that they are not both used at once, so yes. I expect that this will eventually need revisiting to consider all of the file manifest, an incoming manifest.json, the manually specified list of files, and their interactions with ignore files.

hadley: Not in this PR, but I think `standardizeAppFiles()` should also get the dependency discovery code that currently lives in `deployDoc()` (https://github.com/rstudio/rsconnect/blob/config-dir/R/deployDoc.R#L48-L54). That will eliminate the currently confusing `deployApp()` -> `deployDoc()` -> `deployApp()` sequence.

R/appMetadata.R
hadley: That variable was my attempt at conveying that impression 😄 but I like yours better.

R/appMetadata.R
hadley: Yeah, that's much more clear.

hadley: Hmmm, there's another logic bug — the presence of `quarto.yml` is used to determine whether or not quarto is required, but it doesn't actually influence the output type. I think that means the current checks for `hasQuarto` actually should be for `requiresQuarto`.

hadley: But `requiresQuarto` implies `hasQuarto` at this point, so maybe I've just confused myself.

PR #682, opened 2023-02-19
Config refactoring

hadley:
* Move all config path code into `config.R`
* Standardise function names and add `accountConfigFiles()` and `serverConfigFiles()` helpers
* Remove no-longer-needed handling of multiple servers in `accountConfigFile()`
* Reimplement `accounts()` to always return a data frame
* Move migration code into own file (to make main logic easier to see)


aronatkins (approved):
FYI: CI failures (lint error about trailing blank lines) is real.

The lintr configuration we currently have was the simplest "within reach" set of options; lots of the defaults are disabled. Feel free to suggest alterations.

hadley:
@aronatkins I don't have much experience with package linting, so I'm mostly paying attention to the experience.

PR #683, opened 2023-02-20
Capture python metadata into a single object

hadley:
This might be an overly clever approach, but I capture all of the information needed to infer the python environment into a closure. Then that can be passed around instead of separate `python`, `condaMode`, and `forceGenerate` parameters. 
I also added a new helper, `pythonPathOrSkip()` (mimicking `quartoPathOrSkip()`) that gets the path using `getPython()`.


R/bundlePython.R
```diff
@@ -0,0 +1,134 @@
+# Create anonymous function that we can later call to get all needed python
+# metdata for the manifest
+pythonConfigurator <- function(python,
+                               forceGenerate = FALSE,
+                               condaMode = FALSE) {
```
hadley, line 5: Given that `condaMode` is always `FALSE`, is it ok to remove?

R/bundlePython.R
```diff
@@ -0,0 +1,134 @@
...
+      version = environment$python,
+      package_manager = list(
+        name = environment$package_manager,
+        version = environment[[environment$package_manager]],
+        package_file = environment$filename,
+        contents = environment$contents
+      )
+    )
+  } else {
+    cli::cli_abort(environment$error)
```
hadley, line 111: Can you suggest a way to test this branch of code? i.e. is there a simple scenario that will cause `resources/environment.py` to fail?

R/bundlePython.R
```diff
@@ -0,0 +1,134 @@
...
+        cli::cli_abort(
+          "Failed to detect python environment",
+          parent = err
+        )
+      }
+    )
+  }
+}
+
+appUsesPython <- function(quartoInfo) {
```
hadley, line 32: Maybe this should still live in `bundle.R` since that's where it's used?

R/bundle.R
```diff
@@ -425,18 +358,15 @@ createAppManifest <- function(appDir, appMode, contentCategory, hasParameters,
...
-    pyInfo <- inferPythonEnv(appDir, python, condaMode, forceGenerate)
-    if (is.null(pyInfo$error)) {
-      # write the package list into requirements.txt/environment.yml file in the bundle dir
-      packageFile <- file.path(appDir, pyInfo$package_manager$package_file)
-      cat(pyInfo$package_manager$contents, file = packageFile, sep = "\n")
-      pyInfo$package_manager$contents <- NULL
-    } else {
-      stop(paste("Error detecting python environment:", pyInfo$error), call. = FALSE)
-    }
+  needsPython <- appUsesPython(quartoInfo) || "reticulate" %in% names(packages)
```
aronatkins, line 361: If this is the only use, maybe pass `packages` into `appUsesPython`?

R/bundlePython.R
aronatkins: Yes. 🔥

R/bundlePython.R
aronatkins: I'm OK with it here.

R/bundlePython.R
aronatkins: Unfortunately, no. @mmarchetti - could you help create a scenario to exercise the error conditions, here?

tests/testthat/helper-paths.R
```diff
@@ -0,0 +1,45 @@
+pythonPathOrSkip <- function() {
+  skip_if_not_installed("reticulate")
+
+  path <- getPython()
```
aronatkins, line 4: Not sure if it's meaningful, but the existing tests skipped when `Sys.which` indicated that there was no Python.. and then used that Python to check for `pip`. That Python was then used to set the environment variables used internally by `getPython`.

The `getPython` function relies on environment variables.

aronatkins (approved):
You may have adjusted how we determine when to skip Python tests; intentional?

R/bundlePython.R
mmarchetti: Possible causes:
* An error running `python -m pip --version`
* An error running `python -m pip list --format=freeze`
* `requirements.txt` exists but is not readable

Probably the last one would be the easiest condition to create.

hadley: @mmarchetti you mean like create it and then chmod 000 it? Or can I just create something that doesn't parse?

tests/testthat/helper-paths.R
hadley: Ok, I'll contemplate that.

R/bundlePython.R
```diff
@@ -0,0 +1,138 @@
...
+    shQuote(env_py),
+    if (condaMode) "-c",
+    if (forceGenerate) "-f",
+    shQuote(workdir)
+  )
+
+  # First check for reticulate. Then see if python is loaded in reticulate
+  # space, verify anaconda presence, and verify that the user hasn't specified
+  # that they don't want their conda environment captured.
+  hasConda <- is_installed("reticulate") &&
```
hadley, line 88: @aronatkins so `condaMode = TRUE` used to imply that you _didn't_ want to capture your conda environment?

R/bundle.R
hadley: Or maybe inline `appUsesPython` here?

R/deployApp.R
```diff
@@ -287,15 +285,25 @@ deployApp <- function(appDir = getwd(),
...
     logger("Bundle upload started")
     withStatus(paste0("Uploading bundle (", application$id, ")"), {
-      # python is enabled on Connect but not on Shinyapps
-      python <- getPythonForTarget(python, accountDetails)
-      bundlePath <- bundleApp(target$appName, appDir, appFiles,
-                              appPrimaryDoc, contentCategory, verbose, python,
-                              condaMode, forceGeneratePythonEnvironment, quarto,
-                              isCloudServer(accountDetails$server), metadata, image)
+      bundlePath <- bundleApp(
+        target$appName,
```
hadley, line 295: Should supply argument names here

R/bundlePython.R
aronatkins: I'm not really sure what interaction this was trying to capture. We always call `resources/environment.py`, just with different arguments and a different interpreter.

```bash
# hasConda=TRUE, condaMode=FALSE
conda run -p <PYTHON> resources/environment.py 
# hasConda=TRUE, condaMode=TRUE
# IMPOSSIBLE
# hasConda=FALSE, condaMode=FALSE
<PYTHON> resources/environment.py
# hasConda=FALSE, condaMode=TRUE
<PYTHON> resources/environment.py -c
```

Since `condaMode` is never enabled, it looks like we can strip all support (including the `-c` argument to `resources/environment.py`). The `hasConda` logic feels like it still needs to remain.

R/bundle.R
aronatkins: Either option feels fine, and not a permanent decision should we want to further refactor.

R/bundlePython.R
hadley: Chmod 000 worked for me.

tests/testthat/helper-paths.R
hadley: In hindsight, this is an easy fix — just always rely on reticulate to find python. I'll double check that it works on CI, but I think this is the right approach for tests.

PR #684, opened 2023-02-20
Refactor file bundling functions

### LINKED ISSUE #568: .rcsignore file not ignoring files
I am trying to use the .rcsignore file to ignore a couple large data files in the app that do not need to be deployed to shinyapps.io. Each time the file is still uploaded to shinyapps.io. I am able to have these files ignored when I used the `appFiles` argument in `deployApp()`. 

I also noticed this file does not appear in the Files pane of RStudio unless I go into the More settings and select 'Show Hidden Files'.

hadley:
* Move all in `bundleFiles.R`
* Refactor `listBundleFiles()` and `explodeFiles()` to perform `enforceBundleLimits()` at end. IMO this makes the recursion used in these functions much easier to understand.
* `listBundleFiles()` no errors if too big/too many files, rather than silently truncating, thus matching the behavior of `standardizeAppFiles()`.
* I also tested more of the file removal behaviour in `listBundleFiles()`, revealing a bug in the support for `.rscignore`.  Fixes #568.
* Updated errors and warnings


R/appMetadata.R
```diff
@@ -166,6 +166,28 @@ isShinyRmd <- function(filename) {
   }
   is_shiny_prerendered(yaml$runtime, yaml$server)
 }
+
+yamlFromRmd <- function(filename) {
```
hadley, line 170: Discovered this was in the wrong place while extracting the bundling functions.

R/bundleFiles.R
```diff
@@ -0,0 +1,246 @@
...
+  contents <- setdiff(contents, c("renv", "renv.lock"))
+  contents <- contents[!isKnitrCacheDir(contents)]
+  contents <- setdiff(
+    contents,
+    c(".DS_Store", ".git", ".gitignore", ".quarto", ".Rhistory", ".svn")
+  )
+
+  # remove any files lines listed .rscignore
+  if (".rscignore" %in% contents) {
+    ignoreContents <- readLines(file.path(dir, ".rscignore"))
```
hadley, line 160: Two changes here:

* `readLines(file.path(dir, ".rscignore"))` instead of `readLines(".rscignore")`; if that ever worked  before it was by coincidence
* Also remove `.rscignore` itself from the bundle

hadley:
To do:

* [ ] Update `listBundleFile()` docs
* [ ] Update docs for `.rscignore` in `deployApp()`

R/bundleFiles.R
```diff
@@ -239,9 +117,75 @@ isKnitrCacheDir <- function(subdir, contents) {
...
+  }
+
+  # Info for each file lets us know to recurse (directories) or aggregate (files).
+  is_dir <- dirExists(file.path(dir, contents))
+  names(is_dir) <- contents
+
+  children <- character()
+  for (name in contents) {
+    if (isTRUE(is_dir[[name]])) {
+      dirList <- recursiveBundleFiles(file.path(dir, name), depth + 1)
```
aronatkins, line 164: This change removes of `getOption("rsconnect.max.bundle.size")` and `getOption("rsconnect.max.bundle.files")` as the walk progresses. This is an important check because folks occasionally use an `appDir` which is massive (a consequence of deploying `$HOME/document.Rmd`, for example).

Some historical issues with this code:
* https://github.com/rstudio/rsconnect/pull/463 - incorrect size calculations
* https://github.com/rstudio/rsconnect/issues/514 - permissions problems when enumerating
* https://github.com/rstudio/rsconnect/pull/546 - enforcement on `appFiles`

I'm not finding the original issue, but folks observed "hangs" (very long runtime) if there's no feedback until after the entire directory is processed.

aronatkins (changes requested):
Enforcement during the walk is removed.

R/bundleFiles.R
hadley: Oh bummer 😬 I'll think about how to restore this.

R/bundleFiles.R
```diff
@@ -0,0 +1,239 @@
+# build the list of files to deploy -- implicitly (directory contents),
+# explicitly via list, or explicitly via manifest. Always returns paths
+# related to `appDir`
+standardizeAppFiles <- function(appDir,
+                                appFiles = NULL,
+                                appFileManifest = NULL,
+                                error_call = caller_env()) {
+
+  check_exclusive(appFiles, appFileManifest, .require = FALSE, .call = error_call)
```
hadley, line 9: Should test for null, not missing

hadley:
@aronatkins restored eager evaluation and added a test.

R/bundleFiles.R
```diff
@@ -0,0 +1,261 @@
...
+  totalSize <- 0
+
+  for (f in files) {
+    target <- file.path(dir, f)
+    if (!file.exists(target)) {
+      # doesn't exist
+      next
+    } else if (dirExists(target)) {
+      # a directory; explode it
+      contentPaths <- file.path(f, list.files(
```
aronatkins, line 70: nit: This expansion does not terminate the walk when encountering large directories. This existed in the previous implementation, as well, and is only present when someone manually provides a large directory in `appFiles` or `appFileManifest`.

Maybe mark this limitation with a comment and leave for the next iteration?

R/bundleFiles.R
```diff
@@ -0,0 +1,261 @@
...
+  lines <- readLines(appFileManifest, warn = FALSE)
+
+  # remove empty/comment lines
+  lines <- lines[nzchar(lines)]
+  lines <- lines[!grepl("^#", lines)]
+  lines
+}
+
+# given a list of mixed files and directories, explodes the directories
+# recursively into their constituent files, and returns just a list of files
```
aronatkins, line 56: nit: reiterate that the returned set of files are paths relative to the root.

R/bundleFiles.R
```diff
@@ -0,0 +1,261 @@
...
+explodeFiles <- function(dir, files) {
+  exploded <- character()
+
+  totalFiles <- 0
+  totalSize <- 0
+
+  for (f in files) {
+    target <- file.path(dir, f)
+    if (!file.exists(target)) {
+      # doesn't exist
```
aronatkins, line 66: nit: This may be warning-worthy, since we are exploding a set of files provided to us. Unchanged from the previous implementation. Fine to leave as-is and we can revisit.

R/bundleFiles.R
```diff
@@ -0,0 +1,261 @@
...
+  # Info for each file lets us know to recurse (directories) or aggregate (files).
+  is_dir <- dir.exists(file.path(dir, contents))
+  names(is_dir) <- contents
+
+  children <- character()
+  for (name in contents) {
+    if (isTRUE(is_dir[[name]])) {
+      out <- recursiveBundleFiles(
+        dir = file.path(dir, name),
+        root_dir = root_dir,
```
aronatkins, line 145: nit: mixing argument name styles. camels hate snakes. i'm completely aware that i am actively throwing stones and simultaneously living in a glass house. 😜

R/bundleFiles.R
```diff
@@ -0,0 +1,261 @@
...
+      )
+
+      children <- append(children, file.path(name, out$contents))
+      totalFiles <- out$totalFiles
+      totalSize <- out$totalSize
+    } else {
+      children <- append(children, name)
+      totalFiles <- totalFiles + 1
+      totalSize <- totalSize + file_size(file.path(dir, name))
+    }
```
aronatkins, line 158: Enforcement checks should happen after each `name` is processed rather than after the fill set of `list.files`. Consider a directory with 20k files, a directory containing 10 2GB files, or a deep, broad directory trees.

You do not need checks both after each file and prior to exit; checking in the loop body is sufficient.

R/bundleFiles.R
aronatkins: Likely worked because:
* ignore files usually at top level
* working directory almost always the content directory

aronatkins (approved):
The bundle-size enforcement could move into the directory-contents loop rather than outside, which will cause us to exit earlier for additional types of directory structures.

R/bundleFiles.R
hadley: I have some ideas for unifying `exploreFiles()` and `bundleFiles()` that I can try out in a future PR.

R/bundleFiles.R
hadley: EmBrAcE ThE ChAoS

R/bundleFiles.R
hadley: Yeah, I misunderstand the recursion here.

PR #685, opened 2023-02-20
Remove unneeded check

### LINKED ISSUE #619: We should remove the API check for plumber APIs when deploying to cloud or shinyapps.io.
There's validation on the content type when the bundle is uploaded now - so we can remove this check,

https://github.com/rstudio/rsconnect/blob/main/R/deployApp.R#L339-L346

hadley:
As this is now done server side. Fixes #619.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -264,20 +264,10 @@ deployApp <- function(appDir = getwd(),
   target <- deploymentTarget(appPath, appName, appTitle, appId, account, server)
 
   # test for compatibility between account type and publish intent
-  if (isCloudServer(target$server)) {
-    # Publishing an API to shinyapps.io will not currently end well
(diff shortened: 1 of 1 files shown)


aronatkins (approved):
I'm OK with this change, since it means that the target server can evolve without requiring changes to `rsconnect`. 

@ssinnott - will shinyapps.io return a reasonable error when presented with an unsupported app-mode?

@hadley - last time we discussed with the hosted team, there were still no plans to support plumber in shinyapps.io, which is why we left in the client-side check.

ssinnott:
> @ssinnott - will shinyapps.io return a reasonable error when presented with an unsupported app-mode?

We also do a server side check now that errors on building the bundle if the app mode isn't one that we support. Pretty happy to have this removed I'm pretty suspicious this check actually protects shinyapps.io against much.

PR #686, opened 2023-02-20
Test R profile tweaking

### LINKED ISSUE #520: Do not ignore the `renv` folder by default
Note: I'm not sure if it's a `{rsconnect}` or `{renv}` issue, @kevinushey feel free to correct me if you think this feature should be in `{renv}` instead.

If I create a new `{shiny}` app, use `renv::init()`, then do `rsconnect::deployApp()`, it will fail. 

The reason behind this is that the .Rprofile contains a call to `source("renv/activate.R")`, and by default, clicking on the blue button in RStudio sends all the files except `renv/` & `renv.lock` (so does `rsconnect::deployApp()`. 

Also, trying to include the `renv/` folder by hand will fail: 

```r
> rsconnect::deployApp(appFiles = c("app.R", "renv/", "renv.lock", ".Rprofile"))
The following potential problems were identified in the project files:

The following lines contain calls to the browseURL function:
72: | `file_show("path")`                         | `browseURL("path")`               | `open path`                       |

The following lines contain calls to the browseURL function:
58: browseURL("http://localhost:9454/")

Remove browseURL calls; browseURL does not work in deployed applications.
Do you want to proceed with deployment? [Y/n]: Y
Preparing to deploy application...
Update application currently deployed at
https://colinfay.shinyapps.io/renvtest/? [Y/n] Y
DONE
Uploading bundle for application: 4258497...Error: Unable to retrieve package records for the following packages:
- 'foo', 'testRcppClass', 'testRcppInterfaceExporter', 'testRcppInterfaceUser', 'testRcppModule'
In addition: There were 50 or more warnings (use warnings() to see the first 50)
```

Possible solution : 

- From the user perspective : 
    - don't send the .RProfile, but that file is useful for other purposes, and  for example it's suggested to use this file here https://docs.rstudio.com/connect/1.5.4/admin/package-management.html#private-repositories
    - Comment in the .RProfile, but that means that you need to do this manually, and prevents from using an automatic deployment from git
    -  wrap the `source("renv/activate.R")` inside an if statement detecting if we are on Connect / shinyapps.io

- in `{rsconnect}` 
    - Deploy the `renv/` folder alongside the app 
    - Comment the .RProfile if it detects a `source("renv/activate.R")`

- in `{renv}`
	- always wrap the `source("renv/activate.R")` inside an if statement detecting if we are on Connect / shinyapps.io

That being said, maybe you have a better workflow here. 

I'm deploying apps on Connect from Git, that we locally build using `{renv}`, and that would be nice to have them both work flawlessly without tweaking the `.Rprofile` :) 

Cheers, 
Colin

hadley:
No significant changes to behaviour, but it's now tested, verifying that it actually works. 

Closes #520.

### DIFF
--- a/tests/testthat/test-bundle.R
+++ b/tests/testthat/test-bundle.R
@@ -646,6 +646,18 @@ test_that("tarImplementation: checks environment variable and option before usin
 
 # tweakRProfile -----------------------------------------------------------
 
+test_that(".Rprofile tweaked automatically", {
+  dir <- withr::local_tempdir()
+  writeLines('source("renv/activate.R")', file.path(dir, ".Rprofile"))
+
+  bundled <- bundleAppDir(dir, list.files(dir, all.files = TRUE))
+  expect_match(
+    readLines(file.path(bundled, ".Rprofile")),
+    "Modified by rsconnect",
+    all = FALSE
+  )
+})
+
 test_that("removes renv/packrat activation", {
   path <- withr::local_tempfile(lines = c(
     "# Line 1",
--- a/tests/testthat/test-bundle.R
+++ b/tests/testthat/test-bundle.R
@@ -660,7 +660,7 @@ test_that("removes renv/packrat activation", {
       tweakRProfile(path)
       writeLines(readLines(path))
     },
-    transform = function(x) gsub("on .+:", "on <NOW>:", x)
+    transform = function(x) gsub("on \\d{4}.+", "on <NOW>", x)
   )
 })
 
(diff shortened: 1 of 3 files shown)


tests/testthat/test-bundle.R
```diff
@@ -644,6 +644,38 @@ test_that("tarImplementation: checks environment variable and option before usin
 })
 
 
+# tweakRProfile -----------------------------------------------------------
+
+test_that(".Rprofile tweaked automatically", {
```
aronatkins, line 649: Could you confirm that `.Rprofile` files without the patterns are left untouched?

aronatkins (approved):

Issue #669, opened 2023-02-14
More information from `deployApp()`


hadley:
* If `appFiles` not specified, output which files are uploaded
* In `deploymentTarget()` output if using existing deployment (with `appId`) or if it's new. Might be worth outputting fully reified `account` + `server`. And need `appTitle` too. Maybe `createDeploymentTarget()` should report everything if `verbose`?

For QA, need to use that information to verify a few publishing workflows to make sure that we don't clash with how the IDE sets values. Need to check both new, republish, and "other destination" workflows.

PR #688, opened 2023-02-21
Remove assetTypeName

hadley:
Proof of concept for what removing `assetTypeName` might look like. I don't have particularly strong feelings about this PR, but it's one less argument to pass around, and I don't think it has much impact on the quality of user messaging.


aronatkins (approved):
The type name was added in https://github.com/rstudio/rsconnect/pull/63, which shifted away from "application", but may have been before we started using "content" as a generic name.

Connect identifies and communicates the target content type; the `assetTypeName` was an attempt to produce a "humanized" description on the client side. It's an incomplete mapping, though.

I'm OK removing this field.

If it feels worthwhile, we could add a one-time "print content type for humans" rather than passing it around everywhere.

The current output:

```
Preparing to deploy application...DONE
Uploading bundle for application: 9470...DONE
Deploying bundle: 22068 for application: 9470 ...
[Connect] Building Shiny application...
...
```

might become:

```
Preparing to deploy...DONE
Bundling Shiny application...
Uploading bundle: 9470...DONE
Deploying bundle: 22068 (9470) ...
[Connect] Building Shiny application...
...
```

The "Bundling Shiny application..." text could happen immediately after the call to `appMetadata`. I'll share the Connect mapping we use separately.

hadley:
I'll generally review the messaging from `deployApp()` as part of #669.

PR #689, opened 2023-02-21
Improve data frame conversion in showMetrics()

### LINKED ISSUE #528: showMetrics returned malformed data.frame
I used to use this code to save my shinyapps.io app metrics data periodically because only 4 months data are kept.

```
df <- rsconnect::showMetrics("container_status",
                             c("connect_count", "connect_procs"), interval = "60s",
                             # c("connect_count"), interval = "60s",
                             from = as.integer(start_time), appName = "ctmmweb",
                             server="shinyapps.io", account = "ctmm")
```

It used to return a data.frame like this

```
    connect_count connect_procs  timestamp
1               0             0 1564952340
2               0             0 1564952400
3               0             0 1564952460
```

However starting from Aug, 2020, the result became

```
    connect_count    connect_procs timestamp         NA
1               0 container_status         0 1608583560
2               0 container_status         0 1608583620
3               0 container_status         0 1608583680
4               0 container_status         0 1608583740

```

Looks like there are some column misaligned here.

hadley:
Fixes #528

No idea how to unit test this, but I verified that it failed before this fix and worked correctly afterwards by getting the metrics for one of my shiny apps:

```R
showMetrics(
  "container_status",
  c("connect_count", "connect_procs"),
  appName = "ms-upload-validate",
)
```


R/usage.R
```diff
@@ -98,17 +103,10 @@ showMetrics <- function(metricSeries, metricNames, appDir = getwd(), appName = N
...
-    X
-  })
-
-  # convert to data frame
-  df <- data.frame(matrix(unlist(points), nrow = length(points), byrow = TRUE), stringsAsFactors = FALSE)
-  colnames(df) <- c(metricNames, "timestamp")
-  return(df)
+  points <- lapply(data$points, as.data.frame, stringsAsFactors = FALSE)
+  points <- do.call(rbind, points)
+  points$time <- .POSIXct(points$time / 1000)
```
hadley, line 108: This was previously called `timestamp`, but that code has been failing for some time, and I think it's better to keep the column names as close as possible to what shinyapps.io returns?

aronatkins (approved):
Adding @kippandrew as reviewer, as he wrote the initial implementation.

PR #692, opened 2023-02-22
More `deployApp()` refactoring

hadley:
Nothing major here, just pulling out a few helper functions to make the flow of `deployApp()` a bit clearer.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -302,28 +302,7 @@ deployApp <- function(appDir = getwd(),
                               isCloudServer(accountDetails$server), metadata, image)
 
       if (isCloudServer(accountDetails$server)) {
-
-        # Step 1. Create presigned URL and register pending bundle.
-        bundleSize <- file.info(bundlePath)$size
-
-        # Generate a hex-encoded md5 hash.
-        checksum <- fileMD5.as.string(bundlePath)
-        bundle <- client$createBundle(application$id, "application/x-tar", bundleSize, checksum)
-
-        logger("Starting upload now")
-        # Step 2. Upload Bundle to presigned URL
-        if (!uploadBundle(bundle, bundleSize, bundlePath)) {
-          stop("Could not upload file.")
-        }
-        logger("Upload complete")
-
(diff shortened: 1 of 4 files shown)


R/deployApp.R
```diff
@@ -272,21 +266,15 @@ deployApp <- function(appDir = getwd(),
...
-    }
-
-    if (!identical(appVisibility, currentVisibility)) {
-      withStatus(paste0("Setting visibility to ", appVisibility), {
-        client$setApplicationProperty(application$id,
-                                   "application.visibility",
-                                   appVisibility)
-      })
-    }
+  # Change _visibility_ before uploading data
```
aronatkins, line 269: Alternatively, hide the control logic, too:

```r
conditionallyChangeVisibility(accountDetails, application, appVisibility, quiet = quiet)
```

R/deployApp.R
```diff
@@ -385,20 +352,46 @@ deployApp <- function(appDir = getwd(),
...
+  hook <- getOption(option)
+  if (!is.function(hook)) {
+    return()
+  }
+
+  name <- gsub("rsconnect.", "", option, fixed = TRUE)
+  name <- gsub(".", "-", name, fixed = TRUE)
+
+  if (verbose) {
+    cat("Invoking ", name, " hook ", option, "\n", sep = "")
```
aronatkins, line 390: I would be fine simply displaying `Invoking hook: rsconnect.post.deploy` and avoiding the pretty `name`; the hook option is descriptive.

aronatkins (approved):

R/deployApp.R
hadley: Yeah, for sure, that's a bigger win.

hadley: Hmmm, I tried it and I didn't like — you also need to pass `client` (so that's a whole grab bag of arguments), and I kind of like keeping all the big picture status updates in the main function.

aronatkins: Good point; I missed `client` in that main function. Pity that the server / account don't hang off the client object...

PR #695, opened 2023-02-22
Lint refactoring 2

hadley:
* Make lint printing use S3 methods and `cat()` rather than `message()`
* Move `lint()` and friends to `lint.R`
* Reorganise tests a little
* Remove custom relative path code in favour of trusting `standardizeAppFiles()`


hadley:
The tests work for me locally, so I wonder why absolute path detection is failing on CI 🤔

aronatkins (approved):
Approved, contingent on CI passing. The lint failures are real (trailing whitespace). I can't see the reason for CI failures, either.

hadley:
Ok, because that linter requires the directory to actually exist.

PR #696, opened 2023-02-22
Reduce the usage of appPath

hadley:
* I think the hooks should always get a directory, not sometimes a path.

* It's clear that `deploymentTarget` should get the path where the deployments are recorded

@aronatkins this is part of a very gradual unpicking of the complicated `appPath`/`recordDir` logic in `deployApp()`.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -237,7 +237,7 @@ deployApp <- function(appDir = getwd(),
   }
 
   # invoke pre-deploy hook if we have one
-  runDeploymentHook(appPath, "rsconnect.pre.deploy", verbose = verbose)
+  runDeploymentHook(appDir, "rsconnect.pre.deploy", verbose = verbose)
 
   appFiles <- standardizeAppFiles(appDir, appFiles, appFileManifest)
 
@@ -247,7 +247,7 @@ deployApp <- function(appDir = getwd(),
   }
 
   # determine the deployment target and target account info
-  target <- deploymentTarget(appPath, appName, appTitle, appId, account, server)
+  target <- deploymentTarget(recordDir, appName, appTitle, appId, account, server)
 
   # test for compatibility between account type and publish intent
   if (!isCloudServer(target$server) && identical(upload, FALSE)) {
@@ -353,7 +353,7 @@ deployApp <- function(appDir = getwd(),
 
   # invoke post-deploy hook if we have one
   if (deploymentSucceeded) {
-    runDeploymentHook(appPath, "rsconnect.post.deploy", verbose = verbose)
+    runDeploymentHook(appDir, "rsconnect.post.deploy", verbose = verbose)
   }
 
   logger("Deployment log finished")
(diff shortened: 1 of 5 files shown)


hadley:
I think this suggests that `appSourceDoc` doesn't do anything useful and could be deprecated:

* `appSourceDoc` is only used in one place to set `appPath`.
* After this PR, `appPath` is only to set `recordDir`

So `appSourceDoc` is basically just an alias for `recordDir`

I reviewed the code from v0.8.29 (https://github.com/rstudio/rsconnect/blob/v0.8.29/R/deployApp.R) to make sure I hadn't accidentally refactored out some other use case, but it doesn't look like it.  If I deprecate it, we need to adjust `deploySite()` to use `recordDir` instead.

aronatkins (approved):
Unsure if we can safely shift the hooks from using the path (DIR/document.Rmd) to always using the project directory... there are very few public references to be found.

R/deployApp.R
```diff
@@ -237,7 +237,7 @@ deployApp <- function(appDir = getwd(),
   }
 
   # invoke pre-deploy hook if we have one
-  runDeploymentHook(appPath, "rsconnect.pre.deploy", verbose = verbose)
+  runDeploymentHook(appDir, "rsconnect.pre.deploy", verbose = verbose)
```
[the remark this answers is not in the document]
hadley: Yeah, this is a change, and the documentation for the hooks is murky: 

>   \item{\code{rsconnect.pre.deploy}}{A function to run prior to deploying content; it receives as an argument the path to the content that's about to be deployed.}

What does "content about to be deployed" mean? (especially if you've used `appFiles` or friends). 

My guess is that this won't affect any code in the wild, but I'll add a news bullet.

FWIW I think the thematic usage would break when `appPath` is not a directory.

aronatkins:
Agreed that `appSourceDoc` feels redundant and can be replaced by `deploySite` computing an appropriate `recordDir`.

R/deployApp.R
```diff
@@ -247,7 +247,7 @@ deployApp <- function(appDir = getwd(),
   }
 
   # determine the deployment target and target account info
-  target <- deploymentTarget(appPath, appName, appTitle, appId, account, server)
+  target <- deploymentTarget(recordDir, appName, appTitle, appId, account, server)
```
aronatkins, line 250: Internally, `deploymentTarget` calls this argument `appPath`; should it be renamed?

R/deploymentTarget.R
```diff
@@ -18,11 +18,10 @@ deploymentTarget <- function(appPath = ".",
   if (nrow(appDeployments) == 0) {
     fullAccount <- findAccount(account, server)
     if (is.null(appName)) {
-      appName <- generateAppName(appTitle, appPath, account, unique = FALSE)
+      appName <- generateAppName(appTitle, recordPath, account, unique = FALSE)
```
hadley, line 21: This reinforces that the `recordPath` really should be original source of the content.

R/deployApp.R
hadley: I've renamed it to `recordPath`, in anticipation of renaming `recordDir` to `recordPath` in a future PR.

PR #697, opened 2023-02-23
Eliminate use of appSourceDoc in deploySite()

hadley:
The use of `setwd()` really makes these functions hard to reason about 😬. But I think this is correct


R/deploySite.R
```diff
@@ -77,17 +80,17 @@ deploySite <- function(siteDir = getwd(),
     appFiles <- appFiles[!grepl("^.*\\.([Rr]|[Rr]md|md)$", appFiles)]
   }
 
-  # determine appSourceDoc
-  if (file.exists("index.Rmd"))
-    appSourceDoc <- "index.Rmd"
-  else
-    appSourceDoc <- "index.md"
+  if (is.null(recordDir)) {
+    # recordDir should usually be an absolute path, but above we set the
```
aronatkins, line 84: Perhaps `deployApp` should `normalizePath(recordDir)`, as it does for `appDir`.

aronatkins (approved):

R/deploySite.R
hadley: Oh I can just use `file.path()` to make it explicit. This doesn't imply a change of behaviour because we were previously passing a path relative to the working directory, which is set to `siteDir`.

PR #698, opened 2023-02-23
Standardize single doc deployment

hadley:
Extract out `standardizeSingleDocDeployment()` and use it in `deployDoc()` and `deployApp()`. 

I've also documented this feature of `deployApp()`, which lead to some other minor doc changes since other function inherited from `deployApp()`. This does raise a question of whether these functions should also support an `appDir` that's actually a path to a document.


R/deployApp.R
```diff
@@ -154,7 +155,23 @@ deployApp <- function(appDir = getwd(),
 
   condaMode <- FALSE
 
-  check_directory(appDir)
+  check_string(appDir)
+  if (dirExists(appDir)) {
+    # ok
+  } else if (file.exists(appDir) && isStaticFile(appDir)) {
+    doc <- standardizeSingleDocDeployment(appDir)
```
hadley, line 162: One thing that's slightly unappealing about this approach is that we use a different approach to determining `appFiles` depending on whether you do `deployApp("foo.Rmd")` or `deployApp(appPrimaryDoc = "foo.Rmd")`. An alternative approach would be to add an `appPrimaryDoc` argument to `appStandardizeFiles()` and call `rmarkdown::find_external_dependencies()` there when `appPrimaryDoc` is supplied.

hadley: Another approach, since this never appears to have been documented, would be to deprecate this use `appDir` and tell folks to call `deployDoc()` instead.

aronatkins: Interesting. I forgot about this difference.

```r
# Uses the default appDir=getwd() and enumerates all files in the directory for bundling
deployApp(appPrimaryDoc = "foo.Rmd")
# Uses the foo.Rmd to identify the dependencies of the document; only those are bundled
deployDoc("foo.Rmd")
```

The existing `deployApp("foo.Rmd")` feels like a convenience, but we should nudge folks towards using `deployDoc` should they provide a file.

Added in: https://github.com/rstudio/rsconnect/pull/16, and before the notion of a "primary doc".

I'm uneasy about the automatic use of `rmarkdown::find_external_dependencies`. On the one hand, it does feel like it will produce the set of files that most folks want to include -- directory explosion is often too greedy. On the other hand, its use would add to the "magic" that happens in the "lower-level" `deployApp`. Its inclusion could suddenly force more folks to compute their own `appFiles`.

aronatkins: Deprecating the `deployApp -> deployDoc -> deployApp` workflow feels like the right path.

aronatkins (approved):

R/deployApp.R
hadley: Yeah, I think the fundamentally different mechanism of file discovery suggests that this is a `deployDoc()` feature, and we should deprecate this old path.

PR #699, opened 2023-02-23
Deprecate appPrimaryDoc(); implement findRecordPath()

hadley:


R/deployApp.R
```diff
@@ -157,6 +155,25 @@ deployApp <- function(appDir = getwd(),
   check_directory(appDir)
   check_string(appName, allow_null = TRUE)
 
+  if (!is.null(appPrimaryDoc)) {
```
hadley, line 158: Previously `appPrimaryDoc` was only checked in passing when determining `recordDir`.

R/deployApp.R
```diff
@@ -193,20 +210,6 @@ deployApp <- function(appDir = getwd(),
   # normalize appDir path
   appDir <- normalizePath(appDir)
 
-  # create the full path that we'll deploy (append document if requested)
-  # TODO(HW): we use appPrimaryDoc here, but we have not inferred it yet
```
hadley, line 197: I now understand this — there's a subtle difference between deploying a document and a directory containing a single document.

DESCRIPTION
```diff
@@ -18,6 +18,7 @@ Imports:
     curl,
     digest,
     jsonlite,
+    lifecycle,
```
aronatkins, line 21: aside: `lifecycle` also means we depend upon `glue`.

aronatkins (approved):

R/deployApp.R
```diff
@@ -157,6 +155,25 @@ deployApp <- function(appDir = getwd(),
...
+  if (!is.null(appPrimaryDoc)) {
+    check_string(appPrimaryDoc)
+    if (!file.exists(file.path(appDir, appPrimaryDoc))) {
+      cli::cli_abort("`appPrimaryDoc` not found inside `appDir`")
+    }
+  }
+
+  if (!is.null(appSourceDoc)) {
+    lifecycle::deprecate_warn(
+      when = "0.9.0",
```
hadley, line 167: Note that I'm assuming the next release will be 0.9.0; I think this is a reasonable assumption as it's likely to have some bigger new features.

PR #700, opened 2023-02-23
Ignore more files

### LINKED ISSUE #111: Exclude known temporary files
For example, an open Excel file `"foo.xslx"` that has unsaved changes will create a sibling file `"~$foo.xslx"`.
Emacs will create a backup file `"foo.csv~"`

It would nice to exclude these files, at least by default.

### LINKED ISSUE #632: ignore project-local python virtual environments by default
An internal user encountered the following while attempting to deploy a reticulated project:

```
Error in enforceBundleLimits(appDir, files$totalSize, length(files$contents)) : 
  The directory /usr/home/jeremy.allen/projects/reticulate-with-venv cannot be deployed because it contains too many files (the maximum number of files is 10000). Remove some files or adjust the rsconnect.max.bundle.files option.`
```

you can work around this by adjusting the max bundle files or by specifying `appFiles` but ignoring the files to begin with might be a better default.

hadley:
* Ignore backup files. Fixes #111.
* Ignore python venvs. Fixes #632.


R/bundleFiles.R
```diff
@@ -174,19 +174,28 @@ ignoreBundleFiles <- function(dir, contents, depth = 0) {
...
+    contents <- setdiff(contents, c("manifest.json", "app_cache", ".Rproj.user"))
   }
 
-  # exclude renv files, knitr cache dirs, and another well-known files
-  contents <- setdiff(contents, c("renv", "renv.lock"))
-  contents <- contents[!isKnitrCacheDir(contents)]
-  contents <- setdiff(
-    contents,
-    c(".DS_Store", ".git", ".gitignore", ".quarto", ".Rhistory", ".svn")
+  ignore <- c(
```
aronatkins, line 180: Is it intentional that a file like `.Rproj.user` is ignored only at the root while `.Rhistory` is ignored throughout the hierarchy? I don't recall why we differentiate. I'd like to understand the reasoning.

aronatkins (approved):

R/bundleFiles.R
hadley: I don't know either; I'd be in favour of just ignoring everything at every level.

aronatkins: The initial implementation only ignored at the root: https://github.com/rstudio/rsconnect/pull/112
renv files were ignored at all levels: https://github.com/rstudio/rsconnect/pull/367

I think I'm also inclined to ignore throughout the hierarchy, since we cannot assume the appDir doesn't have complicated directory structures beneath. NEWS the change?

aronatkins (approved):

PR #702, opened 2023-02-23
Save deployment refactoring

hadley:
This is a refactoring that shouldn't change behaviour compared to released rsconnect.

I verified expected behaviour using this script:

1. Create a new project called `testdeploy{random number}`
2. Create empty `index.Rmd`
3. Call `rsconnect::deployApp()`. 
4. Deployment should succeed & app opens in browser
5. Call `rsconnect::deployApp()` again.
6. Deployment should succeed & app opens in browser
7. Run `rsconnect::forgetDeployment()` and press enter.
8. Should be prompted to confirm that existing app is the same one.
9. Call `rsconnect::deployApp()` again.
10. Deployment should succeed & app opens in browser


hadley:
Ooops, I should have tested that before implementing — connect requires unique names.

hadley:
Re-tested and back to the original behaviour, but with better user feedback.

R/deployApp.R
```diff
@@ -512,34 +499,46 @@ getAppById <- function(id, account = NULL, server = NULL, hostUrl = NULL) {
...
-  if (is.null(target$appId)) {
-    # list the existing applications for this account and see if we
-    # need to create a new application
-    app <- getAppByName(client, accountInfo, target$appName)
-  } else {
-    # we already know the app's id, so just retrieve the rest of the metadata
-    app <- client$getApplication(target$appId)
+  # Use appId from previous deployment
+  if (!is.null(target$appId)) {
+    return(client$getApplication(target$appId))
```
aronatkins, line 504: What should happen if the target app ID has been deleted? The old flow would leave `app = NULL` and try some other option.

aronatkins (commented):
What is the workflow when rsconnect discovers a default target that has been deleted on the server? The `target` produced by `deploymentTarget` has only considered on-disk state, which may be stale?

R/deployApp.R
hadley: I just tried it out, and I think if the application id no longer exists, then `getApplication()` will error with a 404. That's probably worth considering in more detail (and I'll add a fall back in case `getApplication()` does return `NULL` in other cases), but I think it's out of scope for this PR.

PR #706, opened 2023-02-24
Unify explodeFiles() and bundleFiles()

hadley:


R/bundleFiles.R
```diff
@@ -127,14 +87,25 @@ bundleFiles <- function(appDir) {
   listBundleFiles(appDir)$contents
 }
 
+explodeFiles <- function(dir, files) {
+  files <- files[file.exists(file.path(dir, files))]
```
hadley, line 91: This makes it very clear that `explodeFiles()` is doing something weird. Should I make this a warning or error?

hadley: A warning is likely to be better, since this was previously silent.

R/bundleFiles.R
```diff
@@ -127,14 +87,25 @@ bundleFiles <- function(appDir) {
...
                                  rootDir = dir,
                                  depth = 0,
                                  totalFiles = 0,
-                                 totalSize = 0) {
+                                 totalSize = 0,
+                                 ignoreFiles = TRUE) {
   # generate a list of files at this level
-  contents <- list.files(dir, all.files = TRUE, no.. = TRUE, include.dirs = TRUE)
-  contents <- ignoreBundleFiles(dir, contents, depth = depth)
+  if (is.null(contents)) {
```
aronatkins, line 103: Question: Does this recursion get confused if some of the incoming `contents` are complex paths rather than paths at the root level? The depth calculation could be off with something like: `contents = c("index.Rmd", "images/logos/logo.png")`

aronatkins (approved):
Could you test more complex `appFiles` inputs to `explodeFiles`?

R/bundleFiles.R
hadley: `depth` is only needed for ignoring files and will go away in #700, so we can dodge that question 😄

PR #709, opened 2023-02-24
Save deployments more aggressively

### LINKED ISSUE #270: Simplify download-edit-redeploy workflow
Right now, it's possible to download a bundle, edit it, and redeploy it to the same endpoint using the following steps:

1. Download the bundle using the Source Versions dialog and untar locally.
1. Edit and redeploy using:

```r
rsconnect::deployApp("content/path", server = "host", account = "username", appId = 1234)
```

Where `1234` is the ID of the downloaded app.

Things that make this less than awesome:

- Finding the app id is confusing. A downloaded bundle reads `app-<appId>-<bundleId>.tar.gz`. There is no cue distinguishing app ID from bundle ID other than position. 
    - The app ID is not included in the manifest; arguably it should not be, because the manifest identifies the content of the bundle and not the location where it is deployed.
- This workflow goes to a core collaboration use case that's likely to be repeated many times by many people with varying levels of familiarity with R, RStudio, Shiny/RMarkdown, and RStudio Connect. 
    - There might have to be IDE changes to get this to work as pleasantly as we'd like.

### LINKED ISSUE #677: `saveDeployment()` before upload?
It's possible for bundling or uploading to fail, which means that you lose useful metadata (e.g. `server` and `account`)
Would it make sense to `saveDeployment()` prior to uploading? e.g.

```R
saveDeployment(
  recordDir,
  target$appName,
  target$appTitle,
  target$username,
  target$account,
  accountDetails$server,
  serverInfo(target$server)$hostUrl,
  application$id,
  NULL, # previously bundle$id
  application$url,
  metadata
)
```

Would need to verify that you can still resume even when `bundle$id` is not set. (It doesn't look like `deploymentTarget()` ever uses `bundle$id` so maybe this is trivially true?)

hadley:
Fixes #677. Fixes #270.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # rsconnect 0.8.30 (development version)
 
+* `deployApp()` is more aggressive about saving deployment data, which should
+  make it less likely that you need to repeat yourself after a failed 
+  deployment. In particular, it now saves both before and after uploading the
+  contents (#677) and it saves when you're updating content originally created
+  by someone else (#270).
+  
 * `deployApp("foo.Rmd")` has been deprecated. It was never documented, and
   it does the same job as `deployDoc()` (#698).
 
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -267,6 +267,12 @@ deployApp <- function(appDir = getwd(),
   # get the application to deploy (creates a new app on demand)
   withStatus("Preparing to deploy", {
     application <- applicationForTarget(client, accountDetails, target, forceUpdate)
+    saveDeployment(
+      recordPath,
+      target = target,
+      application = application,
+      metadata = metadata
+    )
   })
 
   # Change _visibility_ before uploading data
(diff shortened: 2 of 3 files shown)


R/deployApp.R
```diff
@@ -298,27 +303,17 @@ deployApp <- function(appDir = getwd(),
         bundle <- client$uploadApplication(application$id, bundlePath)
       }
     })
-  } else {
-    # redeploy current bundle
-    bundle <- application$deployment$bundle
-  }
 
-  # write a deployment record only if this is the account that owns the content
-  if (is.null(application$owner_username) ||
```
hadley, line 307: When you're deploying to content that you don't own (but still have permission to update), you still want to update the `lastSyncTime`, bundle id` etc.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # rsconnect 0.8.30 (development version)
 
+* `deployApp()` is more aggressive about saving deployment data, which should
```
aronatkins, line 3: aside: I totally missed that we have `forgetDeployment` to help clean-up those pesky repeats I'm always creating when testing!

aronatkins (approved):

PR #711, opened 2023-02-24
Improve deployDoc() documentation

### LINKED ISSUE #319: Consider adding a `Rmd` example to the `rsconnect::deployApp()` documentation
Rmarkdown is listed in the description, but I think it would be instructive to indicate in the `appPrimaryDoc` parameter definition and/or the examples that you can specify a `Rmd` file for things like flexdashboard. The documentation is geared more toward proper shiny apps, I think. I hesitate to create a pull request because I'm not sure about the best way to indicate this to users, so I'm just filing as a suggestion based on my experience today.

hadley:
Fixes #319


R/deployDoc.R
```diff
@@ -1,27 +1,27 @@
...
+#' Deploy a single document
 #'
-#' Deploys an application consisting of a single R Markdown document or other
-#' single file (such as an HTML or PDF document).
+#' @description
+#' Deploys a single R Markdown, Quarto document, or other file (e.g. `.html` or
+#' `.pdf`).
+#'
+#' When deploying an `.Rmd`, `.Qmd`, or `.html`, `deployDoc()` will attempt to
+#' automaticaly discover depedencies using [rmarkdown::find_external_resources()].
```
aronatkins, line 8: spelling: dependencies. automatically.

Should we suggest `deploySite` or `bookdown::publish_book`, as well?

aronatkins (approved):

R/deployDoc.R
hadley: We already have cross-references to `deploySite()` in see also, and I think we're gradually moving away from bookdown in favour of quarto so I _think_ we're ok, as is.

PR #713, opened 2023-02-25
Change strategy for generating appName

### LINKED ISSUE #538: Add an option to require unique app names
Our app name generator allows an operation mode wherein unique app names are required:

https://github.com/rstudio/rsconnect/blob/main/R/title.R#L32

However, every call site for this function in `deployApp` passes "FALSE" to that parameter.

We should add an option to permit changing this behavior such that non-unique names produce deployment errors.

hadley:
* No longer use `appTitle`
* Make the reasoning behind use of `recordPath` more clear
* Only munge for shinyApps
* Deprecates `generateAppName()`

Fixes #538

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,14 @@
 # rsconnect 0.8.30 (development version)
 
+* `deployApp()` now derives `appName` from `appDir` and `appPrimaryDoc`, 
+  never using the title. It also tells you what name has used (#538).
+  It now only simplifies the path if you are publishing to shinyapps.io,
+  since its restrictions on application names are much tighter than those
+  of Posit Connect.
+  
+* `generateAppName()` has been deprecated as using the `appTitle` to generate
+  the `appName` leads to confusion if you later try to change the title.
+
 * `deployApp()` is more aggressive about saving deployment data, which should
   make it less likely that you need to repeat yourself after a failed 
   deployment. In particular, it now saves both before and after uploading the
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -33,10 +33,13 @@
 #'   being deployed.
 #' @param appSourceDoc `r lifecycle::badge("deprecated")` Please use
 #'   `recordDir` instead.
-#' @param appName Name of application (names must be unique within an account).
-#'   If not supplied on the first deploy, it will be generated from the base
-#'   name of `appDir` and `appTitle`, if supplied. On subsequent deploys,
-#'   it will use the previously stored value.
+#' @param appName Application name, a string consisting of letters, numbers,
+#'   `_` and `-`. The application name is used to identify applications on a
+#'   server, so much be unique.
+#'
+#'   If not specified, the first deployment will be automatically it from the
+#'   `appDir` for directory and website, and from the `appPrimaryDoc` for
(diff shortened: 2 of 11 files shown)


R/deployApp.R
```diff
@@ -269,7 +280,7 @@ deployApp <- function(appDir = getwd(),
       recordPath,
       target = target,
       application = application,
-      metadata = metadata
+      metadata = metadata,
```
aronatkins, line 283: Intentional trailing comma?

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # rsconnect 0.8.30 (development version)
 
+* `deployApp()` now derives `appName` from `appDir` and `appPrimaryDoc`, 
+  never using the title. It also tells you what name has used (#538).
+  It now only simplifies the path if you are publishing to shinyapps.io,
+  since its restrictions on application names are much tighter than those
+  of Posit Connect.
+  
+* `generateAppName()` has been deprecated as using the `appTitle` to generate
```
aronatkins, line 9: How does this change alter the workflow that folks use to repeatedly deploy the same content?

Today, I can use the publishing drop-down "Other Destination..." to repeatedly deploy the same Shiny application to a single server. They happen to have the same title (because I didn't override the IDE-suggested default) and come from the same directory (`shiny`, in this case).

The content I just created were given names "shiny3" and "shiny4".

hadley: Hmmm, good question. Looks like the IDE is using `generateAppName()` for that, https://github.com/rstudio/rstudio/blob/3fba3707760ee74e14f1d682ec49e726084ac3d3/src/cpp/session/modules/SessionRSConnect.R#L592, so I'll undeprecate.

hadley: I also realised that we should be using `generateAppName()` in https://github.com/rstudio/rsconnect/blob/defaultAppName/R/deployApp.R#L546-L549 to be consistent with the IDE, but that will take some refactoring in order to get the updated `appName` in the right place.

I've undeprecated `generateAppName()`

R/title.R
```diff
@@ -22,6 +23,7 @@
 #' (note that it is not guaranteed to be unique on the server). This behavior
 #' can be disabled by setting `unique = FALSE`.
 #'
+#' @keywords internal
```
hadley, line 26: I've kept this internal because I don't think it's really a user facing function

hadley:
@aronatkins I've minimised all other changes to focus on the core of this PR: changing the default strategy for generating an `appName`.

R/deploymentTarget.R
```diff
@@ -68,3 +70,29 @@ createDeploymentTarget <- function(appName,
     server = server
   )
 }
+
+defaultAppName <- function(recordPath, server = NULL) {
+  if (isDocumentPath(recordPath)) {
+    name <- file_path_sans_ext(basename(recordPath))
+    if (name == "index") {
+      # deploying a website; use name of parent directory
```
aronatkins, line 78: Why does `index` indicate a directory? My convention is to _ALWAYS_ use `index.Rmd`, with the directory as the meaningful name - even for single-document deploys.

The code feels right, but the comment might be misleading.

aronatkins (approved):

Issue #701, opened 2023-02-23
`deploymentTarget()` doesn't use `appId`


hadley:
Shouldn't it?

hadley:
Think about this: If the parameters supplied don't match any deployments, but other deployments exist, should we warn the user?

hadley:
@aronatkins do you have any sense of how `appId` is supposed to be used in `deployApp()`? There are many scenarios I can imagine, but none of them seem particularly compelling compared to supplying `appName`.

aronatkins:
The `rsconnect-python` examples show how to use the CLI (a Python application) to target content created by someone else. This example probably should be in terms of the content GUID rather than the numeric identifier, but does show the workflow.

https://docs.posit.co/connect/user/publishing-cli/#updating-a-different-deployment

I couldn't find an example using this R package, but the same idea applies.

The `appId` is not directly used by `deploymentTarget` other than recording the value on the returned object. `applicationForTarget` then uses the `appId` to fetch that specific content.

The `appName` is insufficient when collaborating because _your_ content named "shiny" is not _my_ content named "shiny". The name-based search will find _owned_ items with that name; it restricts its search to content owned by the requesting user (see the `account_id` filter in `listApplicaitons`).

hadley:
What if we just error if `appId` is supplied and we find an existing deployment? Then we can reserve specifying `appId` only for the case where you're attempting to update an add deployed by someone else, where the original deployment record has been lost.

aronatkins:
The `appId` feels like it should take higher priority than the name/title.

1. I deploy something. I lose the `rsconnect` directory but want to continue to deploy to that location. I use `appId`. Empty `rsconnect` directory.
2. I deploy something. I give you permission to deploy to that location. You use `appId`, which is not related to the contents of your `rsconnect` directory.
3. You and I can both deploy to some location; we use a shared script, containing `appId` to run that deploy. The contents of the `rsconnect` directory may or may not overlap with `appId`.

Specifying `appName` feels like "search for this thing", while specifying `appId` feels like "use this exact thing".

hadley:
Hmmm, interesting. I was thinking that you'd supply `appId` once, in order to set up the deployment record, then never again, relying on the information stored on disk.

In your scenario, what happens if you supply `appName` and `appId` and they don't match?

aronatkins:
* Multiple accounts: err, as the target location is ambiguous.
[...]

PR #714, opened 2023-02-26
Refactor `deployments()`

hadley:
This reimplements `deployment()` to read and combine all deployment dcf files, and then filter them. This shouldn't change any behaviour (except that the timestamp columns are now made POSIXct), but I think code is quite a bit easier to understand. (So will make implementing #701 easier)


R/config.R
```diff
@@ -110,21 +110,26 @@ accountConfigFiles <- function(server = NULL) {
...
-deploymentConfigFile <- function(appPath, name, account, server) {
-  accountDir <- file.path(deploymentConfigDir(appPath), server, account)
+deploymentConfigFile <- function(recordPath, name, account, server) {
+  accountDir <- file.path(deploymentConfigDir(recordPath), server, account)
   if (!file.exists(accountDir))
     dir.create(accountDir, recursive = TRUE)
   file.path(accountDir, paste0(name, ".dcf"))
 }
 
+deployConfigFiles <- function(recordPath) {
```
aronatkins, line 128: Should this be `deploymentConfigFiles` to mirror the naming of `deploymentConfigFile` and `deploymentConfigDir`?

tests/testthat/test-deployments.R
```diff
@@ -0,0 +1,96 @@
...
+  expect_equal(nrow(out), 2)
+
+  out <- deployments(dir, nameFilter = "my-app")
+  expect_equal(nrow(out), 1)
+
+  out <- deployments(dir, serverFilter = "foo")
+  expect_equal(nrow(out), 1)
+
+  out <- deployments(dir, accountFilter = "foo")
+  expect_equal(nrow(out), 1)
```
aronatkins, line 49: These tests cover the number of records, but not their shape.

These tests don't seem to cover the widening of metadata, proving that all columns are always returned, even when the original dcf contains a subset of fields.

aronatkins (approved):
Thinking in vectors. So concise.

tests/testthat/test-deployments.R
hadley: That's a bit hard to test at the moment because of the way these functions are structured; let me do a bit more refactoring and I'll include in the next PR

PR #715, opened 2023-02-27
Get locale name from registry

### LINKED ISSUE #233: Error detecting locale - incomplete final line
Hi, 

When deploying to a RStudio Connect server with `rsconnect`, I have this warning. 

```
Warning message:
Error detecting locale: Error in read.table(file = file, header = header, sep = sep, quote = quote, : incomplete final line found by readTableHeader on 'raw'
 (Using default: en_US) 
```

I believe there is something wrong in the process of detecting local on Windows. 

I found that `rsconnect:::systemInfo` is used for detecting local on windows. The command `systeminfo /FO csv` is called with `system`. I get back a result in a csv format. Currently, the function `systemInfo` use `read.csv` that cause the error. If I use `readr::read_csv` no more error. 

``` r

raw <- system("systeminfo /FO csv", intern = TRUE, wait = TRUE)
# I get a warning
info.csv <- read.csv(textConnection(raw))
#> Error in read.table(file = file, header = header, sep = sep, quote = quote, : incomplete final line found by readTableHeader on 'raw'
# I get no warning
info_csv <- readr::read_csv(raw)
#  and locale can be detected
locale <- as.character(info_csv[[20]])
locale
#> [1] "fr;Fran<U+0087>ais (France)"
# but I get an encoding issue that cause error after
strsplit(unlist(strsplit(locale, ";", fixed=TRUE)), "-", fixed=TRUE)
#> Warning in strsplit(locale, ";", fixed = TRUE): la chaîne de caractères
#> entrée 1 est incorrecte comme encodage UTF-8
#> [[1]]
#> [1] NA

# If I try to provide raw as text input, no warning
info.csv.txt <- read.csv(text = raw)
# I get a "strange" character
locale <- as.character(info.csv.txt[[20]])
locale
#> [1] "fr;Fran‡ais (France)"
# but it works
strsplit(unlist(strsplit(locale, ";", fixed=TRUE)), "-", fixed=TRUE)
#> [[1]]
#> [1] "fr"
#> 
#> [[2]]
#> [1] "Fran‡ais (France)"
```

Not sure what is the encoding of the string return by `systeminfo /FO csv`.
So I tested with what I know

``` r
readr::guess_encoding(raw)
#> # A tibble: 2 x 2
#>       encoding confidence
#>          <chr>      <dbl>
#> 1 windows-1252       0.38
#> 2 windows-1250       0.22
stringi::stri_enc_detect(raw)
#> [[1]]
#> [[1]]$Encoding
#> [1] "windows-1252" "windows-1250" "windows-1254" "UTF-16BE"    
#> [5] "UTF-16LE"     "GB18030"      "EUC-JP"       "EUC-KR"      
#> [9] "Big5"        
#> 
#> [[1]]$Language
#> [1] "fr" "ro" "tr" ""   ""   "zh" "ja" "ko" "zh"
#> 
#> [[1]]$Confidence
#> [1] 0.72 0.37 0.14 0.10 0.10 0.10 0.10 0.10 0.10
#> 
#> 
#> [[2]]
#> [[2]]$Encoding
#> [1] "windows-1252" "windows-1250" "UTF-16BE"     "UTF-16LE"    
#> [5] "EUC-JP"       "EUC-KR"       "windows-1254"
#> 
#> [[2]]$Language
#> [1] "fr" "ro" ""   ""   "ja" "ko" "tr"
#> 
#> [[2]]$Confidence
#> [1] 0.22 0.15 0.10 0.10 0.10 0.10 0.06
stringi::stri_enc_detect2(raw, locale = "fr")
#> [[1]]
#> [[1]]$Encoding
#>  [1] "macintosh"          "x-mac-turkish"      "ISO-8859-15"       
#>  [4] "windows-1258"       "ibm-1258_P100-1997" "ibm-1129_P100-1997"
#>  [7] "windows-1252"       "windows-1254"       "ibm-1252_P100-2000"
#> [10] "ibm-1254_P100-1995"
#> 
#> [[1]]$Language
#>  [1] NA NA NA NA NA NA NA NA NA NA
#> 
#> [[1]]$Confidence
#>  [1] 0.7500000 0.7500000 0.5833333 0.5833333 0.5833333 0.5833333 0.4027778
#>  [8] 0.4027778 0.4027778 0.4027778
#> 
#> 
#> [[2]]
#> [[2]]$Encoding
#>  [1] "ISO-8859-15"        "windows-1252"       "windows-1254"      
#>  [4] "windows-1258"       "ibm-1252_P100-2000" "ibm-1254_P100-1995"
#>  [7] "ibm-1258_P100-1997" "ibm-1129_P100-1997" "macintosh"         
#> [10] "x-mac-turkish"     
#> 
#> [[2]]$Language
#>  [1] NA NA NA NA NA NA NA NA NA NA
#> 
#> [[2]]$Confidence
#>  [1] 0.9655172 0.9655172 0.9655172 0.9655172 0.9655172 0.9655172 0.9655172
#>  [8] 0.9655172 0.7413793 0.7413793
```

If we try to provide encoding in some way

``` r
# If I specify the encoding in textConnection, no more warning
info.csv <- read.csv(textConnection(raw, encoding = "UTF-8"))
# I get also another "strange" character
locale <- as.character(info.csv[[20]])
locale
#> [1] "fr;Franâ€¡ais (France)"
# but it works
strsplit(unlist(strsplit(locale, ";", fixed=TRUE)), "-", fixed=TRUE)
#> [[1]]
#> [1] "fr"
#> 
#> [[2]]
#> [1] "Franâ€¡ais (France)"
```

Do you have any idea on this ? What fix could be done ?

Created on 2018-01-10 by the [reprex package](http://reprex.tidyverse.org) (v0.1.1.9000).

hadley:
Fixes #233

### DIFF
--- a/R/locale.R
+++ b/R/locale.R
@@ -1,84 +1,26 @@
-overrideWindowsLocale <- function(locale) {
-  map <- list()
-  map[["el_EL"]] <- "el_GR"
-  if (locale %in% names(map)) {
-    locale <- map[[locale]]
-  }
-  return(locale)
-}
-
 detectLocale <- function() {
-  sysName <- Sys.info()[["sysname"]]
-  if (identical(sysName, "Windows")) {
-    locale <- detectLocale.Windows()
+  if (!isWindows()) {
+    locales <- strsplit(Sys.getlocale("LC_CTYPE"), ".", fixed = TRUE)[[1]]
+    locales[[1]]
   } else {
-    locale <- detectLocale.Unix()
+    locale <- HCU_registry_key("Control Panel\\International\\User Profile")$Languages[[1]]
+    if (is.null(locale)) {
+      # Try approach that works on Windows 7
+      locales <- HCU_registry_key("Control Panel\\International")$LocaleName
+      if (is.null(locale)) {
+        # Otherwise fall back US English
+        locale <- "en-US"
+      }
+    }
+    gsub("-", "_", locale)
   }
-  return(locale)
 }
 
-detectLocale.Unix <- function() {
-  unlist(strsplit(Sys.getlocale("LC_CTYPE"), ".", fixed = TRUE))[1]
-}
-
-detectLocale.Windows <- function(useCache =
-                                  getOption("rsconnect.locale.cache", TRUE)) {
-
-  # default locale
-  locale <- "en_US"
-
-  cacheFile <- localeCacheFile()
-  if (file.exists(cacheFile) && useCache) {
-
-    # get chached
-    cache <- as.list(readDcf(cacheFile, all = TRUE))
-
-    locale <- unlist(cache$locale)
-
-  } else {
-
-    tryCatch({
-
-      # get system locale
-      locale <- systemLocale()
-
-      # write the user info
-      write.dcf(list(locale = locale),
-                cacheFile,
-                width = 100)
-
-    }, error = function(e) {
-      warning(paste0("Error detecting locale: ", e,
-                     " (Using default: ", locale, ")"), call. = FALSE)
-    })
-  }
-  return(overrideWindowsLocale(locale))
-}
-
-localeCacheFile <- function() {
-  normalizePath(file.path(rsconnectConfigDir(), "locale.dcf"), mustWork = FALSE)
-}
-
-systemLocale <- function() {
-  message("Detecting system locale ... ", appendLF = FALSE)
-
-  # get system locale
-  info <- systemInfo()
-  raw <- as.character(info[[20]])
-  parts <- strsplit(unlist(strsplit(raw, ";",  fixed = TRUE)), "-", fixed = TRUE)
-
-  if (length(parts[[1]]) >= 2) {
-    # normalize locale to something like en_US
-    locale <- paste(tolower(parts[[1]][1]), toupper(parts[[1]][2]), sep = "_")
-  } else {
-    locale <- paste(tolower(parts[[1]][1]), toupper(parts[[1]][1]), sep = "_")
-  }
-  message(locale)
-  return(locale)
-}
-
-systemInfo <- function() {
-  raw <- system("systeminfo /FO csv", intern = TRUE, wait = TRUE)
-  info <- read.csv(textConnection(raw))
-  return(info)
+HCU_registry_key <- function(key, default = NULL) {
+  tryCatch(
+    utils::readRegistry(key, hive = "HCU"),
+    error = function(err) {
+      default
+    }
+  )
 }
--- a/R/locale.R
+++ b/R/locale.R
@@ -3,22 +3,25 @@ detectLocale <- function() {
     locales <- strsplit(Sys.getlocale("LC_CTYPE"), ".", fixed = TRUE)[[1]]
     locales[[1]]
   } else {
-    # Possible values listed at https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/70feba9f-294e-491e-b6eb-56532684c37f
-    locales <- HCU_registry_key("Control Panel\\International")$LocaleName
-
-    if (is.null(locale)) {
-      # Otherwise fall back US English
-      locale <- "en-US"
-    }
-    gsub("-", "_", locale)
+    tryCatch(windowsLocale(), error = function(err) "en_US")
   }
 }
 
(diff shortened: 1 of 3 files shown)


R/locale.R
```diff
@@ -1,84 +1,26 @@
-overrideWindowsLocale <- function(locale) {
```
hadley, line 1: Might need to restore, but this seems rather ad hoc, and I can't see the motivating issue at https://github.com/rstudio/shinyapps/commit/b5f9f7cd5a5d9a5cafabd2ef7e234cfa643a663e

aronatkins: Here's a stackoverflow mentioning the same:
https://stackoverflow.com/questions/30163941/r-fail-to-deploy-on-shinyapps-io-unsupported-locale-greek-el-el

aronatkins (approved):
+1 from me, but I'm not familiar enough with the side-effects of locale problems to see potential fallout.

CC @kippandrew

R/locale.R
hadley: Original issue: https://groups.google.com/g/shinyapps-users/c/nUgepQwRcO4/m/na3hkd6Va5UJ?utm_medium=email&utm_source=footer&pli=1

hadley: Looks like that's the same person; so it's not two independent instances.

kippandrew: I think this approach is a big improvement, but I'm not sure how to confirm that Windows actually returns the correct language code with this method. The Greek case could have been a bug, but is it fixed?

hadley: Given that the previous fix was applied for a single person who had a problem in 2018, I'd prefer to err on the side of keeping the code simple. However, if you feel strongly about it, I'll add it back in.

R/locale.R
```diff
@@ -1,84 +1,26 @@
...
-
-  if (length(parts[[1]]) >= 2) {
-    # normalize locale to something like en_US
-    locale <- paste(tolower(parts[[1]][1]), toupper(parts[[1]][2]), sep = "_")
+  if (!isWindows()) {
+    locales <- strsplit(Sys.getlocale("LC_CTYPE"), ".", fixed = TRUE)[[1]]
+    locales[[1]]
   } else {
-    locale <- paste(tolower(parts[[1]][1]), toupper(parts[[1]][1]), sep = "_")
+    locale <- HCU_registry_key("Control Panel\\International\\User Profile")$Languages[[1]]
```
gaborcsardi, line 6: Actually, it is better to only use the other registry key, because this one can return the language code only, without the region. E.g. for Hungarian I get
```r
> HCU_registry_key("Control Panel\\International\\User Profile")$Languages
[1] "hu"    "en-US" "de-DE" "es-ES"
> HCU_registry_key("Control Panel\\International")$LocaleName
[1] "hu-HU"
```

OTOH, the second one is probably just the duplication of the language code. Which ends up being a correct locale for Hungary: `hu-HU`, but the same thing for Greek and `el` does not work. 

The PDF here has the list of the valid locale codes: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/70feba9f-294e-491e-b6eb-56532684c37f

R/locale.R
kippandrew: That’s fine with me.

R/locale.R
gaborcsardi: Another good resource, where they suggest the `$LocaleName` registry entry: https://ss64.com/locale.html

hadley: Does that linked doc imply that sometimes the `LocaleName` might include a script identify that we should strip?

gaborcsardi: Good point, the table at https://ss64.com/locale.html is auto-generated, and has locales with script ids, even without a region sometimes, e.g.:
```
az-Cyrl
az-Cyrl-AZ
```

So I _think_ we need to split by `-`, keep the first part, and remove the part that is not uppercase (if any). If we always need a region code, then add one if needed. 

There is potentially also a sort order at the end, separated by an underscore, e.g. `hu-HU_technl` but I am not sure if `LocaleName` would ever include that.

R/locale.R
gaborcsardi: I installed Greek on Windows 10, the version without a region, and the registry has `el-GR`, so this PR is probably OK.

gaborcsardi (approved):

PR #716, opened 2023-02-27
Fix bug in bundlePackages()

hadley:
Introduced in #670.


hadley:
@aronatkins FYI — I missed this because I haven't been testing the full deployment pathway as often as I should 😞.

PR #718, opened 2023-02-28
Improve output from deployApp()

### LINKED ISSUE #669: More information from `deployApp()`

hadley:
* Add some bigger structure with `cli::cat_rule()`
* Inform about appName and destination
* Don't display bundle/application ID so many times
* List uploaded files
* Strip timestamps from connect logs, highlight headings, and display on stdout, not stderr

Fixes #669

### DIFF
--- a/R/connect.R
+++ b/R/connect.R
@@ -163,15 +163,17 @@ connectClient <- function(service, authInfo) {
       while (TRUE) {
         path <- paste0(file.path("/tasks", taskId), "?first_status=", start)
         response <- handleResponse(GET(service, authInfo, path))
+        messages <- unlist(response$status)
 
-        labeledMessage <- function(msg) {
-          message(paste("[Connect]", msg))
-        }
+        # Strip timestamps, if found
+        timestamp_re <- "\\d{4}/\\d{2}/\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d{3,} "
+        messages <- gsub(timestamp_re, "", messages)
 
-        if (length(response$status) > 0) {
-          lapply(response$status, labeledMessage)
-          start <- response$last_status
-        }
+        # Made headers more prominent.
+        heading <- grepl("^# ", messages)
+        messages[heading] <- cli::style_bold(messages[heading])
+        cat(paste0(messages, "\n", collapse = ""))
+
         if (length(response$finished) > 0 && response$finished) {
           return(response)
         }
--- a/R/connect.R
+++ b/R/connect.R
@@ -163,16 +163,21 @@ connectClient <- function(service, authInfo) {
       while (TRUE) {
         path <- paste0(file.path("/tasks", taskId), "?first_status=", start)
         response <- handleResponse(GET(service, authInfo, path))
-        messages <- unlist(response$status)
 
-        # Strip timestamps, if found
-        timestamp_re <- "\\d{4}/\\d{2}/\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d{3,} "
-        messages <- gsub(timestamp_re, "", messages)
+        if (length(response$status) > 0) {
+          messages <- unlist(response$status)
 
-        # Made headers more prominent.
-        heading <- grepl("^# ", messages)
-        messages[heading] <- cli::style_bold(messages[heading])
-        cat(paste0(messages, "\n", collapse = ""))
+          # Strip timestamps, if found
+          timestamp_re <- "\\d{4}/\\d{2}/\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d{3,} "
+          messages <- gsub(timestamp_re, "", messages)
+
+          # Made headers more prominent.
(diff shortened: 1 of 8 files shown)


R/deployApp.R
```diff
@@ -191,7 +191,7 @@ deployApp <- function(appDir = getwd(),
     check_directory(appSourceDoc)
     recordDir <- appSourceDoc
   } else if (!is.null(recordDir)) {
-    check_directory(recordDir)
+    check_file(recordDir)
```
hadley, line 194: Badly named argument strikes again

R/deployApp.R
```diff
@@ -288,23 +292,26 @@ deployApp <- function(appDir = getwd(),
     python <- getPythonForTarget(python, accountDetails)
     pythonConfig <- pythonConfigurator(python, forceGeneratePythonEnvironment)
 
+    if (!quiet) {
+      cli::cli_alert_info("Bundling files: {.file {appFiles}}")
```
hadley, line 296: cli will automatically truncate this list if it's long

R/connect.R
```diff
@@ -163,15 +163,17 @@ connectClient <- function(service, authInfo) {
       while (TRUE) {
         path <- paste0(file.path("/tasks", taskId), "?first_status=", start)
         response <- handleResponse(GET(service, authInfo, path))
+        messages <- unlist(response$status)
 
-        labeledMessage <- function(msg) {
-          message(paste("[Connect]", msg))
-        }
+        # Strip timestamps, if found
+        timestamp_re <- "\\d{4}/\\d{2}/\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d{3,} "
```
aronatkins, line 169: The `[Connect]` preamble was trying to make it more clear what information was produced on the client and what information was sent back from the server. Do you feel like this separation is still clear?

R/deployApp.R
```diff
@@ -288,23 +292,26 @@ deployApp <- function(appDir = getwd(),
...
-        contentCategory = contentCategory,
-        verbose = verbose,
-        pythonConfig = pythonConfig,
-        quarto = quarto,
-        isCloudServer = isCloudServer(accountDetails$server),
-        metadata = metadata,
-        image = image
-      )
-
+    withStatus("Uploading bundle", {
```
aronatkins, line 314: Is it silly to have `logger` immediately followed by `withStatus`, which also logs?

R/deployApp.R
aronatkins: `bundleApp` can be (relatively) expensive. It would be nice to have similar to `withStatus` which marks the start/end of this type of operation (with timestamps). The existing `withStatus` is a little too awkward for this purpose, especially since `bundleApp` can produce output -- its "DONE" message expects it is emitted on the same line as the original status message.

Not for this PR, for sure; idle musing about better ways to mark the start and end of different processing phases.

R/deployApp.R
```diff
@@ -324,29 +331,28 @@ deployApp <- function(appDir = getwd(),
     bundle <- application$deployment$bundle
   }
 
-  if (length(bundle$id) > 0 && nzchar(bundle$id)) {
```
aronatkins, line 327: Do you think that we still log enough (early) information about the deployment target? Output from rsconnect is often attached to support tickets, and breadcrumbs like these are occasionally helpful.

aronatkins (approved):

R/connect.R
hadley: I think so, because the rules help delineate the output from connect:

Screenshot 2023-02-28 at 09 01 21

Maybe the rule could have a more informative title? I can't think what it should say though.

R/deployApp.R
hadley: Oh I assumed that `bundleApp()` would be cheap. Is it just the cost of tar-ing up a bunch of files?

In general, would you be ok with moving to something that looks more like:

```
* Bundling files...
✅ Bundling complete
```

The dangling line makes all sorts of output (including errors) look wonky.

R/deployApp.R
hadley: Yeah, let me try again.

R/deployApp.R
aronatkins: `bundleApp` includes manifest creation, which means dependency discovery (finding explicit package references, then exploding the entire package requirement tree). It has been expensive in the past. https://github.com/rstudio/packrat/pull/615, for example.

hadley:
I'm sure there's still more work to be done, but I think this looks pretty good and it sets up a good foundation for the future:

Screenshot 2023-02-28 at 11 19 50

PR #719, opened 2023-02-28
More deployment refactoring

hadley:
* Remove unused code from `deploymentRecord()`, and return a list instead of data frame
* Perform all record value standardisation in `deploymentRecord()`
* Add more tests

To do:
* [ ] After review, organise functions in more logical order

### DIFF
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -1,36 +1,3 @@
-saveDeployment <- function(recordDir,
-                           target,
-                           application,
-                           bundleId = NULL,
-                           hostUrl = serverInfo(target$server)$hostUrl,
-                           metadata = list()) {
-
-  deployment <- deploymentRecord(
-    name = target$appName,
-    title = target$appTitle,
-    username = target$username,
-    account = target$account,
-    server = target$server,
-    hostUrl = hostUrl,
-    appId = application$id,
-    bundleId = bundleId,
-    url = application$url,
-    metadata = metadata
-  )
-  path <- deploymentConfigFile(recordDir, target$appName, target$account, target$server)
-  writeDeploymentRecord(deployment, path)
-
-  # also save to global history
-  addToDeploymentHistory(recordDir, deployment)
-
-  invisible(NULL)
-}
-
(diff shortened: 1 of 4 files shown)


R/deployments.R
```diff
@@ -121,76 +113,55 @@ deployments <- function(appPath = ".",
...
+    appId = appId %||% "",
+    bundleId = bundleId %||% "",
+    url = url %||% "",
+    when = when,
+    lastSyncTime = lastSyncTime
+  )
+  c(standard, metadata)
 }
 
+# Workbench uses to show a list of recently deployed content on user dashboard
```
aronatkins, line 150: I had no idea that this was a thing!

R/deployments.R
```diff
@@ -121,76 +113,55 @@ deployments <- function(appPath = ".",
...
-  write.dcf(deploymentRecord, newHistory, width = 4096)
-  cat("\n", file = newHistory, append = TRUE)
+  newHistory <- deploymentHistoryPath(new = TRUE)
+  writeDeploymentRecord(deploymentRecord, newHistory)
 
+  history <- deploymentHistoryPath()
   # append existing history to new history
-  if (file.exists(history))
+  if (file.exists(history)) {
+    cat("\n", file = newHistory, append = TRUE)
```
aronatkins, line 162: Why did the addition of a newline shift from its current location? Are we leaving a newly created DCF file without a trailing newline?

Expand the tests to confirm the behavior when creating the file and when appending to the file.

aronatkins (approved):

R/deployments.R
hadley: This should be the newline that separates records in the dcf file; you're right that I should test the behaviour.

aronatkins: Thanks. I'd like to see the actual file contents, but it feels as if we have lost a trailing newline on newly created files.

My theory:

New: `second-record\nfirst-record`
Old: `second-record\nfirst-record\n`

hadley: I'm pretty sure `write.dcf()` adds that already, but I'll definitely double check.

PR #720, opened 2023-02-28
Log startup script execution by default

### LINKED ISSUE #542: rsconnect profile options ignored
When using rsconnect version 0.8.24, the settings in these files aren't taking effect:

```
$R_HOME/etc/rsconnect.site
~/.rsconnect_profile
$PROJECT/.rsconnect_profile
```

The two settings used were:

```
options(rsconnect.check.certificate = FALSE)
options(rsconnect.max.bundle.size = 5000000000)
```

Downgrading to the previous version used, 0.8.16, resolved the issue.

hadley:
Fixes #542

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -205,7 +205,7 @@ deployApp <- function(appDir = getwd(),
   displayStatus <- displayStatus(quiet)
 
   # run startup scripts to pick up any user options and establish pre/post deploy hooks
-  runStartupScripts(appDir, logLevel)
+  runStartupScripts(appDir, quiet = quiet, verbose = verbose)
 
   # at verbose log level, turn on all tracing options implicitly for the
   # duration of the call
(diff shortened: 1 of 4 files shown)


R/deployApp.R
```diff
@@ -611,22 +611,22 @@ openURL <- function(client, application, launch.browser, on.failure, deploymentS
...
-    if (file.exists(script)) {
-      if (logLevel == "verbose") {
-        cat("----- Sourcing startup script", script, "-----\n")
-      }
-      source(file = script, verbose = (logLevel == "verbose"))
-    }
+    taskStart(quiet, "Running {script}")
+
+    env <- new_environment(parent = globalenv())
+    source(script, verbose = verbose, local = env)
```
aronatkins, line 630: Idle musing: This isn't a current use case, but we could eventually have a stable environment across all of these user tasks, which might let folks (somehow) build state during the "pre" step that can then get consumed during the "post" step..

aronatkins (approved):

PR #722, opened 2023-03-01
Generate index earlier

### LINKED ISSUE #106: manifest.json is missing generated index.htm from files property
The `index.htm` file is created by writeRmdIndex:
https://github.com/rstudio/rsconnect/blob/0a350894eaf4c7addb9424cad03a4f0348132461/R/rmdindex.R#L4

which is called from `bundleApp`:
https://github.com/rstudio/rsconnect/blob/aac8077b3d9d0d056d75e42093e70ad1d757ac38/R/bundle.R#L67

Unfortunately, this is after we create the `manifest.json`:
https://github.com/rstudio/rsconnect/blob/aac8077b3d9d0d056d75e42093e70ad1d757ac38/R/bundle.R#L59-L64

As a result, the `index.htm` file is not included in `manifest.files` and therefore never makes it into the Docker container used by shinyapps.io when running the application.

hadley:
Fixes #106

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -466,6 +466,10 @@ bundleApp <- function(appName,
       appPrimaryDoc = appMetadata$appPrimaryDoc)
   on.exit(unlink(bundleDir, recursive = TRUE), add = TRUE)
 
+  # if necessary write an index.htm for shinydoc deployments
+  logger("Writing Rmd index if necessary")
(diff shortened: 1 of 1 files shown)


aronatkins (approved):
Conditionally approved, but we may remove `index.htm` support entirely.

hadley:
Yeah, let's not worry about this PR then.

PR #723, opened 2023-03-02
Remove writeRmdIndex

### LINKED ISSUE #106: manifest.json is missing generated index.htm from files property

### LINKED ISSUE #109: index.htm is incorrectly generated when single-file publishing in a multi-file directory
We choose to generate an `index.htm` file based on the number of files in the `appDir`:
https://github.com/rstudio/rsconnect/blob/1665cb875eda99e06be21ffcdbe5010e480b67fb/R/rmdindex.R#L16-L17

This is incorrect, as the caller may have chosen a subset of those files to include in the deployment bundle.

We should only look at files already included in the bundle when deciding if we need to generate an `index.htm` file.

hadley:
Fixes #106. Fixes #109.


aronatkins (approved):
Could we NEWS this to remind our future selves when the change happened?

Issue #667, opened 2023-02-13
Better tools for environment variables


hadley:
* Upload selected env vars on first release (avoid fail on first deploy problem)
* Scan all projects and list which env vars are used (particularly needed when credentials get rolled)

(These are both somewhat bandaids over bigger auth issues but I think they're worth doing)

hadley:
Should record env vars to update in the deployment record; that way if we implement a dry run feature, we could copy only those env vars to the dry environment.

hadley:
Currently env vars are only supported on Posit connect, so don't need to worry about other end points yet.

PR #727, opened 2023-03-03
Bundle .Rprofile with documents

hadley:
This is important because .Rprofile is the only way to control various knitr settings, and is used when you run `rmarkdown::render()` locally.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # rsconnect 0.8.30 (development version)
 
-* `deployDoc()` includes a `.Rprofile` in the bundle, if one found in the 
+* `deployDoc()` includes a `.Rprofile` in the bundle, if one is found in the 
   same directory as the document.
 
 * Removed Rmd generation code (`writeRmdIndex()`) which had not worked, or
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # rsconnect 0.8.30 (development version)
 
+* `deployDoc()` includes a `.Rprofile` in the bundle, if one found in the
```
aronatkins, line 3: nit "if one _is_ found"

R/deployDoc.R
```diff
@@ -53,6 +53,11 @@ standardizeSingleDocDeployment <- function(path,
     taskComplete(quiet, "Document dependencies discovered")
 
     appFiles <- c(basename(path), resources$path)
+
+    if (file.exists(file.path(dirname(path), ".Rprofile"))) {
```
aronatkins, line 57: What about `.Renviron`?

aronatkins (approved):

R/deployDoc.R
aronatkins: I can see an argument both ways for `.Renviron`...

hadley: I think I give the most weight to the fact that `.Renviron` usually contains secrets so they shouldn't be uploaded to a file that other people could access. Plus my goal is to provide more tools for env vars in https://github.com/rstudio/rsconnect/issues/667.

PR #728, opened 2023-03-03
Export standardizeAppFiles()

hadley:
This gives a natural place to document all of the details of `appFiles` & `appFileManifest`.

* Supersedes `listBundleFiles()`
* Add `appFileManifest` argument to `writeManifest()`, since that makes the new docs align better. That required some cleaning up of the tests


R/bundleFiles.R
```diff
@@ -1,6 +1,38 @@
...
+#'
+#' @description
+#' Given an app directory, and optional `appFiles` and `appFileManifest`
+#' arguments, returns vector of paths to bundle in the app.
+#'
+#' When neither `appFiles` nor `appFileManifest` is supplied,
+#' `standardizeAppFiles()` will include all files under `appDir`, apart
+#' from the following:
+#'
+#' *  Certain files and folders that don't need to be bundled, such as
```
aronatkins, line 11: "such as like"

R/bundleFiles.R
```diff
@@ -1,6 +1,38 @@
...
+#'    like version control, internal config, and RStudio state, are
+#'    automatically excluded.
+#'
+#' *  You can exclude additional files by listing them in in a `.rscignore`
+#'    file. This file must have one file or directory per line (with path
+#'    relative to the current directory). It doesn't support wildcards, or
+#'    ignoring files in subdirectories.
+#'
+#' `standardizeAppFiles()` will error if the total file size exceeds the
```
aronatkins, line 20: nit: "`standardizeAppFiles()` will produce an error" or "`standardizeAppFiles()` will err" or "`standardizeAppFiles()` errs"

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # rsconnect 0.8.30 (development version)
 
+* New `standardizeAppFiles()`, which supsersedes `listBundleFiles()`.
```
aronatkins, line 3: This isn't my favorite name for a public function, but I don't have great suggestions. Are `enumerateTargetFiles` or `listTargetFiles` too similar to `listBundleFiles`?

Should the documentation hint about the common pattern for Rmd, which uses `rmarkdown::find_external_resources` to enumerate its requirements and populate `appFIles`?

aronatkins (approved):

NEWS.md
aronatkins: Something like `listDeploymentFiles` feels most fitting. I would like to eventually use "app" less. Use "content" where we must, but in contexts like this one, it does not feel necessary.

PR #732, opened 2023-03-06
Move `checkAppLayout()` into `appMetadata()`

hadley:
I think it makes sense to keep all of the code dealing with determining app structure in one place, especially since `checkAppLayout()` and `inferAppMode()` are tightly coupled.

There are two minor downsides to this change:

* There's no longer a way to opt-out of this check with `lint = FALSE`. But that seems unlikely to affect much code and this isn't really linting.
* The check occurs a little later in `deployApp()`, but still well before deployment.


R/appMetadata.R
```diff
@@ -59,6 +60,53 @@ appMetadata <- function(appDir,
...
+
+  if (any(satisfiedLayouts)) {
+    return()
+  }
+
+  cli::cli_abort(c(
+    "Cancelling deployment: invalid project layout.",
+    i = "The project should have one of the following layouts:",
+    " " = "1. 'server.R' and 'ui.R' in the application base directory.",
+    " " = "2. 'server.R' and 'www/index.html' in the application base directory.",
```
aronatkins, line 99: Presenting this list of layouts is misleading, since R Markdown using `runtime: shinyrmd` can have `server.R` and `ui.Rmd` (other Rmd names also fine). The description of layouts also does not hint at R Markdown sites (bookdown).

The `satisfiedLayouts` checks check that the project looks to have the minimal structure for something that _might_ successfully deploy. Listing them in text makes the reader feel like they need to confirm to that exact structure.

I'm aware that you're just moving code around.

R/appMetadata.R
```diff
@@ -7,6 +7,7 @@ appMetadata <- function(appDir,
                         metadata = list()) {
 
   appFiles <- standardizeAppFiles(appDir, appFiles)
+  checkAppLayout(appDir, appPrimaryDoc)
```
aronatkins, line 10: Moving this "lint-check" here means that we have already created a deployment target on the remote server. If this check is firing, that means that folks most likely are trying to deploy the wrong thing; we've just created something that they will need to remove on the remote server.

aronatkins (commented):
I'm not sure that we want to take this change, since deployment targets will be created for mistakes which should never be deployed.

R/appMetadata.R
hadley: Ah, good catch. Then let's maybe leave this to mature for a while, until `appMetadata()` is computed before we create the deployment target. (Which I think I'm gradually moving towards anyway.)

hadley: Can you mark this as "needs changes" so I don't get over zealous and move it?

aronatkins (changes requested):
early linting prevents unnecessary content creation

hadley:
`appMetadata()` now called earlier and `checkAppLayout()` is a little less specific in its error message (in the long run, I hope `checkAppLayout()` can align better with `inferAppMode()`)

R/deployApp.R
```diff
@@ -344,15 +354,6 @@ deployApp <- function(appDir = getwd(),
     pythonConfig <- pythonConfigurator(python, forceGeneratePythonEnvironment)
 
     logger("Inferring App mode and parameters")
-    appMetadata <- appMetadata(
```
aronatkins, line 347: Did the associated `logger` call get left behind?

R/deployApp.R
```diff
@@ -344,15 +354,6 @@ deployApp <- function(appDir = getwd(),
...
-      appDir = appDir,
-      appFiles = appFiles,
-      appPrimaryDoc = appPrimaryDoc,
-      quarto = quarto,
-      contentCategory = contentCategory,
-      isCloudServer = isCloudServer,
-      metadata = metadata
-    )
```
aronatkins, line 355: The `upload==FALSE` feels like a forgotten workflow. The amount of unnecessary work that is done when `upload==FALSE` surprises me; "redeploy without uploading" does not feel like it should interleave with the primary deploy workflow.

Should we make the separation between "upload" and "redeploy" more explicit?

Not a today task, but this code-movement made the additional work more obvious, since this call is moving out of the `upload==TRUE` path.

R/appMetadata.R
```diff
@@ -60,6 +61,50 @@ appMetadata <- function(appDir,
...
+
+  if (any(satisfiedLayouts)) {
+    return()
+  }
+
+  cli::cli_abort(c(
+    "Cancelling deployment: invalid project layout.",
+    i = "Expecting one of the following publication types:",
+    " " = "1. A Shiny app with `app.R` or `server.R` + `ui.R`",
+    " " = "2. An R Markdown (`.Rmd`) or Quarto (`.qmd`) document.",
```
aronatkins, line 100: nit: Maybe pluralize "document".

aronatkins (approved):

R/deployApp.R
hadley: What does deploy without upload mean?

R/appMetadata.R
aronatkins: It looks like it only applies to shinyapps.io and ends up running the `deployApplication` against whatever we have stored as `bundle <- application$deployment$bundle`.

What this probably means is that rsconnect should run the "deployApplication" endpoint against whatever was the most recent bundle. I'm not sure about its history or if it still has a purpose.

@kippandrew, do you know if the `upload=FALSE` workflow (where the deploy happens without an upload) is valuable?

For Connect, the server supports such a thing, even though it's not made available today by rsconnect. If you call the "deploy" endpoint without a bundle identifier, Connect will "activate" the most recent bundle. That can be helpful if, for example, a systems change has been made that allows some package to successfully install. Typically, folks just upload again.

R/deployApp.R
aronatkins: It looks like it only applies to shinyapps.io and ends up running the `deployApplication` against whatever we have stored as `bundle <- application$deployment$bundle`.

What this probably means is that rsconnect should run the "deployApplication" endpoint against whatever was the most recent bundle. I'm not sure about its history or if it still has a purpose.

@kippandrew, do you know if the `upload=FALSE` workflow (where the deploy happens without an upload) is valuable?

For Connect, the server supports such a thing, even though it's not made available today by rsconnect. If you call the "deploy" endpoint without a bundle identifier, Connect will "activate" the most recent bundle. That can be helpful if, for example, a systems change has been made that allows some package to successfully install. Typically, folks just upload again.

PR #735, opened 2023-03-06
Add more interactive selections

### LINKED ISSUE #691: Interactively prompt user for server/account?
In `findAccount()`. 

Some existing work in this in `getDefaultServer()` used in `connectApiUser()` and `connectUser()`.

hadley:
* `findAccount()` offers you a choice when neither `server` nor `account` are supplied.
* `findServer()` replaces `getDefaultServer()` (revealing a few bugs in empty server case)
* `deploymentTarget()` gives prompt in ambiguous case

Fixes #691.


R/deploymentTarget.R
```diff
@@ -59,15 +59,31 @@ deploymentTarget <- function(recordPath = ".",
...
+    )
+    idx <- cli_menu(
+      "This directory has been previously deployed in multiple places.",
+      "Which deployment do you want to use?",
+      choices = choices,
+      not_interactive = not_interactive,
+      error_call = error_call
+    )
+    createDeploymentTarget(
+      appDeployments$name[[idx]],
```
aronatkins, line 80: Does it make sense to use a temporary and `xx[[idx]]` once rather than seven distinct indexing operations? This is effectively the same code as on line 52; is there an obvious way to avoid that repetition?

R/servers.R
```diff
@@ -105,7 +102,7 @@ cloudServerInfo <- function(name = "posit.cloud") {
 #' @export
 discoverServers <- function(quiet = FALSE) {
   # TODO: Better discovery mechanism?
-  discovered <- getOption("rsconnect.local_servers", "http://localhost:3939/__api__")
+  discovered <- getOption("rsconnect.local_servers", NULL)
```
aronatkins, line 105: Yet another feature I had never encountered. I do not know what "auto-discovery" means or how it might be configured by administrators. The old default _probably_ uses that value in case RSP and RSC are installed onto the same server, but I don't have enough context to be sure.

First landed: https://github.com/rstudio/rsconnect/commit/83642fca83e7bd8f8f0ee45a521a0e1ab8390b16

I can find no references outside the `rsconnect` repository and cannot find any documentation for the `rsconnect.local_servers` option.

My vote is to deprecate this function unless we can find additional context.

aronatkins (approved):

R/servers.R
hadley: It looks like it was added to facilitate testing, and was never documented, so I agree that deprecating seems safe. I'll do that in a separate PR to avoid making this one more complicated.

R/deploymentTarget.R
hadley: Maybe the way forward is to add a function that updates an existing deployment record. Let me try that and see how it feels.

R/servers.R
```diff
@@ -123,30 +120,26 @@ discoverServers <- function(quiet = FALSE) {
...
-  if (nrow(existing) > 1 && prompt && interactive()) {
-    name <- as.character(existing[1, "name"])
-    message("Registered servers: ", paste(existing[, "name"], collapse = ", "))
-    input <- readline(paste0(
-      "Which server (default '", name, "')? "))
-    if (nchar(input) > 0) {
-      name <- input
-    }
-    return(serverInfo(name))
+findServer <- function(prompt = TRUE) {
```
hadley, line 123: Need to add `server` argument so can do argument checking like in `findAccount()`

R/servers.R
hadley: Now in #737

aronatkins (approved):

PR #736, opened 2023-03-06
Inline `lucidClientForAccount()` into `clientForAccount()`

hadley:
And slightly normalize usage throughout package.

@aronatkins the logic took a bit of unpicking here, but `serverInfo()` already handled the three types of server, so we can call that first to get all the info we need.


R/configureApp.R
```diff
@@ -143,7 +142,7 @@ unsetProperty <- function(propertyName, appPath = getwd(), appName = NULL,
   # resolve the application target and target account info
   target <- deploymentTarget(appPath, appName, NULL, NULL, account)
   accountDetails <- accountInfo(target$account)
-  lucid <- lucidClientForAccount(accountDetails)
+  lucid <- clientForAccount(accountDetails)
```
aronatkins, line 145: nit: These "client" instances are named `lucid`. I'm fairly sure that these functions apply to shinyapps.io and posit.cloud (they're both hosted by the "lucid" code-base, but that's a company-internal name). We could just call them `client`.

tests/testthat/test-client.R
```diff
@@ -0,0 +1,15 @@
+test_that("cloud accounts create cloud clients", {
```
aronatkins, line 1: Could also test creating a Connect client (even if that's a fake hostname)?

aronatkins (approved):

PR #737, opened 2023-03-06
Update servers.R

hadley:
* Deprecate `discoverServer()`
* Simplify `servers()` implementation and test that it works with no servers
* Code style improvements to `shinyappsServerInfo()` and `cloudServerInfo()`.
* Redact certificate in `serverInfo()`


R/servers.R
```diff
@@ -125,11 +119,6 @@ discoverServers <- function(quiet = FALSE) {
 
 getDefaultServer <- function(local = FALSE, prompt = TRUE) {
    existing <- servers(local)
-   # if there are no existing servers, silently try to discover one to work with
```
aronatkins, line 128: observation: In looking at the callers to `getDefaultServer`, I noticed that `connectUser` and `connectApiUser` have quite a bit of shared code, including the calls which err when no server target is found. Some of that shared logic could move into this function.

R/servers.R
```diff
@@ -229,21 +218,20 @@ removeServer <- function(name) {
...
   if (identical(name, "shinyapps.io")) {
-    return(shinyappsServerInfo())
-  }
+    info <- shinyappsServerInfo()
+  } else if (identical(name, cloudServerInfo(name)$name)) {
+    info <- cloudServerInfo(name)
+  } else {
+    configFile <- serverConfigFile(name)
+    if (!file.exists(configFile))
+      stop(missingServerErrorMessage(name))
```
aronatkins, line 228: nit: braces always (existing code lacked them, but since this PR modified these lines ...)

aronatkins (approved):

R/servers.R
hadley: Note; I'll reduce that duplication in a future PR.

PR #738, opened 2023-03-07
Refactor `cloudServerInfo()`

hadley:
* Eliminates `shinyappsServerInfo()`
* Use `isCloudServer()` in more places


R/applications.R
```diff
@@ -159,7 +159,7 @@ applicationTask <- function(taskDef, appName, account, server, quiet) {
 # streams application logs from ShinyApps
 streamApplicationLogs <- function(authInfo, applicationId, entries, skip) {
   # build the URL
-  url <- paste0(shinyappsServerInfo()$url, "/applications/", applicationId,
+  url <- paste0(serverInfo("shinyapps.io")$url, "/applications/", applicationId,
```
aronatkins, line 162: This should not rely on the "shinyapps.io" server name and instead rely on some target server object. We're getting away with this today because shinyapps.io and posit.cloud use the same API hostnames.

The caller (`showLogs`) has an appropriate client; maybe log streaming should be implemented there, alongside `getLogs`?

R/client.R
```diff
@@ -5,7 +5,7 @@ clientForAccount <- function(account) {
 
   if (isShinyappsServer(account$server)) {
     shinyAppsClient(serverUrl, account)
-  } else if (account$server == cloudServerInfo(account$server)$name) {
+  } else if (isCloudServer(account$server)) {
```
aronatkins, line 8: Note to future us: The `shinyapps.io` server record returns `TRUE` for `isCloudServer`; you may also need `isShinyappsServer` to distinguish from other cloud server records.

R/servers.R
```diff
@@ -38,52 +38,41 @@
...
-  # We encode the current and prior product names here and call this function to
-  # see if a configured server identifier references the cloud product.
-  if (!is.element(name, c("posit.cloud", "rstudio.cloud"))) {
-    name <- "posit.cloud"
-  }
-  list(
+cloudServerInfo <- function(name) {
+  name <- arg_match0(name, cloudServers)
+
+  data.frame(
```
aronatkins, line 69: I've found it a little challenging to track the ever-changing response type boundaries - things return a `list`, then a `data.frame`, then back to a `list`, then a `data.frame` again. There's not quite that much churn, but there has been quite a bit of adjustment.

tests/testthat/_snaps/servers.md
```diff
@@ -15,15 +15,15 @@
     Output
       List of 3
        $ name       : chr "posit.cloud"
-       $ certificate: Amazon... (redacted)
        $ url        : chr "https://api.shinyapps.io/v1"
+       $ certificate: Amazon... (redacted)
```
aronatkins, line 19: Is there any consequence to reordering this data frame?

R/client.R
aronatkins: Annoyances:

1. `account$server` is the name of the server, not a server object. 
1. The `server` variable name can be either a name or an object, depending on context.

aronatkins (approved):

R/applications.R
hadley: I think this is probably missing a method on the client object, but it's going to be a while before we get around to fixing that. IMO this code is no worse that the previous, because all server logic is now routed through `serverInfo()`

R/servers.R
hadley: Yeah, I don't think we're done refactoring here, but I think this logic is clearer because now all servers are handled in `serverInfo()` and `servers()`, and the special cases are slightly fewer.

tests/testthat/_snaps/servers.md
hadley: Shouldn't be — most uses will be via variable name, and this change makes the variable order consistent across all invocations.

aronatkins: I like grouping the human-friendly fields together.

PR #741, opened 2023-03-07
Eliminate manual use of `cloudClient()`

hadley:
@aronatkins I missed two places where we were previously using `cloudClient()` and I've updated them to use `clientForAccount()` instead. That allows us to simplify the UI a bit because `cloudClient()` does more work for us.


R/accounts.R
```diff
@@ -51,7 +51,7 @@ accounts <- function(server = NULL) {
 #'
 #' @family Account functions
 #' @export
-connectApiUser <- function(account = NULL, server = NULL, apiKey = NULL, quiet = FALSE) {
+connectApiUser <- function(account = NULL, server = NULL, apiKey, quiet = FALSE) {
```
hadley, line 54: I'm pretty sure this is required; it certainly doesn't work without it.

aronatkins: Agreed. This is part of the public package API; NEWS?

aronatkins (approved):

PR #742, opened 2023-03-07
More servers()/serverInfo() refactoring

hadley:
* New `serverNames()` function that lists names of registered servers
* `servers()` now just combines the results of `serverInfo()` applied to server names
* Split up docs for metadata and modifying functions
* Use `findServer()` in `removeServer()` and `serverInfo()`


R/accounts.R
```diff
@@ -417,6 +417,17 @@ str.rsconnect_secret <- function(object, ...) {
   cat(" ", format(object), "\n", sep = "")
 }
 
+#' @export
+as.data.frame.rsconnect_secret <- function(x,
+                                           ...) {
```
aronatkins, line 422: style nit/question: Is this indenting intentional? The deep indent for a line starting with `...` feels inconsistent with many of the other multi-line function signatures, which would place the `x` on a separate line.

R/servers.R
```diff
@@ -139,14 +150,44 @@ findServer <- function(server = NULL,
...
+#' * `addServer()` registers a server.
+#' * `addConnectServer()` registers a Posit connect server. Once it has been
+#'   registered, you can connect to an account on the server using
+#'   [connectUser()].
+#' * `removeServer()` removes a server from the registry.
+#' * `addServerCertificate()` adds a certificate to a server.
+#'
+#' @param name Optional nickname for the server. If none is given, the nickname
+#'   is inferred from the server's hostname.
+#' @param url Server's URL. Should look like `http://servername/` or
```
aronatkins, line 167: Could you suggest HTTPS in the examples?

The server URL is the base URL for your server. Some proxies may use an additional sub-path, like `https://proxy.company.com/connect/`.

The `ensureConnectServerUrl` portion of the `addConnectServer` is irritating. It would be quite nice if we could somehow figure out how to hide that beneath the "Connect" client. The `__api__` URL path prefix isn't something that is an important difference to users.

I guess what I'm saying: We might be able to figure out how to collapse `addServer` and `addConnectServer`. Not a "today" task, but it feels like unnecessary complexity to force on the user.

aronatkins (approved):

R/servers.R
hadley: What if we just deprecated `addConnectServer()` and moved `ensureConnectServerUrl()` into `addServer()`? It's not like there a non-Connect servers that you can register (since the cloud servers are automatically registered).

aronatkins: That feels right. I don't know why `addServer` and `addConnectServer` both exist. 

Maybe the IDE called `addConnectServer` at some point in time. It was added to support the IDE:
https://github.com/rstudio/rsconnect/commit/c1ef2e062c3c45b0441a5e5a914b498cbb641599

But the IDE later shifted to using `:::validateServerUrl`:
https://github.com/rstudio/rsconnect/commit/6e30d9cd750533c021ca56910b610fe4e493c590
https://github.com/rstudio/rstudio/blob/967bfed5f1f8ed617baf9933ef6fc328b5dc1df0/src/cpp/session/modules/SessionRSConnect.R#L224-L229
https://github.com/rstudio/rstudio/commit/9e9c7b46667372b65f3eeb511287643c5370ab49

I cannot find history indicating that the IDE has called either of the add-server functions. Even in 0.99, the IDE was using `:::validateServerUrl` but not explicitly registering servers. I'm completely failing to find how the IDE creates server entries.

aronatkins: CC @jmcphers - would you mind helping us understand how the IDE constructs rsconnect "server" records? Does it directly manipulate the DCF files? Something else?

jmcphers: The IDE doesn't manipulate any of rsconnect's state directly (it always uses package functions). The `validateServerUrl` function has the side effect of registering the URL if it turns out to be valid. This should probably be split into separate steps, but that's how server records come to be as things are written today.

https://github.com/rstudio/rsconnect/blob/e8f0771fd6c3c16ebfbfa02683a09994b89ac654/R/ide.R#L85-L97

hadley: Coincidentally I was doing just that 😄 #746

aronatkins: Ah, thanks. I missed that `findLocalServer` had side-effects.

PR #743, opened 2023-03-07
Pass metadata around instead of individual components

hadley:
Includes drive-by extraction of `writeManifest()` tests in to `test-writeManifest.R`


R/appMetadata.R
```diff
@@ -54,8 +54,10 @@ appMetadata <- function(appDir,
     appMode = appMode,
     appPrimaryDoc = appPrimaryDoc,
     hasParameters = hasParameters,
+    contentCategory = contentCategory,
     documentsHavePython = documentsHavePython,
-    quartoInfo = quartoInfo
+    quartoInfo = quartoInfo,
+    isCloudServer = isCloudServer
```
aronatkins, line 60: `isCloudServer` feels out-of-place. It is a characteristic of the server, not the on-disk files. It's passed to `inferAppMode` so we can downgrade `rmd-static` to `rmd-shiny`.

Let's keep it as an argument to `createAppManifest`.

R/bundle.R
```diff
@@ -202,8 +195,8 @@ createAppManifest <- function(appDir,
   }
 
   # indicate whether this is a quarto app/doc
-  if (!is.null(quartoInfo) && !isCloud) {
-    manifest$quarto <- quartoInfo
+  if (!is.null(appMetadata$quartoInfo) && !appMetadata$isCloudServer) {
```
aronatkins, line 198: Do you know if shinyapps.io errs if we include the `quarto` manifest field? It would be nice to remove the `isCloudServer` check here and unconditionally add Quarto to the manifest.

I can't recall if they started parsing the Quarto field when they added Python.

aronatkins (approved):

R/bundle.R
hadley: Hmmmm, I added it unconditionally and deployed to shinyapps and it worked. Do you think that's sufficient evidence to proceed?

aronatkins: That's awesome. I think so, but let's double-check.

@ssinnott - is your manifest parser more permissive than it was previously? I seem to remember that it used to err if it encountered unexpected fields. This change might see `rsconnect` include a top-level `quarto` field / object.

ssinnott: So I was trying to do this on my end from the branch and I ran into this message when trying to set the account (it seems to work on 0.8.29).

```
> setAccountInfo(name='new-account-13', token='<stuff>', secret='<stuff>')
Error in serverInfo$name : object of type 'closure' is not subsettable
```

We do validate the top level elements of the manifest so adding quarto there could be problematic. If you have deployed something successfully and have the application id I take a look at the bundles manifest to double check.

aronatkins: No, Shiny+Quarto only works today if folks deploy up-to-date HTML alongside the Shiny Qmd file.

We use `rmarkdown::run` to execute the document, which uses the already-generated HTML and avoids a render.

It's awkward and easy to end up with inconsistent doc/HTML.

The next step would be to perform some type of server-side render at the time of deploy, but that's not available yet for shinyapps.io / posit.cloud.

aronatkins: If we're able to add a `quarto` manifest field, it makes the manifest ready for the hosted offerings to eventually add support, since everything else in the bundle would already be in shape.

If they can't handle the additional field today, continue with what we've got.

hadley: If I run this on main:

```R
deployApp("tests/testthat/quarto-proj-r-shiny/", server = "shinyapps.io", quarto = Sys.which("quarto"))
```

I get https://hadley.shinyapps.io/quarto-proj-r-shiny/, which presumably fails because it doesn't have quarto installed. Would it be better to fail earlier in rsconnect? Do we expect shinyapps to add quarto support, or will that only be in posit.cloud?

aronatkins: Remember when I said "awkward"?

```bash
# In a terminal:
quarto render tests/testthat/quarto-proj-r-shiny
```

```R
deployApp("tests/testthat/quarto-proj-r-shiny/", server = "shinyapps.io", quarto = Sys.which("quarto"))
```

https://aronatkins.shinyapps.io/quarto-proj-r-shiny/

hadley: I guess in ideal world that would be detected as something other than appMode `quarto-shiny` and it wouldn't even send `quarto`.

hadley: But I can confirm that unconditionally including `quarto` leads to this error:

```
! Unhandled Exception: Child Task 1280000096 failed: Error parsing manifest: Schema validation failed: Additional properties are not allowed ('quarto' was unexpected)
```

aronatkins: That's the error I had been expecting. So sad. I guess this condition stays for now.

ssinnott: I can add an issue on our end to tweak the manifest parsing.

hadley: IMO the manifest parsing should ignore unknown fields; that way rsconnect can start sending extra data before shinyapps.io knows how to handle it.

PR #746, opened 2023-03-08
Extract out validateConnectUrl()

hadley:
* Test it and resolve problem with posit.cloud
* Use it in `addServer()`
* Add new `addTestServer()`

### DIFF
--- a/R/ide.R
+++ b/R/ide.R
@@ -1,27 +1,21 @@
-
 # These functions are intended to be called primarily by the RStudio IDE.
 
-# attempts to validate a server URL by hitting a known configuration endpoint
-# (which does not require authentication). returns a list containing (valid =
-# TRUE) and server settings, or a list containing (valid = FALSE) and an error
-# message.
-#
-# the URL may be specified with or without the protocol and port; this function
-# will try both http and https and follow any redirects given by the server.
+# This function is poorly named because as well as validating the server
+# url it will also register the server if needed.
 validateServerUrl <- function(url, certificate = NULL) {
-
   valid <- validateConnectUrl(url, certificate)
 
   if (valid$ok)  {
-    c(
-      list(valid = TRUE, url = valid$url, name = findLocalServer(url)),
-      valid$response
-    )
+    name <- findAndRegisterLocalServer(url)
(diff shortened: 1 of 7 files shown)


R/ide.R
```diff
@@ -9,6 +9,20 @@
 # the URL may be specified with or without the protocol and port; this function
 # will try both http and https and follow any redirects given by the server.
 validateServerUrl <- function(url, certificate = NULL) {
+
+  valid <- validateConnectUrl(url, certificate)
```
aronatkins, line 13: Could we document that this function has a side-effect of registering a server if it discovers that the input URL can be transformed into a valid server?

R/servers.R
```diff
@@ -179,13 +181,19 @@ addConnectServer <- function(url, name = NULL, certificate = NULL,
...
 #' @rdname addConnectServer
 #' @export
-addServer <- function(url, name = NULL, certificate = NULL, quiet = FALSE) {
+addServer <- function(url, name = NULL, certificate = NULL, validate = TRUE, quiet = FALSE) {
   check_string(url)
   check_name(name, allow_null = TRUE)
 
-  serverUrl <- parseHttpUrl(url)
+  if (validate) {
+    out <- validateConnectUrl(url, certificate)
```
aronatkins, line 189: Sneaky indirection. `addServer -> validateConnectUrl -> findLocalServer -> addServer`.

Which means: When validating, `addServer` has already been called. Return immediately (or somehow restructure things to not be so sneaky).

aronatkins (changes requested):
When validating (with the new flag), `addServer` writes the record twice.

R/servers.R
hadley: `validateConnectUrl()` doesn't call `findLocalServer()`; `validateUrl()` does. `validateUrl()` needs a better name, but we can't change it because the IDE depends on it.

aronatkins: Ugh. Sorry. These names!

aronatkins (approved):

R/servers.R
hadley: I've done what I can to improve the comments and clarify what function names I can affect.

PR #747, opened 2023-03-08
Mild http refactoring

hadley:
* Restyle with tidyverse style
* Eliminate `httpInvokeRequest()` which was only used in one place
* Extract out code duplicated between `httpRequest` and `httpRequestWithBody`


R/http.R
```diff
@@ -354,6 +326,58 @@ queryString <- function(elements) {
...
+  if (identical(protocol, "https")) {
+    createCertificateFile(certificate)
+  } else {
+    NULL
+  }
+}
+
+# Auth --------------------------------------------------------------------
+
+authHeaders1 <- function(authInfo, method, url, file) {
```
hadley, line 353: Auth in `httpRequest` and `httpRequestWithBody` appears to have slightly drifted apart, but I think the two functions are equivalent, and since `authHeaders1()` is slightly simpler, I'd suggest standardising on it.

R/http.R
```diff
@@ -131,12 +146,13 @@ POST_JSON <- function(service,
                       query = NULL,
                       headers = list()) {
   POST(service,
-       authInfo,
-       path,
-       query,
-       "application/json",
-       content = toJSON(json, pretty = TRUE, digits = 30),
-       headers = headers)
+    authInfo,
```
aronatkins, line 149: nit: should `service` also be on its own line? i guess the same question applies to the function definitions. any time we end up with a series of enumerated, newline-separated arguments, the first should also be on its own line, not with the opening paren.

i find myself doing this for any nontrivial function calls and definitions, like the `POST` call here or its surrounding `POST_JSON`.

on the other hand, i find myself avoiding many-line argument lists for `cat` and `sprintf`. that's no rule, though. :)

R/http.R
```diff
@@ -354,6 +326,58 @@ queryString <- function(elements) {
...
 
+buildPath <- function(apiPath, path, query = NULL) {
+  # prepend the service path
+  url <- paste(apiPath, path, sep = "")
+
+  # append the query
+  if (!is.null(query)) {
+    # URL encode query args
+    query <- utils::URLencode(query)
+    url <- paste(url, "?", query, sep = "")
```
aronatkins, line 337: it surprises me that we construct our own query string and concatenate it with the URL, but maybe that's necessary because the different transports don't all accept query lists in similar ways.

R/http.R
aronatkins: Here is a path with certificate-only `authInfo`:
https://github.com/rstudio/rsconnect/blob/e9ec50cfc2c5c797afddf8f7ebe10cedc8b041ca/R/ide.R#L50-L53

Added in https://github.com/rstudio/rsconnect/commit/96ba75b944b0e35a88dc7b8a940233739b96eb4f

It looks like this is the only case where `authInfo` is not an account, but something else, and might be the only instance where it contains a certificate.

In other cases, we use `clientForAccount` to obtain a client, which augments the `account` object with a certificate.

It looks as if the other auth checks (those in `authHeaders1`) won't fire when the other fields (secret, private key, api key) are missing from `authInfo`. I agree; it looks like we can ditch `authHeaders2`.

Could we add tests to help document the fact that `authInfo` is usually more than an account (because `certificate` is added), but sometimes only has `certificate`?

aronatkins (approved):

R/http.R
hadley: Yeah, agreed. I think this is a remnant of an older style that styler supports that we no longer like.

R/http.R
```diff
@@ -336,7 +306,9 @@ urlDecode <- function(x) {
 }
 
 urlEncode <- function(x) {
-  if (inherits(x, "AsIs")) return(x)
+  if (inherits(x, "AsIs")) {
+    return(x)
+  }
```
hadley, line 311: I'm confused that this function works, given that it depends on RCurl, which is only in suggests.

aronatkins: Also, sometimes we use `utils::urlEncode` and the the private `urlDecode` uses `curl::curl_unescape`.

A partial shift away from RCurl happened in https://github.com/rstudio/rsconnect/pull/348, and RCurl was still an import dependency at that point in time, moving to suggests in https://github.com/rstudio/rsconnect/commit/35ba0497f7266da2c9157cbcd1b55e29c2876b4d

Also surprised.

PR #748, opened 2023-03-08
Update http tests

hadley:
* Use webfakes package instead of homegrown solution
* Move tests to individual http files to make iteration easier
* Add test of libCurl `PUT` streaming

### DIFF
--- a/tests/testthat/test-http-libcurl.R
+++ b/tests/testthat/test-http-libcurl.R
@@ -11,7 +11,11 @@ test_that("basic HTTP methods work", {
 test_that("can stream PUT content from disk", {
   service <- httpbin_service()
 
-  path <- withr::local_tempfile(lines = c("1", "2", "3"))
+  path <- withr::local_tempfile()
+  con <- file(path, "wb")
+  writeLines(c("1", "2", "3"), con = con)
+  close(con)
+
   resp <- PUT(
     service,
     authInfo = NULL,


tests/testthat/test-http-libcurl.R
```diff
@@ -0,0 +1,26 @@
...
+  withr::local_options(rsconnect.http = "libcurl")
+
+  test_http_GET()
+  test_http_POST_JSON()
+  test_http_POST_empty()
+  test_http_POST_file()
+  test_http_headers()
+})
+
+test_that("can stream PUT content from disk", {
```
aronatkins, line 11: Future: Test PUT, PUT_JSON, DELETE across all transports.

tests/testthat/helper-http.R
```diff
@@ -0,0 +1,77 @@
...
+  expect_equal(resp$status, 200)
+
+  contents <- handleResponse(resp)
+  expect_equal(contents$data, "1\n2\n3\n")
+}
+
+test_http_headers <- function() {
+  service <- httpbin_service()
+
+  resp <- GET(service, authInfo = list(apiKey = "abc123"), path = "get")
```
aronatkins, line 72: Should the header-check be against all methods? Otherwise, we're not validating that the auth information is passed to the underlying `httpRequest*` helper and received by the HTTP server.

tests/testthat/test-http-libcurl.R
aronatkins: Future: Validate that the correct content-type header is set.

tests/testthat/helper-http.R
aronatkins: We should confirm that the httpbin service received a GET request with the correct path (and possibly query args).

aronatkins (approved):

tests/testthat/test-http-libcurl.R
hadley: I only tested `PUT` for libcurl because it's the only wrapper that has a custom path for PUT compared to the other methods.

tests/testthat/helper-http.R
hadley: I'll add a test for a `POST` since that goes through a different path `httpRequestWithBody()`.

The method and path are somewhat entangled; if `/get` isn't called with a GET, it'll error.

tests/testthat/helper-http.R
```diff
@@ -0,0 +1,77 @@
+cache <- new_environment()
+
+httpbin_service <- function() {
+  app <- env_cache(
+    cache,
+    "test_app",
+    webfakes::new_app_process(webfakes::httpbin_app())
```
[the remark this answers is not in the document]
aronatkins: It's installed in CI because we have `any::rcmdcheck` in our workflow, and that package requires `callr`.

https://github.com/rstudio/rsconnect/blob/38c19e93170a56e27137652d36113e580b98d32e/.github/workflows/R-CMD-check.yaml#L49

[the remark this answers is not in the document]
hadley: Ah, good catch.

PR #750, opened 2023-03-08
Deprecate `addConnectServer()`

hadley:


R/servers.R
```diff
@@ -199,7 +161,7 @@ addServer <- function(url, name = NULL, certificate = NULL, validate = TRUE, qui
   # of its URL
   if (is.null(name)) {
     name <- serverUrl$host
-    if (!quiet && interactive()) {
+    if (!quiet && is_interactive()) {
```
hadley, line 164: Makes this function behave better in interactive tests.

aronatkins (approved):

PR #751, opened 2023-03-09
Automatically handle responses

hadley:
Almost every case where we call `GET`, `POST` etc, we immediately handle the response. Now we do that automatically, as a prelude to handling redirects globally.

For now, I've attached the "raw" response as an attribute, but I think we can eliminate that in a few more steps.


R/http.R
```diff
@@ -73,6 +76,65 @@ httpRequestWithBody <- function(service,
...
+    } else {
+      json <- list()
+    }
+
+    if (response$status %in% 200:399) {
+      out <- json
+    } else if (!is.null(json$error)) {
+      reportError(json$error)
+    } else {
+      reportError(paste("Unexpected json response:", response$content))
```
aronatkins, line 107: Note to self: `readHttpResponse` ensures that `response$content` is a character vector, not a connection, so it is OK to use its value repeatedly.

Given that we have already parsed the content, we could pretty-print the object, but that might introduce representation adjustments (because R). OK using the raw `content` here.

R/ide.R
```diff
@@ -47,11 +47,12 @@ validateConnectUrl <- function(url, certificate = NULL) {
   retry <- TRUE
   while (retry) {
     tryCatch({
-      httpResponse <- GET(parseHttpUrl(url),
+      response <- GET(parseHttpUrl(url),
                           list(certificate = certificate),
                           settingsEndpoint,
                           timeout = getOption("rsconnect.http.timeout", 10))
 
+      httpResponse <- attr(response, "httpResponse")
```
aronatkins, line 55: Note to self: It confused me at first when the result from `GET` was not a response object. If callers to `GET` need to react to its status, maybe we should think about augmenting the lower-level `result` object with a `json` field.

```r
response <- GET(...)
result <- response$json
status <- response$status
```

That's a pretty broad change, and excessive for the majority of cases, which do not want to react to status codes.

aronatkins (approved):

PR #753, opened 2023-03-09
Focus bundlePackages() on bundling; make DESCRIPTION handling more clear

hadley:
By extracting out `needsR()` and `inferRPackageDependencies()` to be called from `createAppManifest()`. Also adds to `appDependencies()` so that it now includes inferred dependencies.


R/appDependencies.R
```diff
@@ -61,9 +62,48 @@
...
+  # Currently R is only supported by the "knitr" engine, not "jupyter" or
+  # "markdown"
+  is.null(appMetadata$quartoInfo) ||
+    "knitr" %in% appMetadata$quartoInfo[["engines"]]
+}
+
+inferRPackageDependencies <- function(appMetadata) {
+  deps <- switch(appMetadata$appMode,
+    "rmd-static" = c("rmarkdown", if (appMetadata$hasParameters) "shiny"),
+    "quarto-static" = "rmarkdown",
```
aronatkins, line 99: Note to self: I was curious about whether or not we could make "quarto-static" need only "knitr" rather than "rmarkdown". Even though Quarto clams to use the "knitr" engine, there are references to the rmarkdown package in the Quarto source. I cannot easily tell if any of those references are needed other than for Shiny document execution with `rmarkdown::run`.

TL;DR: Keep `rmarkdown` as a dependency for `quarto-static` that uses the `knitr` engine.

R/bundlePackage.R
```diff
@@ -1,34 +1,17 @@
...
-    hasParameters = hasParameters,
-    documentsHavePython = documentsHavePython
-  )
-  deps <- snapshotRDependencies(appDir, inferredRDependencies, verbose = verbose)
   checkBundlePackages(deps, call = error_call)
 
+  copyPackageDescriptions(bundleDir, deps$Package)
   deps$description <- lapply(deps$Package, function(nm) {
     # Remove packageDescription S3 class so jsonlite can serialize
-    # TODO: should we get description from packrat/desc folder?
```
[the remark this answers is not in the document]
aronatkins: We also could update `deps$description` first and then write that information to disk; that would give us one source of truth (the manifest), and one set of calls to `packageDescription`.

Not a firm recommendation.

R/bundlePackage.R
```diff
@@ -74,63 +57,15 @@ checkBundlePackages <- function(deps, call = caller_env()) {
...
-    })
-  }, error = function(e) {
-    warning("Unable to package DESCRIPTION files: ", conditionMessage(e), call. = FALSE)
-    if (dirExists(descDir)) {
-      unlink(descDir, recursive = TRUE)
-    }
-  })
+  dir.create(descDir, showWarnings = FALSE, recursive = TRUE)
+
+  descPaths <- file.path(find.package(packages), "DESCRIPTION")
```
aronatkins, line 67: The try/catch and file-existence checks feel like they could help identify malformed libraries, but I don't know that those cases have ever been tripped.

The previous pattern was in place in the initial implementation, rather than reacting to reported issues:
https://github.com/rstudio/rsconnect/pull/42

aronatkins (approved):

R/bundlePackage.R
hadley: `checkBundlePackages()` catches the most common case of the package not being installed and if your library is sufficiently messed up that the `DESCRIPTION` isn't installed, I suspect you'll get an upstream error (i.e. packrat will also have problems).

R/bundlePackage.R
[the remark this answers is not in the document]
hadley: But then you have to round trip from DCF -> list -> DCF and worry about the potential problems that might cause. (Which I don't think are major; it's just not a clear win to me, because you'd still want to carefully test that code.)

PR #754, opened 2023-03-09
Handle HTTP redirects

### LINKED ISSUE #674: libCurl HTTP handler doesn't handle redirects
Because of 

```
curl::handle_setopt(handle, followlocation = FALSE)
```

This is probably low-priority, but it caused a confusing error for me when attempting to publish to colarado.rstudio.com which has been renamed to colorado.posit.co

hadley:
Fixes #674

Uploading to shinyapps.io works without redirects because the result of `client$updateBundleStatus()`, which needs the redirect, was not used.

@aronatkins any ideas for testing? I verified that this works with an interactive deployment to shinyapps, and the `httpRequest()` path is tested via `validateConnectUrl()`. But currently no automated tests hit the `httpRequestWithBody()`  path.  Would you mind also verifying that connect/shinyapps doesn't emit any 307 or 308 redirects that I should be respecting?

### DIFF
--- a/R/http.R
+++ b/R/http.R
@@ -33,6 +33,21 @@ httpRequest <- function(service,
     timeout = timeout,
     certificate = certificate
   )
+
+  while (httpResponse$status >= 300 && httpResponse$status < 400) {
+    service <- parseHttpUrl(httpResponse$location)
+    httpResponse <- http(
+      protocol = service$protocol,
+      host = service$host,
+      port = service$port,
+      method = method,
+      path = service$path,
+      headers = headers,
+      timeout = timeout,
+      certificate = certificate
+    )
+  }
+
   handleResponse(httpResponse, error_call = error_call)
 }
 
@@ -60,7 +75,7 @@ httpRequestWithBody <- function(service,
   }
 
   path <- buildPath(service$path, path, query)
-  headers <- c(headers, authHeaders(authInfo, method, path, file))
+  authed_headers <- c(headers, authHeaders(authInfo, method, path, file))
   certificate <- requestCertificate(service$protocol, authInfo$certificate)
 
   # perform request
@@ -71,18 +86,35 @@ httpRequestWithBody <- function(service,
     port = service$port,
     method = method,
     path = path,
-    headers = headers,
+    headers = authed_headers,
     contentType = contentType,
     contentFile = file,
     certificate = certificate
   )
+  while (httpResponse$status >= 300 && httpResponse$status < 400) {
+    # This is a simplification of the spec, since we should preserve
+    # the method for 307 and 308
+    # https://www.rfc-editor.org/rfc/rfc9110.html#name-redirection-3xx
+    service <- parseHttpUrl(httpResponse$location)
+    authed_headers <- c(headers, authHeaders(authInfo, "GET", service$path))
+    httpResponse <- http(
+      protocol = service$protocol,
+      host = service$host,
+      port = service$port,
+      method = "GET",
+      path = service$path,
+      headers = authed_headers,
+      certificate = certificate
+    )
+    httpResponse
+  }
+
   handleResponse(httpResponse, error_call = error_call)
 }
 
 handleResponse <- function(response, error_call = caller_env()) {
   reportError <- function(msg) {
-    req <- response$req
-    url <- paste0(req$protocol, "://", req$host, req$port, req$path)
+    url <- buildHttpUrl(response$req)
 
     cli::cli_abort(
       c("<{url}> failed with HTTP status {response$status}", msg),
@@ -293,6 +325,10 @@ parseHttpUrl <- function(urlText) {
   url$port <- components[[4]]
   url$path <- components[[5]]
   url
+}
+
+buildHttpUrl <- function(x) {
+  paste0(x$protocol, "://", x$host, x$port, x$path)
 }
 
 urlDecode <- function(x) {
--- a/R/ide.R
+++ b/R/ide.R
@@ -17,84 +17,45 @@ validateServerUrl <- function(url, certificate = NULL) {
 # The URL may be specified with or without the protocol and port; this function
 # will try both http and https and follow any redirects given by the server.
 validateConnectUrl <- function(url, certificate = NULL) {
-  tryAllProtocols <- TRUE
-
-  if (!grepl("://", url, fixed = TRUE))
-  {
+  # Add protocol if missing, assuming https except for local installs
+  if (!grepl("://", url, fixed = TRUE)) {
     if (grepl(":3939", url, fixed = TRUE)) {
-      # assume http for default (local) connect installations
       url <- paste0("http://", url)
     } else {
-      # assume https elsewhere
       url <- paste0("https://", url)
     }
   }
-
-  # if the URL ends with a port number, don't try http/https on the same port
-  if (grepl(":\\d+/?$", url)) {
-    tryAllProtocols <- FALSE
-  }
-
-  settingsEndpoint <- "/server_settings"
   url <- ensureConnectServerUrl(url)
+  is_http <- grepl("^http://", url)
 
-  # populate certificate if supplied
-  certificate <- inferCertificateContents(certificate)
+  GET_server_settings <- function(url) {
+    auth_info <- list(certificate = inferCertificateContents(certificate))
+    GET(
+      parseHttpUrl(url),
+      auth_info,
+      "/server_settings",
+      timeout = getOption("rsconnect.http.timeout", 10)
+    )
+  }
 
-  # begin trying URLs to discover the correct one
   response <- NULL
-  errMessage <- ""
-  retry <- TRUE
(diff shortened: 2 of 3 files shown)


R/http-libcurl.R
```diff
@@ -78,40 +79,22 @@ httpLibCurl <- function(protocol,
     con <- file(contentFile, "rb")
     on.exit(if (!is.null(con)) close(con), add = TRUE)
 
-    if (identical(method, "POST")) {
```
hadley, line 81: This is possibly overly aggressive, but the comment doesn't make sense to me as the 3xx redirects only apply to the result of the upload, not the upload itself. This is possibly a misdiagnosis of an older problem, because I don't think there should be a general difference between PUTs and POSTs. It's possible that this might have also resolved #544.

R/http-libcurl.R
```diff
@@ -122,19 +105,10 @@ httpLibCurl <- function(protocol,
...
-    error = function(e, ...) {
-      # ignore errors resulting from timeout or user abort
-      if (identical(e$message, "Callback aborted") ||
-          identical(e$message, "transfer closed with outstanding read data remaining"))
-        return(NULL)
-      # bubble remaining errors through
-      else
-        stop(e)
-    }))
+  time <- system.time(
```
hadley, line 108: I also simplified this since the user interrupt error message has changed so that had no effect and I'm pretty sure that "transfer closed" (https://github.com/curl/curl/blob/37a3c63427e84a56703e19b6ab7c78fa52e566a2/lib/transfer.c#L1215-L1228) is a legit error we do want to bubble up.

R/http.R
```diff
@@ -71,18 +86,35 @@ httpRequestWithBody <- function(service,
...
     contentFile = file,
     certificate = certificate
   )
+  while (httpResponse$status >= 300 && httpResponse$status < 400) {
+    # This is a simplification of the spec, since we should preserve
+    # the method for 307 and 308
+    # https://www.rfc-editor.org/rfc/rfc9110.html#name-redirection-3xx
+    service <- parseHttpUrl(httpResponse$location)
+    authed_headers <- c(headers, authHeaders(authInfo, "GET", service$path))
+    httpResponse <- http(
```
aronatkins, line 100: This redirect handling is assuming that the original request (and its payload) do not need to be re-sent. Is that the situation you're seeing?

It feels like the redirection that's set-up for colorado might demand a replay. 

```
curl -v http://colorado.rstudio.com/rsc/__api__/server_settings
#> HTTP/1.1 301 Moved Permanently
#> Location: https://colorado.rstudio.com/rsc/__api__/server_settings
curl -v https://colorado.rstudio.com/rsc/__api__/server_settings
#> HTTP/1.1 302 Found
#> Location: http://colorado.posit.co/rsc/__api__/server_settings
```

The original request never landed.

Some systems support 100-continue request/response to avoid sending the payload before receiving confirmation; I'm not sure that any of these servers support that workflow.

Most resources will indicate that you should not automatically replay the request in cases like colorado - a shift from one hostname to another - without explicit confirmation.

We probably want at least:
1. Be transparent about the redirect.
2. Offer an option to err rather than walk through the redirect.

R/http.R
```diff
@@ -355,6 +391,7 @@ authHeaders <- function(authInfo, method, path, file = NULL) {
   }
 }
 
+# https://github.com/rstudio/connect/wiki/token-authentication#request-signing-rsconnect
```
aronatkins, line 394: fyi: this is an internal link.

aronatkins (commented):
I'll dig into some of the possible response codes tomorrow.

R/http-libcurl.R
hadley: Ok, I'll have a go with some uploading some really large files and see if I can recreate it.

R/http.R
aronatkins: The colorado examples give us two motivating examples:
1. HTTP => HTTPS
2. Host rename

I'm not sure that there's a good answer, especially since the colorado example shows us that 301/302 are much more common than 307/308.

Thinking out loud:

* Automatic replay of all redirects requires no user intervention, which is a smooth experience.
* Server URLs are left stale, meaning every subsequent request hits the redirect penalty.
* Server names are left stale, meaning the UI / metadata will always reflect the old name.
* Account records reflect the stale server name.

If we ignore updating the server and account records for a moment, we could run a pre-flight check against the client URL to make sure it is still "correct" and update the URL in-memory. That would correct the following requests without hitting multiple redirects and would give us a single place to ask for confirmation, log, or reject the redirect.

In other words, run something like `validateServerUrl` before using the client to do meaningful work.

hadley: Notes to self: need something like `validateConnectServer()` that we run in `deployApp()` before hitting any endpoints. If url has redirected, need to prompt the user to update the record on disk (if interactive), or otherwise warn or continue.  Might want to consider automatically updating if the hostname stays the name.

hadley: While making coffee this morning, I suddenly realised that we can't just update an existing server record on disk, because that will break existing deployments. We need some way to also "redirect" existing server records to a new server record. So I won't tackle that here, but consider it in a future PR.

DESCRIPTION
```diff
@@ -28,6 +28,7 @@ Imports:
     lifecycle,
     openssl (>= 2.0.0),
     packrat (>= 0.6),
+    prettyunits,
```
hadley, line 31: I only use this in one place, but it's also a very lightweight dep. I filed a cli issue so we can eventually get rid of it: https://github.com/r-lib/cli/issues/588

R/http-libcurl.R
```diff
@@ -78,40 +79,24 @@ httpLibCurl <- function(protocol,
...
-           bin
-        })
-    } else {
-      # why was a file specified for this endpoint?
-      warning("Content file specified, but not used because the '", method, "' request ",
-              "type does not accept a body.")
-    }
+    curl::handle_setopt(
+      handle,
+      noprogress = fileLength <= 10 * 1024^2,
```
hadley, line 84: Show a progress bar when uploading more than 10 Mb and increase the buffer size from 64 kb to 1 Mb.

R/http-libcurl.R
hadley: I uploaded a 1 Gb app to shinyapps and Posit connect, and didn't see any problems. I'll follow up on the rpubs issue later since it's already individually tracked.

R/http.R
aronatkins: Yeah; the association between the server records and the accounts and deployments records came up during our chat. The pre-flight can ask if folks should proceed through the redirect, but it's not straightforward to rewrite the records.

R/http-libcurl.R
aronatkins: What happens in non-interactive contexts? Is the progress output reasonable if the `rsconnect::deployApp` is captured by some CI system or otherwise sent to disk?

hadley: I've reverted the buffer size, so it'll create a line every 64kb; so I'll make it only show when interactive.

R/http-libcurl.R
hadley: I uploaded a 33 Mb html with no problems. You can see it in all its glory at https://rpubs.com/hadley/large-file.

R/http.R
hadley: Tracking this in #760.

R/http-libcurl.R
hadley: Also tested updating, since that uses a `PUT` instead of `POST`.

aronatkins: Was that a large Rmd or a large bundle? Because we gzip, it's harder to get a large bundle. I ended up using a specific document when last playing with that issue, but it's likely changed since I tried.. I think I have a copy locally somewhere.  https://github.com/rstudio/rsconnect/issues/450#issuecomment-780114862

R/http-libcurl.R
```diff
@@ -78,40 +79,25 @@ httpLibCurl <- function(protocol,
...
-      # why was a file specified for this endpoint?
-      warning("Content file specified, but not used because the '", method, "' request ",
-              "type does not accept a body.")
-    }
+    progress <- is_interactive() && fileLength >= 10 * 1024^2
+
+    curl::handle_setopt(
+      handle,
+      noprogress = !progress,
+      upload = TRUE,
```
aronatkins, line 87: Have we lost `post = TRUE`, which uses a `POST` request? Is that redundant with the `curl::handle_setopt(handle, customrequest = method)`?

aronatkins (approved):

R/http-libcurl.R
hadley: @aronatkins From the progress bar 😄 ` Uploaded 23796573 bytes...` so ~22 Mb.

R/http-libcurl.R
aronatkins: OK dropping it.

PR #755, opened 2023-03-10
Improve handling of user supplied `appId`

### LINKED ISSUE #701: `deploymentTarget()` doesn't use `appId`

hadley:
* Ignore existing deployment records and retrieve from the server
* Warn if `appId` and `appName` are both specified
* Improve documentation of deployment records

Fixes #701

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -4,19 +4,25 @@
 #' [RMarkdown][rmarkdown::rmarkdown-package] document, a plumber API, or HTML
 #' content to a server.
 #'
-#' ## Updating existing apps
+#' ## Deployment records
 #'
-#' If you have previously deployed an app, `deployApp()` will do its best to
-#' update the existing deployment. In the simple case where you have only one
-#' deployment in `appDir`, this should just work with the default arguments.
-#' If you want multiple deployments to the same server, supply `appName`.
-#' If you want multiple deployments to different servers, use `account` and/or
-#' `server`.
+#' When deploying an app, `deployApp()` will save a deployment record that
+#' makes it easy to update the app on server from your local source code. This
+#' generally means that you need to only need to supply important arguments
+#' (e.g. `appName`, `appTitle`, `server`/`account`) on the first deploy, and
+#' rsconnect will reuse the same settings on subsequent deploys.
 #'
-#' The metadata needs to make this work is stored in the `rsconnect/` directory
-#' beneath `appDir`. You should generally check these files into version
-#' control to ensure that future you and other collaborators will publish
-#' to the same location.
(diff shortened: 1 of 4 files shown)


R/deployApp.R
```diff
@@ -252,15 +265,27 @@ deployApp <- function(appDir = getwd(),
...
+      recordPath = recordPath,
+      appName = appName,
+      appTitle = appTitle,
+      account = account,
+      server = server,
+      forceUpdate = forceUpdate
+    )
+  } else {
+    if (!is.null(appName)) {
+      cli::cli_warn("{.arg appName} is ignored when {.arg appId} is set")
```
hadley, line 279: @aronatkins this is not quite what you suggested, but your idea seemed like a lot of work and I think this is fairly close. I also like being able to say "if you specified `appId`, then existing deployments are ignored.

R/deploymentTarget.R
```diff
@@ -72,6 +70,29 @@ deploymentTarget <- function(recordPath = ".",
...
 }
 
+deploymentTargetForApp <- function(appId,
+                                   appTitle = NULL,
+                                   account = NULL,
+                                   server = NULL) {
+  accountDetails <- accountInfo(account, server)
+  client <- clientForAccount(accountDetails)
+
+  withCallingHandlers(
```
aronatkins, line 80: This looks really similar to the (now uncalled? IDE-only?) `getAppById`.

R/deploymentTarget.R
```diff
@@ -72,6 +70,29 @@ deploymentTarget <- function(recordPath = ".",
   }
 }
 
+deploymentTargetForApp <- function(appId,
+                                   appTitle = NULL,
+                                   account = NULL,
+                                   server = NULL) {
+  accountDetails <- accountInfo(account, server)
+  client <- clientForAccount(accountDetails)
```
aronatkins, line 78: aside: `clientForAccount` might be the right place for the URL validation pre-flight we chatted about, even though we might construct multiple client objects during the deploy workflow.

aronatkins (approved):

R/deploymentTarget.R
hadley: Looks like it's still use: https://github.com/rstudio/rstudio/blob/ee56d49b0fca5f3d7c3f5214a4010355d1bb0212/src/cpp/session/modules/SessionRSConnect.R#L213-L214. But should move it to `ide.R`

hadley: I've made a note to rework `getAppById()`

hadley:
No automated tests, but I tried updating an app on connect, and it worked.

PR #758, opened 2023-03-12
Rewrite `getAppById()`

hadley:
This moves `getAppById()` and its helper `accountInfoFromHostUrl()` to `ide.R`, and re-implements from first principles.

While writing tests for the new function, I realised it would be simpler to properly simulate the actual set of on disk server and account records, so I wrote a new `addTestAccount()` function and make `addTestServer()` easier to use in tests by rearranging the arguments and tweaking the defaults. I then updated the existing tests to use these new functions.


tests/testthat.R
```diff
@@ -1,18 +1,4 @@
 library(testthat)
 library(rsconnect)
 
-run_tests <- function() {
```
hadley, line 4: This is no longer needed because we do it per test, where necessary. And that also means that you can run the tests interactively without messing with your personal server/account config.

R/ide.R
```diff
@@ -147,3 +147,31 @@ showRstudioSourceMarkers <- function(basePath, lint) {
                       basePath = basePath,
                       autoSelect = "first")
 }
+
+# getAppById() -----------------------------------------------------------------
+
+# https://github.com/rstudio/rstudio/blob/ee56d49b0fca5f3d7c3f5214a4010355d1bb0212/src/gwt/src/org/rstudio/studio/client/rsconnect/ui/RSConnectDeploy.java#L699
+
+getAppById <- function(id, account, server, hostUrl) {
```
aronatkins, line 155: Findings when trying to determine if it was safe to remove the `NULL` default values:

When first added, `hostUrl` was not an argument passed by the IDE (the 0.99 series):
https://github.com/rstudio/rstudio/commit/47a0fbf1c7e0f1c45355eb5e343966b336c5334d

The 1.1 IDE builds include hostUrl.
https://github.com/rstudio/rstudio/commit/d99f8ea6d98e6c2e7cfb2b0b058f3c726d8ed6db

Removing the default values makes this NEWS-worthy, right? Is that considered a breaking change?

aronatkins (approved):

R/ide.R
hadley: This is an internal function so only affects the IDE.  Do you think that 0.99 is recent enough that I should ensure it works without `hostUrl`?

tests/testthat.R
hadley: It should be picked up by CI, since we did some work to ensure that check is run on GHA.

R/ide.R
aronatkins: No need to maintain hostUrl as a maybe-null value. RStudio 1.0 was 2016. I was only reminding us of the timeline.

PR #761, opened 2023-03-13
Use cli_rule() instead of cat_rule()

### LINKED ISSUE #756: latest github build of `rsconnect` log output now excludes deploy url
We're updating our automated CI tests in Connect to use the latest github builds for the `rsconnect` package and running into an issue.  Our tests grab the deploy url from Connect to check if the deploy succeeded via `httr`.  However, it looks like `rsconnect::deployApp()` output was updated and now excludes the line with the deployment URL.

To capture the output we use:
```
original_response <- capture.output(rsconnect::deployApp(appDir = "./bundles/deployApp/alpha",
                                                         account = CONNECT_ADMIN_USERNAME,
                                                         server = "test_box",
                                                         appName = app_name[1]))
```

The tail end of that output now looks like:

```
[156] "curl: HTTP 200 https://cran.rstudio.com//src/contrib/Archive/shiny/shiny_1.7.1.tar.gz"                                 
[157] "Caching shiny."                                                                                                        
[158] "Using cached shiny."                                                                                                   
[159] "\tOK (built source)"                                                                                                   
[160] "Completed packrat build against R version: '3.6.3'"                                                                    
[161] "Launching Shiny application..."                                                                                        
[162] "── Deployment complete ────────────────────────────────────────────────────────────────────" 
```

However when manually deploying, in the console I see:

```
Using cached shiny.
	OK (symlinked cache)
Completed packrat build against R version: '3.6.3'
Launching Shiny application...
── Deployment complete ────────────────────────────────────────────────────────────────────
✓ Successfully deployed to <http://localhost:3939/content/c6685c63-1b99-4db0-be93-8d9b41b78f3d/>
```

Can the deploy log continue to output the deploy URL so we can programmatically capture it? Or is there another way to grab the deployment URL?

hadley:
This ensures that the rules go to the same stream as the rest of the feedback. Fixes #756.

@aronatkins I'm not sure if you want more here. It would be nice to include the deployed url in the output, but it's currently a logical vector, so if we wanted to avoid breaking existing code, we'd need to add it as an attribute, which doesn't seem super appealing. Another alternative would be to implement something like `lastDeployment()` which would read the most recent deployment from the global deployment history.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -259,7 +259,7 @@ deployApp <- function(appDir = getwd(),
   }
 
   if (!quiet) {
-    cli::cat_rule("Preparing for deployment")
+    cli::cli_rule("Preparing for deployment")
   }
 
   # determine the deployment target and target account info
@@ -389,14 +389,14 @@ deployApp <- function(appDir = getwd(),
   }
 
   if (!quiet) {
-    cli::cat_rule("Deploying to server")
+    cli::cli_rule("Deploying to server")
   }
   task <- client$deployApplication(application$id, bundle$id)
   taskId <- if (is.null(task$task_id)) task$id else task$task_id
   # wait for the deployment to complete (will raise an error if it can't)
   response <- client$waitForTask(taskId, quiet)
   if (!quiet) {
-    cli::cat_rule("Deployment complete")
+    cli::cli_rule("Deployment complete")
   }
 
   # wait 1/10th of a second for any queued output get picked by RStudio


aronatkins (approved):
#756 documents that feedback like the target URL are not present in the `capture.output(rsconnect::deployApp())` result. This isn't changing how the URL is emitted, which was the original complaint.

Should NEWS indicate the change to using `message` output?

hadley:
I'd generally consider the output stream to which messages are sent to be an internal implementation detail. I certainly don't mind adding it to the news bullets, but it's not something I'd generally do by myself.

Issue #631, opened 2022-10-24
Need better support for private repositories


dbkegley:
When using a private package repository that requires authentication, some users store their credentials in `~/.Rprofile` and others store credentials by adding the private package repo directly in the IDE via `Settings > Packages > Management`. When creating a `manifest.json` with `rsconnect::writeManifest()`, the `Repository` field includes the credentials that are used to connect to the private repository.

```
https://{{USER}}:{{TOKEN}}@cran.myorg.com
```

This is a blocker for publishers who wish to perform git-backed deployments using a private package repository that requires authentication because the repo credentials would be committed to source control.

aronatkins:
There are ways to avoid using URLs-with-credentials, which avoids leaking them into your `manifest.json` (or the bundled `packrat/packrat.lock`). The most straight-forward way is probably to use `curl` for downloads with either an `.curlrc` file or the `download.file.extra` R option.

```R
options(download.file.method = "curl")

# .curlrc is preferred to the "extra" option:
options(download.file.method = "-u user:password")
```

```
# .curlrc
user = user:password
```

This may not be appropriate when multiple repository URLs are involved, but is suitable when all installation needs to use a single, authenticated repository.

hadley:
Need to confirm whether or not this works with the curl package; if not need to document `.rsconnect_profile`

PR #762, opened 2023-03-13
Deprecate all non-libcurl backends

hadley:
@aronatkins I think we agree that we want slowly move towards a single backend that is thoroughly tested. The main question is the speed and I'm happy to take your suggestion; treat this as a starting proposal.


hadley:
Maybe should resolve #631 before this one.

aronatkins (approved):

PR #764, opened 2023-03-13
Create single registerAccount() function

hadley:
And match its argument names to the fields it creates in the dcf

This code was a confusion of changing argument names, which make it hard to follow. I'm fairly confident that I've refactored this correctly because I interactively tested `connectUser()`, `connectApiUseR()`, and `setAccountInfo()`.


R/accounts.R
```diff
@@ -157,83 +157,54 @@ setAccountInfo <- function(name, token, secret, server = "shinyapps.io") {
...
   if (!is.null(info$private_key)) {
-    info$private_key <- secret(gsub("[[:space:]]", "", info$private_key))
+    info$private_key <- gsub("[[:space:]]", "", info$private_key)
   }
 
-  if (!is.null(info$secret)) {
-    info$secret <- secret(info$secret)
-  }
+  # Hide credentials
+  info$private_key <- secret(info$private_key)
```
aronatkins, line 205: Seeing this shuffle together makes me appreciate the consistency of `private_key` and `apiKey`. 😏

R/accounts.R
```diff
@@ -157,83 +157,54 @@ setAccountInfo <- function(name, token, secret, server = "shinyapps.io") {
...
+    accountId = accountId,
+    token = token,
+    secret = secret
+  )
+  invisible()
+}
+
+# A user can have multiple accounts, so iterate over all accounts looking
+# for one with the specified name
+findAccountId <- function(name, token, secret, server) {
```
aronatkins, line 174: comment indicating that this is only for shinyapps.io?

aronatkins (approved):

PR #767, opened 2023-03-13
Improve writeManifest() docs

### LINKED ISSUE #539: Incorrect content_category set by rsconnect::writeManifest() for Bookdown when published via git
From related customer support ticket:

I found out that for some reason `rsconnect::writeManifest()` doesn't set the proper content_category to the manifest generated, it is showing up as `content_category: null`, and should be `site` instead of `null`.

The customer can change this directly in the `manifest.json`, push the change to the repo and deploy again; that should work.

hadley:
And re-style file. Fixes #539

### DIFF
--- a/R/writeManifest.R
+++ b/R/writeManifest.R
@@ -1,8 +1,8 @@
-#' Create a `manifest.json` for Git-backed content
+#' Create a `manifest.json`
 #'
 #' @description
-#' Use `writeManifest()` to generate a `manifest.json` and then commit it to
-#' git to activate
+#' Use `writeManifest()` to generate a `manifest.json`. Among other things,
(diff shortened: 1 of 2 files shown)


R/writeManifest.R
```diff
@@ -1,9 +1,18 @@
-#' Create a manifest.json describing deployment requirements.
+#' Create a `manifest.json` for Git-backed content
```
aronatkins, line 1: Manifests can be useful when using the APIs, when using `rsconnect-python` (to deploy R content), or when using Git-backed content.

Let's not imply that this is only used by Git workflows.

aronatkins (approved):

PR #769, opened 2023-03-14
Tidy up utils

hadley:
* Remove unused functions
* Use `cli::cat_rule()` instead of `hr()`
* Move `activeEncoding()` to `lint.R`
* Simplify `fileMD5` by introducing a `raw` argument
* Restyle


R/http.R
```diff
@@ -422,7 +419,7 @@ signatureHeaders <- function(authInfo, method, path, file = NULL) {
     signature <- paste(openssl::base64_encode(hmac), "; version=1", sep = "")
   } else if (!is.null(authInfo$private_key)) {
     # the raw content hash is base64 encoded hex values when using private key.
-    md5 <- openssl::base64_encode(md5)
+    md5 <- openssl::base64_encode(fileMD5(file, raw = TRUE))
```
aronatkins, line 422: If we eventually get to the point where we drop the openssl dependency, this could use the `jsonlite` equivalent (since that will remain).

R/http.R
```diff
@@ -406,12 +406,9 @@ signatureHeaders <- function(authInfo, method, path, file = NULL) {
   # generate date
   date <- rfc2616Date()
 
-  # generate contents hash as hex values.
-  md5 <- fileMD5(file)
-
   if (!is.null(authInfo$secret)) {
     # the content hash is a string of hex characters when using secret.
-    md5 <- md5.as.string(md5)
+    md5 <- fileMD5(file)
```
aronatkins, line 411: Confirmed that this change still performs at most one MD5 pass (expensive for large files).

aronatkins (approved):

PR #771, opened 2023-03-14
Mild tidying testing & refactoring to `httpLibCurl()`

hadley:
* Minor refactoring & restyling
* Add test that cookies are preserved across calls
* Add support for relative urls & test
* Test JSON tracing


R/http-libcurl.R
```diff
@@ -87,75 +44,101 @@ httpLibCurl <- function(protocol,
...
+    location = headers$location,
+    contentType = contentType,
+    content = contentValue
+  )
+}
 
-  # Parse cookies from header; bear in mind that there may be multiple headers
-  cookieHeaders <- headers[names(headers) == "set-cookie"]
-  storeCookies(list(protocol = protocol, host = host, port = port, path = path), cookieHeaders)
+createCurlHandle <- function(method,
```
aronatkins, line 101: nit: Moving this function in a commit with functional code changes makes for poor diffs.

R/http-libcurl.R
```diff
@@ -87,75 +44,101 @@ httpLibCurl <- function(protocol,
...
-  }
+  contentType <- headers[["content-type"]] %||% "text/plain"
+  contentValue <- rawToChar(response$content)
 
   # emit JSON trace if requested
-  jsonTracingEnabled <-
-    httpTraceJson() &&
-    !is.null(contentFile) &&
-    identical(contentType, "application/json")
+  jsonTracingEnabled <- httpTraceJson() && contentType == "application/json"
```
aronatkins, line 83: nit: This could be a tracing function passed by the caller that receives the content-type and payload and hides the logging decisions from this function.

```r
traceResponse(contentType, contentFile, contentValue)
```

aronatkins (approved):

R/http-libcurl.R
hadley: I think the intent here is to offer something that the end user can turn on/off in case they need extra debug info.

aronatkins: Sorry. I wasn't clear: I was suggesting that we construct a function dynamically, similar to how we construct the `logger` in `deployApp`. This code may be too low-level to make that type of dependency-injection workflow straightforward.

```r
traceResponse <- makeTraceResponse()

makeTraceResponse <- function() {
    if conditions-to-allow-tracing and httpTraceJson() {
        function(contentType, contentFile, contentValue) {
            # do the real logging.
        }
    } else {
        function(contentType, contentFile, contentValue) {}
    }
}
```

This was a really soft suggestion. Maybe the more important piece is to extract the logging from the http function...

PR #772, opened 2023-03-15
More tidying up of accounts 

hadley:
Explanations in individual commits


R/accounts.R
```diff
@@ -60,74 +60,121 @@ connectApiUser <- function(account = NULL, server = NULL, apiKey, quiet = FALSE)
 
   registerAccount(
     serverName = server,
-    accountName = account %||% user$username,
+    accountName = user$username,
```
aronatkins, line 63: This change means that the `account` argument is unnecessary going forward. Something for the list.

R/accounts.R
```diff
@@ -60,74 +60,121 @@ connectApiUser <- function(account = NULL, server = NULL, apiKey, quiet = FALSE)
...
     accountId = user$id,
     apiKey = apiKey
   )
 
   if (!quiet) {
-    message("Account registered successfully: ", account %||% user$username)
+    accountId <- accountId(user$username, server)
```
aronatkins, line 69: nit: There are two notions of `accountId`: the user identifier from the API, and `username@server`. That's bound to confuse our future selves.

R/accounts.R
```diff
@@ -60,74 +60,121 @@ connectApiUser <- function(account = NULL, server = NULL, apiKey, quiet = FALSE)
...
-  repeat {
-    Sys.sleep(1)
-    user <- getAuthedUser(server, token = token)
-    if (!is.null(user))
-      break
+  if (isFALSE(launch.browser)) {
+    cli::cli_alert_warning("Open {.url {claim_url}} to authenticate")
+  } else {
+    cli::cli_alert_info("A browser window should open to complete authentication")
+    cli::cli_alert_warning("If doesn't open, then visit {.url {claim_url}}")
```
aronatkins, line 122: "if _ONE_ doesn't open"

R/accounts.R
```diff
@@ -60,74 +60,121 @@ connectApiUser <- function(account = NULL, server = NULL, apiKey, quiet = FALSE)
...
-                               user$username, "'): "))
-      if (nchar(input) > 0)
-        user$username <- input
+  # keep trying to authenticate until we're successful; server returns
+  # 500 "Token is unclaimed error" while waiting for interactive auth to complete
+  cli::cli_progress_bar(format = "{cli::pb_spin} Waiting for authentication...")
+  repeat {
+    for (i in 1:10) {
+      Sys.sleep(0.1)
+      cli::cli_progress_update()
```
aronatkins, line 131: Why 10 progress updates between attempts? Maybe one progress update per attempt with ~1s between attempts?

R/rsa.R
aronatkins, line 1: With `generateToken` removed, there's nothing "RSA" about this file... the remaining `createUniqueId` exists only to assist `generateToken` and probably should float with it.

aronatkins (approved):

R/accounts.R
hadley: Because then you get a cool spinner 😄

R/rsa.R
hadley: Ah, I missed that. And what do you think about replacing `createUniqueId(16)` with `paste(openssl::rand_bytes(16), collapse = "")` ?

hadley: (FWIW I'll probably split `accounts.R` into multiple pieces in a future PR)

R/accounts.R
hadley: Hmmm, maybe I should leave as is for now. If we fix, I think it should be part of a general re-thinking of these functions (e.g. more consistent function names, argument names, and defaults)

R/accounts.R
hadley: Yeah, I'll brainstorm a better name.

hadley: I have failed to come up with a better name; but I've made a note to keep thinking about it.

R/rsa.R
aronatkins: Any random data is fine. Watch out for R environments that might force their global random seed to a fixed state, as that means our "random token" ends up not-so-random.

Prior work:
https://github.com/rstudio/rsconnect/issues/221 / https://github.com/rstudio/rsconnect/pull/222
https://github.com/rstudio/rsconnect/issues/378 / https://github.com/rstudio/rsconnect/pull/379

hadley: Using openssl for this means that you get cryptographical secure random numbers, which is probably not important here, but they definitely affected by R's RNG state.

aronatkins: Did you mean "unaffected"? I'm +1 as long as caller cannot initialize (R or openssl) seed state to produce predictable identifiers.

PR #773, opened 2023-03-15
Servers tidying up

hadley:
* Drop interactive prompt for server name
* Extract out `registerServer()` to reduce duplicated code
* Move tests from `test-cert.R` to `test-servers.R`
* Add more tests


R/http-libcurl.R
```diff
@@ -108,10 +108,10 @@ httpLibCurl <- function(protocol,
 
   # make the request
   response <- NULL
-  time <- system.time(
-    response <- curl::curl_fetch_memory(url, handle = handle),
-    gcFirst = FALSE
-  )
+
+  start <- proc.time()
```
aronatkins, line 112: It is more interesting to obtain the elapsed time than the process time, since an HTTP request could take a long time, but consume no CPU.

aronatkins (approved):

R/http-libcurl.R
hadley: Confusingly `proc.time()` returns user, system, and elapsed time:

```R
> proc.time()
      user     system    elapsed 
   766.150    229.515 246336.283 
```

PR #779, opened 2023-03-16
Specify additional headers and cookies

### LINKED ISSUE #405: User defined cookies / headers are overwritten
For the purposes of my example, I have only tested with `libcurl`. However, hopefully we can remedy this issue with the other download handlers as well.

Straight invocation of `libcurl`:
``` r
hm <- curl::new_handle()
curl::handle_setopt(hm, .list = list(httpheader = "key:value"))
curl::curl_fetch_echo("http://localhost:32776/upload", handle = hm)$headers
#>                                                    accept 
#>                                                     "*/*" 
#>                                           accept-encoding 
#>                                           "deflate, gzip" 
#>                                                      host 
#>                                         "localhost:32776" 
#>                                                       key 
#>                                                   "value" 
#>                                                user-agent 
#> "R (3.4.3 x86_64-apple-darwin15.6.0 x86_64 darwin15.6.0)"
curl::handle_setopt(hm, .list = list(httpheader = "key2:value2"))

# NOTE: key:value missing
curl::curl_fetch_echo("http://localhost:32776/upload", handle = hm)$headers
#>                                                    accept 
#>                                                     "*/*" 
#>                                           accept-encoding 
#>                                           "deflate, gzip" 
#>                                                      host 
#>                                         "localhost:32776" 
#>                                                      key2 
#>                                                  "value2" 
#>                                                user-agent 
#> "R (3.4.3 x86_64-apple-darwin15.6.0 x86_64 darwin15.6.0)"
curl::handle_setopt(hm, .list = list(cookielist = c("key=value;key2=value2")))

# cannot see it here, but there are key=value; key2=value2 cookies defined on this request
curl::curl_fetch_echo("http://localhost:32776/upload", handle = hm)$headers
#>                                                    accept 
#>                                                     "*/*" 
#>                                           accept-encoding 
#>                                           "deflate, gzip" 
#>                                                      host 
#>                                         "localhost:32776" 
#>                                                      key2 
#>                                                  "value2" 
#>                                                user-agent 
#> "R (3.4.3 x86_64-apple-darwin15.6.0 x86_64 darwin15.6.0)"
curl::handle_setopt(hm, .list = list(httpheader = c("cookie:key=value")))

# cannot see here, but key2:value2 cookie is missing. Also, key2=value2 header missing
curl::curl_fetch_echo("http://localhost:32776/upload", handle = hm)$headers
#>                                                    accept 
#>                                                     "*/*" 
#>                                           accept-encoding 
#>                                           "deflate, gzip" 
#>                                                    cookie 
#>                                  "key=value; key2=value2" 
#>                                                      host 
#>                                         "localhost:32776" 
#>                                                user-agent 
#> "R (3.4.3 x86_64-apple-darwin15.6.0 x86_64 darwin15.6.0)"
```

<sup>Created on 2020-02-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

As you can see, the `httpheader` setting is not additive, but overwrites. It even overwrites previously specified `cookie`s. As a result, when we add the user's options here:

https://github.com/rstudio/rsconnect/blob/fef0bf6ac300e9823e4d348dffa68bcaacc8ee37/R/http-libcurl.R#L9-L12

And then don't incorporate them later:

https://github.com/rstudio/rsconnect/blob/fef0bf6ac300e9823e4d348dffa68bcaacc8ee37/R/http-libcurl.R#L121

Everything the user has specified gets overwritten.

Are there any other workarounds to specify arbitrary user-defined headers / cookies via `.rsconnect_profile`? This is something that we have a few customers stuck at when deploying to RStudio Connect (specifically when headers / cookies need to be specified to satisfy a proxy).

hadley:
Implements new `rsconnect.http.headers` and `rsconnect.http.cookies` options which allow you to set extra arbitrary additional headers/cookies on each request. Their use is documented in the new `vignette("custom-http")`.

Fixes #405.


vignettes/custom-http.Rmd
```diff
@@ -0,0 +1,88 @@
...
+Proxy settings can include a host-name, port, and username/password if necessary. The following
+are all valid values for the `http_proxy` environment variable:
+
+* `http://proxy.example.com/`
+* `http://proxy.example.com:1080/`
+* `http://username:password@proxy.example.com:1080/`
+
+## Custom headers and cookies
+
+If you need to supply additional headers or cookies, you can use the options `rsconnect.http.headers` and `rsconnect.http.cookies` respectively. `rsconnect.http.headers` needs a named vector of header names and values:
```
aronatkins, line 54: How do these configurations interact with other header values / cookies, such as auth headers and cookies sent by Connect or another service?

Any specific headers/values that folks should avoid?

aronatkins (commented):
It appears as if `clearCookies` is never called; how are old cookies cleared on exit? Are there lingering cookies within an active R session? Are they used on the next deploy call from that session?

PR #781, opened 2023-03-17
Improve package standardization

### LINKED ISSUE #659: Warn if package versions are newer than those on CRAN

hadley:
There were a lot of layers to unpeel here, but I think I've got to a place where the code both does what we want (including thinking about archived packages) and is easy to reason about.

Fixes #659

### DIFF
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -34,8 +34,8 @@
 #'   ```
 #'
 #'   Then packages installed from your corporate package repository will
-#'   have source `CORPORATE`. Posit Connect [can be configured]
-#'   (https://docs.posit.co/connect/admin/appendix/configuration/#RPackageRepository)
+#'   have source `CORPORATE`. Posit Connect
+#'   [can be configured](https://docs.posit.co/connect/admin/appendix/configuration/#RPackageRepository)
 #'   to override their repository url so that (e.g.) you can use different
 #'   packages versions on stagning and production servers.
 #'
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -67,12 +67,13 @@
 #' @inheritParams listDeploymentFiles
 #' @param appDir Directory containing application. Defaults to current working
 #'   directory.
-#' @returns A data frame one row for each depedency (direct and indirect) and
-#'   4 columns:
+#' @returns A data frame with one row for each dependency (direct, indirect,
+#'   and inferred), and 4 columns:
 #'
 #'   * `Package`: package name.
 #'   * `Version`: local version.
-#'   * `Source`: where the package was installed from, as described above.
+#'   * `Source`: a short string describing the source of the package install,
+#'      as described above.
 #'   * `Repository`: for CRAN and CRAN-like repositories, the URL to the
 #'      repository. This will be ignored by the server if it has been configured
 #'      with it's own repository name -> repository URL mapping.
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -11,19 +11,43 @@
 #' # Remote installation
 #'
 #' When deployed, the app must first install all of these packages, and
-#' rsconnect does its best to ensure that all the packages have the same
-#' version as you are running locally. It knows how to install packages from
-#' the following sources:
+#' rsconnect ensures the versions used on the server will match the versions
+#' you used locally. It knows how to install packages from the following
+#' sources:
 #'
-#' * CRAN (`CRAN`) and CRAN-like repositories (`CustomCRANLikeRepository`).
-#' * BioConductor (`Bioconductor`)
-#' * Packages installed from GitHub (`github`), GitLab (`gitlab`), or
-#'   BitBucket (`bitbucket`).
+#' * CRAN and BioConductor (`Source: CRAN` or `Source: Bioconductor`). The
+#'   remote server will ignore the specific CRAN or Bioconductor mirror that
+#'   you use locally, always using the CRAN/BioC mirror that has been configured
+#'   on the server.
 #'
-#' It does not know how to install packages that you have built and installed
-#' locally so if you attempt to deploy an app that depends on such a package
-#' it will fail. To resolve this issue, you'll need to install from a known
-#' source.
+#' * Other CRAN like and CRAN-like repositories. These packages will have
+#'   a `Source` determined by the value of `getOptions("repos")`. For example,
+#'   if you've set the following options:
+#'
+#'   ```R
+#'   options(
+#'      repos = c(
+#'        CRAN = "https://cran.rstudio.com/",
+#'        CORPORATE = "https://corporate-packages.development.company.com"
+#'      )
+#'   )
+#'   ```
+#'
+#'   Then packages installed from your corporate package repository will
+#'   have source `CORPORATE`. Posit Connect [can be configured]
+#'   (https://docs.posit.co/connect/admin/appendix/configuration/#RPackageRepository)
+#'   to override their repository url so that (e.g.) you can use different
+#'   packages versions on stagning and production servers.
+#'
+#' * Packages installed from GitHub, GitLab, or BitBucket, have `Source`
+#'   `github`, `gitlab`, and `bitbucket` respectively. When deployed, the
+#'   bundle contains the additional metadata needed to precisely recreated
+#'   the installed version.
+#'
+#' It's not possible to recreate the packages that you have built and installed
+#' from a directory on your local computer. This will have `Source: NA` and
+#' will cause the deployment to error. To resolve this issue, you'll need to
+#' install from one of the known sources described above.
 #'
 #' # Suggested packages
 #'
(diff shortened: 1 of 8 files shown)


R/bundlePackage.R
```diff
@@ -103,75 +102,72 @@ findPackageRepoAndSource <- function(records, repos) {
...
+  }
+  list(Source = source, Repository = repository)
 }
 
+findRepo <- function(availablePackages, pkg) {
+  if (pkg %in% rownames(availablePackages)) {
+    repo <- availablePackages[pkg, "Repository"]
+    # Strip /src/contrib from repository URL recorded in package record to
+    # get to repository url
+    gsub("/src/contrib$", "", repo)
```
aronatkins, line 153: It looks like `/src/contrib` is always the path appended by `contrib.url` for "source". I wasn't sure previously, which is why the old version maps from `url` to `contrib.url(url."source")`.

https://github.com/wch/r-source/blob/trunk/src/library/utils/R/packages.R#L832-L866

R/bundlePackage.R
```diff
@@ -246,37 +242,3 @@ performPackratSnapshot <- function(bundleDir, verbose = FALSE) {
...
-
-# generate a random name prefixed with "repo_".
-random.repo.name <- function() {
-  paste("repo_", paste(sample(LETTERS, 8, replace = TRUE), collapse = ""), sep = "")
-}
-
-# Given a list of optionally named repository URLs, return a list of
-# repository URLs where each element is named. Incoming names are preserved.
-# Un-named repositories are given random names.
-name.all.repos <- function(repos) {
```
aronatkins, line 265: A couple of things about repository naming:

1. The previous implementation set "Source" to the name of the repository in the "repos" list. This implementation uses the "Source" from Packrat, which is often the `Repository` field in the `DESCRIPTION`.

2. Because we were using the `repos` names, this `name.all.repos` function helped make sure that all repository URLs had names, for cases where folks use bare URLs:

    ```r
    options(repos = "https://cran.rstudio.com/")
    ```

There was some uncertainty and variability about how the `Repository` field was populated for different package repositories (CRAN, Package Manager, others), and we saw some package repositories not include `Repository`, and manifests end up with empty `Source`. I don't recall the details, but maybe using the `repos` names meant that `CustomCRANLikeRepository` wasn't in the manifest?

Connect wants a nonempty "Source" value.
If Connect sees one "Source" name with multiple URLs, that is converted into:

```r
options(repos=c(NAME="firstURL", NAME_2="secondURL", ...))
```

Related, prior work: https://github.com/rstudio/rsconnect/pull/309

aronatkins (approved):

R/bundlePackage.R
hadley: How does `Source` end up empty? I don't see any [cases in packrat](https://github.com/rstudio/packrat/blob/v0.9.0/R/pkg.R#L328) where it would not be set.

For the case of [custom repositories](https://github.com/rstudio/packrat/blob/v0.9.0/R/pkg.R#L385-L391), isn't the the repository url going to be an ok `Source` as it's already unique? 

Or are you saying that if `findRepo()` is called with an unknown `pkg` (i.e. it's not a package in a known repository) then we should return something other than `NA`?

aronatkins: Custom repositories: this change uses the `Source` coming out of Packrat, which is either the `Repository` field from the `DESCRIPTION` or `CustomCRANLikeRepository`, right? That looks to be fine.

aronatkins: Connect errs if it receives an empty `Source` field in the manifest; that's a constraint on whatever we do here.

aronatkins: You're right -- we should not see empty source, since `checkBundlePackages` will have already erred if any incoming `Source` is NA.

By ignoring the names in `options("repos")`, there may be a subtle problem.

Let's say that you have some custom repository and configure it with `options(repos = c(CORPORATE = "URL"))`. This is convention within your company.

That `CORPORATE` repository name may not be in the `DESCRIPTION` files (because you are using some custom repository). Coming out of Packrat, the "Source" today will have a value of `CORPORATE`, but with this change, will become `CustomCRANLikeRepository` (I think this is accurate from reading the code).

Your Connect configuration may contain an `RPackageRepository` configuration which rewrites `CORPORATE` repository URLs. https://docs.posit.co/connect/admin/appendix/configuration/#RPackageRepository

> This isn't a perfect strategy (because the repository names are client-controlled), but is how the feature functions.
> We want to improve this capability; it's fairly restrictive today.

Maybe: If the `DESCRIPTION` does not name the repository and we see a value of `CustomCRANLikeRepository`, it should be rewritten to the name given to the `options("repos")` repository URL.

R/bundlePackage.R
```diff
@@ -103,75 +102,89 @@ findPackageRepoAndSource <- function(records, repos) {
...
-      #
-      # The "cran" source is a well-known location for shinyapps.io.
-      #
-      # shinyapps.io isn't going to use the manifest-provided CRAN URL,
-      # but other consumers (Connect) will.
-      if (tolower(source) != "cran") {
-        source <- package.repo$name
-      }
-      repository <- package.repo$url
+  } else if (source == "CustomCRANLikeRepository") {
```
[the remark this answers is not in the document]
aronatkins: Do we need `findRepoName` here, as well?

R/bundlePackage.R
```diff
@@ -103,75 +102,89 @@ findPackageRepoAndSource <- function(records, repos) {
...
-    # validatePackageSource will emit a warning for packages with NA repository.
-    data.frame(Source = source, Repository = repository, stringsAsFactors = FALSE)
-  })
-  do.call("rbind", tmp)
+  } else if (source %in% c("CRAN", "Bioconductor")) {
+    # Installed from standard repository
+    repository <- findRepoUrl(pkg, availablePackages)
+  } else {
+    # Installed from custom repository: source is currently the repo url,
+    # but we need it to be the repo name (as defined in `getOptions("repos")`).
```
aronatkins, line 153: Am I misunderstanding the Packrat choice in `pkg.R`? I was thinking that the incoming source was the value of `packageDescription()$Repository`, not the repository URL.

hadley:
```R
install.packages('cffr', repos = c('https://ropensci.r-universe.dev', 'https://cloud.r-project.org'))
packageDescription("cffr")$Repository
#> [1] "https://ropensci.r-universe.dev"
```

`packageDescription()$Repository` is the repository URL?

hadley:
Build failures are because pak installed packages have `Repository: RSPM`. Maybe we need to special case this?

hadley:
And that's because RSPM adds that field.

R/bundlePackage.R
hadley: Oh but

```R
install.packages("filelock", repos = "http://rspm.infra/all/__linux__/bionic/latest")
packageDescription("filelock")$Repository
#> [1] "RSPM"
```

```R
install.packages("filelock", repos = "http://cloud.r-project.org")
packageDescription("filelock")$Repository
#> [1] "CRAN"
```

aronatkins: The RSPM/CRAN cases were the ones I've commonly seen. I've also seen cases where the `Repository` fields was not populated. I had not seen the ropensci case with a URL as the `Repository` field.

R/appDependencies.R
```diff
@@ -11,19 +11,43 @@
...
+#' * CRAN and BioConductor (`Source: CRAN` or `Source: Bioconductor`). The
+#'   remote server will ignore the specific CRAN or Bioconductor mirror that
+#'   you use locally, always using the CRAN/BioC mirror that has been configured
+#'   on the server.
 #'
-#' It does not know how to install packages that you have built and installed
-#' locally so if you attempt to deploy an app that depends on such a package
-#' it will fail. To resolve this issue, you'll need to install from a known
-#' source.
+#' * Other CRAN like and CRAN-like repositories. These packages will have
```
aronatkins, line 23: Thanks for capturing this; recording the current behavior will help us understand the effects of other options (e.g. preserving `Repository` from package descriptions, when available).

R/appDependencies.R
```diff
@@ -43,12 +67,15 @@
...
+#' @returns A data frame one row for each depedency (direct and indirect) and
+#'   4 columns:
+#'
 #'   * `Package`: package name.
 #'   * `Version`: local version.
-#'   * `Source`: where the package was installed from.
-#'   * `Repository`: for CRAN and CRAN-like repositories, the url to the
-#'      repo.
+#'   * `Source`: where the package was installed from, as described above.
+#'   * `Repository`: for CRAN and CRAN-like repositories, the URL to the
```
aronatkins, line 76: The fact that _THIS_ `Repository` field is a URL, while the package description `Repository` is a _name_ has certainly caused confusion!

R/bundlePackage.R
```diff
@@ -103,73 +107,87 @@ findPackageRepoAndSource <- function(records, repos) {
...
+    } else {
+      repository <- findRepoUrl(pkg, availablePackages)
+      source <- findRepoName(repository, repos)
     }
-    # validatePackageSource will emit a warning for packages with NA repository.
-    data.frame(Source = source, Repository = repository, stringsAsFactors = FALSE)
-  })
-  do.call("rbind", tmp)
+  } else if (source %in% c("CRAN", "Bioconductor")) {
+    # Not strictly necessary since server will typically ignore in favour
```
aronatkins, line 154: What is this comment trying to convey?

For shinyapps.io / posit.cloud, "CRAN" and "Bioconductor" are well-known sources and use server-configured URLs. For Connect, this is not the case - the URL in the manifest is used during restore.

aronatkins (approved):

R/bundlePackage.R
hadley: Oooh, I had assumed that https://docs.posit.co/connect/admin/appendix/configuration/#RPackageRepository would be set by default

aronatkins: No; by default, the incoming repositories are used without modification. An admin may choose to override that choice (directing CRAN URLs to an internal RSPM mirror, for example).

PR #786, opened 2023-03-20
Use `forceUpdate = TRUE` when called from IDE

### LINKED ISSUE #785: IDE fails to update existing app
(if deployment records lost)

Screenshot 2023-03-20 at 13 19 00

With CRAN rsconnect I get an error:

```
Preparing to deploy document...DONE
Uploading bundle for document: 3100...DONE
Error in if (is.null(application$owner_username) || accountDetails$username ==  : 
  missing value where TRUE/FALSE needed
Calls: <Anonymous>
Execution halted
```

And with dev rsconnect I get a different error:

```
── Preparing for deployment ────────────────────────────────────────────────────
Error in `shouldUpdateApp()` at rsconnect/R/deploymentTarget.R:33:6:
! Discovered a previously deployed app named "test"
(View it at
<https://connect.rstudioservices.com/content/90764696-1620-41bb-a195-5dcc2cb02975/>)
ℹ Set `forceUpdate = TRUE` to update it.
ℹ Supply a unique `appName` to deploy a new application.
```

The second error suggests that maybe the IDE should be setting `forceUpdate = TRUE`?

hadley:
Fixes #785

@aronatkins I wonder if we should use this explicit check for being called from the IDE in other places.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -94,7 +94,9 @@
 #'   will always update the previously-deployed app. If `FALSE`, will ask
 #'   the user what to do, or fail if not in an interactive context.
 #'
-#'   Defaults to the value of `getOption("rsconnect.force.update.apps", FALSE)`.
+#'   Defaults to `TRUE` when called automatically by the IDE, and `FALSE`
+#'   otherwise. You can override the default by setting option
+#'   `rsconnect.force.update.apps`.
 #' @param python Full path to a python binary for use by `reticulate`.
 #'   Required if `reticulate` is a dependency of the app being deployed.
 #'   If python = NULL, and RETICULATE_PYTHON or RETICULATE_PYTHON_FALLBACK is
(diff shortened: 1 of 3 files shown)


aronatkins (approved):
Is there an argument against changing the default behavior to `TRUE` (update an item owned by this user having a matching name)?

Is this because the deploy pane cannot prompt for input?

hadley:
Yeah, think having a matching `appName` is not quite enough to know that you've got the right app; I worry particularly about people who have a bunch of directories called `shinyApp` where they put shiny apps.

The IDE does actually prompt in this case (see the dialog in #785). The cancel button will abort the deploy, but the replace button just continues the deploy, expecting that `deployApp()` will overwrite the existing app. I think the IDE could always set `forceUpdate = TRUE`, but it's easier to fix here so it works with all versions of the IDE.

PR #789, opened 2023-03-21
remove deployTFModel; Connect no longer supports hosting TensorFlow Model APIs

aronatkins:


R/deployTFModel.R
```diff
@@ -1,45 +0,0 @@
...
-#'   *saved_model.pb* or *saved_model.pbtxt*
-#' @param ... Additional arguments to [deployApp()].
-#'
-#' @references <https://www.tensorflow.org/guide/saved_model>
-#'
-#'
-#' @details
-#'
-#' @family Deployment functions
-#' @export
```
hadley, line 39: Might be worth keeping this function around with an informative error? Not that important but it is nice to know that it’s definitely gone away, rather than you forgetting the name of the function.

hadley (approved):

R/deployTFModel.R
aronatkins: @kellobri - What are your thoughts about this? My feeling is that this content type saw so little use that there is almost no risk to fully removing the TF deploy function. I'm happy to retain a skeleton implementation that errs if it'll be helpful.

R/appDependencies.R
```diff
@@ -110,7 +110,7 @@ appDependencies <- function(appDir = getwd(), appFiles = NULL) {
 }
 
 needsR <- function(appMetadata) {
-  if (appMetadata$appMode %in% c("static", "tensorflow-saved-model")) {
+  if (appMetadata$appMode %in% c("static")) {
```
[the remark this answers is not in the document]
aronatkins: I considered this, but the `c` reminds us that there might be more than one thing that is not R. I don't feel strongly.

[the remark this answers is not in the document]
toph-allen: I'm guessing that this function is not called for things other than `static` which don't need R? Perhaps it makes more sense to list the app does that *do* need R in that case, as it's more self-explanatory than requiring on this conditional + other conditionals elsewhere.

(I fell behind on @hadley's PRs when I had that two-week escalations stretch, so I recognize there may be context I'm missing here. I'll catch up on them when I'm done with server migration work.)

toph-allen (approved):

R/deployTFModel.R
kellobri: Yes, we saw extremely low usage. Errors are always nice though, I like the idea of the skeleton implementation if that's possible.

aronatkins:
Closing this PR; will push a new branch that maintains a defunct stub.

PR #790, opened 2023-03-21
initial implementation for static outputs in cloud

mslynch:
Adding content id makes things a bit weird, but I think it should work.

Keeping this as a draft until I can verify this works end-to-end on staging.

We shouldn't need to fork rsconnect – devtools supports branch package installs from github. This PR is directed at a feature branch that will merge to main once we get closer to go-live.


R/ide.R
```diff
@@ -114,7 +114,7 @@ showRstudioSourceMarkers <- function(basePath, lint) {
 
 # https://github.com/rstudio/rstudio/blob/ee56d49b0fca5f3d7c3f5214a4010355d1bb0212/src/gwt/src/org/rstudio/studio/client/rsconnect/ui/RSConnectDeploy.java#L699
 
-getAppById <- function(id, account, server, hostUrl) {
+getAppById <- function(id, account, server, hostUrl, contentId = NULL) {
```
mslynch, line 117: The IDE will need to pass contentId here.

PR #791, opened 2023-03-21
Respect renv lockfiles

### LINKED ISSUE #671: Allow separate "sources of truth" for R packages collected in deployment manifest
Currently, rsconnect relies on Packrat for lockfile generation, which implies that package dependencies are always inferred based on what's currently installed or available in the user's R library.

However, users might reasonably want to be able to declare their project's package requirements in other ways. For example, they may want to:

1. Override some package dependencies in a piecemeal way; e.g. use a different version on deployment, or install from a separate source,
2. Read package dependencies from a separate "source of truth" altogether; e.g. `renv.lock`; [pkg.lock](https://pak.r-lib.org/reference/lockfile_install.html); something else.

I believe part of (1) is supported on the Connect side with some side-channels for providing packages during a deployment, but (2) could be useful for users who want to be able to declaratively state what packages are to be used during deployment.

For an implementation sketch, I think we could have a function like `askManifestType()` that (if appropriate) would prompt the user for the "source of truth" for R packages, and then that could be passed as a parameter to `deployApp()`. There could also be a mechanism for setting these via R options / environment variables if it felt appropriate, or was asked for.

This request was from a professional customer, who would like more granular control over what packages are used and installed in deployments.

### LINKED ISSUE #809: `writeManifest` fails when encountering archived packages
Using [rowr](https://packagemanager.posit.co/client/#/repos/2/packages/rowr) as an example:

* Install `rowr` using `devtools::install_version('rowr', '1.1.3')` as directed by Posit Package Manager
* Run `rsconnect::writeManifest()` to create the `manifest.json`
* [Using CRAN version of `rsconnect`]: `manifest.json` is written with a warning, but the "Repository" field is `null`, which Connect will fail to deploy without an accompanying packrat.lock (for example, using Git deployment)
* [Using devel version of `rsconnect`]: fails immediately with the following error:
```
rsconnect::writeManifest()
Error in `createAppManifest()`:
! All packages must be installed from a reproducible location.
✖ Can't re-install packages installed from source: rowr.
ℹ See `rsconnect::appDependencies()` for more details.
```

This looks related to #508 with the note that while the current CRAN version deploys cleanly from push-button deployment to Connect, it will fail on a Git deployment (which only uses the `manifest.json`).

hadley:
@aronatkins and @kevinushey: this is still an early draft, but I thought it would be useful to get some feedback now that I've put a couple of hours into it. It still needs a bunch of testing, and I doubt i have all the details of the renv -> packrat translation are fully correct yet, but this overall this feels like a solid approach to me. What do you think?

Fixes #671. Fixes #809.


R/bundleFiles.R
```diff
@@ -178,8 +178,6 @@ ignoreBundleFiles <- function(dir, contents) {
   ignore <- c(
     # rsconnect packages
     "rsconnect", "rsconnect-python", "manifest.json",
-    # packrat + renv,
-    "renv", "renv.lock", "packrat",
```
hadley, line 182: Not sure if we should still ignore some of these? Or some parts of the renv/packrat directories?

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # rsconnect (development version)
 
+* `deployApp()` and friends will now respect packrat and renv lock files,
```
kevinushey, line 3: Do we report to the user during deploy how we're determining the bundle dependencies? (either by default, or only when using verbose logging?)

R/renv.R
```diff
@@ -0,0 +1,86 @@
...
+
+  # Convert renv sources to packrat sources
+  # https://github.com/rstudio/renv/blob/0.17.2/R/snapshot.R
+  if (identical(pkg$Repository, "CRAN")) {
+    pkg$Source <- "CRAN"
+  } else if (pkg$Source == "unknown") {
+    pkg$Source <- "source"
+  }
+
+  # Remove Remote fields that renv adds for "standard" installs from CRAN
```
kevinushey, line 52: This is normally from `pak` I think

R/renv.R
```diff
@@ -0,0 +1,86 @@
...
+  if (identical(pkg$RemoteType, "standard")) {
+    pkg <- pkg[!grepl("^Remote", names(pkg))]
+  }
+
+  # Drop hash since packrat hash != renv hash
+  pkg$Hash <- NULL
```
kevinushey, line 58: One worry I have is that, if I understand correctly, rsconnect tries to recompute the hash from the package DESCRIPTION files, but in this scenario we can't assume those DESCRIPTION files are available (or match whatever is recorded in the renv lockfile).

This might warrant some further discussion.

R/renv.R
```diff
@@ -0,0 +1,86 @@
...
+  renv <- jsonlite::read_json(renvLockFile(path))
+  packrat <- packratFromRenv(renv)
+  showDcf(packrat)
+}
+
+showDcf <- function(df) {
+  path <- tempfile()
+  on.exit(path, add = TRUE)
+
+  write.dcf(df, path)
```
kevinushey, line 83: I think `write.dcf()` supports connections, so you should be able to just `write.dcf(df, file = stdout())`?

R/renv.R
```diff
@@ -0,0 +1,86 @@
...
+}
+
+packratMeta <- function(renv) {
+  repos <- vapply(
+    renv$R$Repositories,
+    function(x) paste0(x$Name, "=", x$URL),
+    character(1)
+  )
+  data.frame(
+    PackratFormat = NA,
```
kevinushey, line 23: I'm guessing it's okay for these to be NA but you might want to set them to some "compatible" version just in case.

R/renv.R
aronatkins: Hashes are recalculated by Connect + Packrat at restore-time using the `DESCRIPTION` files included in the `packrat/desc` directory.

When we have only a `manifest.json`, the entire `packrat` directory is reconstructed using the package descriptions contained in the manifest.

The most straightforward approach is for `rsconnect` to stop including the `packrat` directory in the bundle and have Connect rely only on the `manifest.json`. The manifest structure is unchanged (including the full package description). That makes `rsconnect` driven deploys more similar to Git-backed deploys and `rsconnect-python` driven deploys.

NEWS.md
```diff
@@ -1,5 +1,10 @@
...
+  if present. If you don't want to use these lockfiles, and instead return
```
aronatkins, line 4: Any reason we want to support both packrat and renv? Are there reasons for folks to not migrate?

R/renv.R
```diff
@@ -0,0 +1,86 @@
+translateRenvToPackrat <- function(bundleDir) {
```
aronatkins, line 1: What about materializing the manifest directly from the renv.lock?

aronatkins (commented):
Could we renv->manifest rather than renv->packrat->manifest?

R/renv.R
kevinushey: What if we don't have any DESCRIPTION files available? For example, suppose that the package records declared in `renv.lock` aren't installed at all. More broadly, we might not be able to assume that `renv.lock` and the user's R library are actually in sync.

In theory, we could obtain those DESCRIPTION files based on the package records provided (I feel like we had some discussions about a "meta-CRAN" for something like this).

R/bundleFiles.R
hadley: If we ignore them, they won't be included in the `bundleDir()` and `snapshotRDependencies()` won't see them.

NEWS.md
hadley: It's trivial (i.e. a no-op) to support packrat, so why not?

R/renv.R
hadley: I thought we still needed the `packrat.lock` file because that's what connect uses?

R/renv.R
hadley: I think we assume that you can only deploy content that you can run locally; i.e. all the `DESCRIPTION` files are available. But it does seem wrong to get these `DESCRIPTION` files from the current library, not the library that renv used.

hadley: But we can at least take the `DESCRIPTION`s from renv library, not the system library.

R/bundleFiles.R
hadley: Based on further discussions we'll now ignore the `packrat/` and `renv/` directories automatically, and remove the `renv.lock` once we're done with it.

NEWS.md
hadley: I'll add. I think it makes sense to emit when `quiet = FALSE` (the default), since it's important for the average user.

R/renv.R
hadley: Good catch, thanks!

hadley:
Remaining to do:

* [ ] Figure out how to pass implicit/explicit `type` from `deployApp()` and `writeManifest()`
* [ ] Write script for `renv.lockfile` deployment integration tests with varying package types
* [ ] Perform integration tests

@aronatkins @kevinushey there's still a bit more work to be done, but I think the big picture is there, so if you have time it'd be worth casting your eyes over once more.

R/bundlePackagePackrat.R
```diff
@@ -0,0 +1,200 @@
...
+  )
+
+  # And strip trailing /
+  repos <- gsub("/$", "", repos)
+
+  repos
+}
+
+standardizePackratPackage <- function(record, availablePackages, repos = character()) {
+  pkg <- record$Package
```
hadley, line 62: Might also want to consider `RemoteRepos` for packages installed by pak.

hadley:
I'm still fighting some buglet with BioC packages. In a project that uses `Biobase`, I get the following error:

```
# Installing required R packages with `packrat::restore()` ---------------------
Installing BiocGenerics (0.44.0) ... 
Using cached BiocGenerics.
	OK (symlinked cache)
Installing BiocManager (1.30.18) ... 
Using cached BiocManager.
	OK (symlinked cache)
Installing BiocVersion (3.16.0) ... 
curl: HTTP 200 https://bioconductor.org/packages/3.16/data/annotation/src/contrib/PACKAGES.rds
curl: HTTP 200 https://bioconductor.org/packages/3.16/data/experiment/src/contrib/PACKAGES.rds
curl: HTTP 200 https://bioconductor.org/packages/3.16/bioc/src/contrib/PACKAGES.rds
curl: HTTP 200 https://bioconductor.org/packages/3.16/workflows/src/contrib/PACKAGES.rds
curl: HTTP 200 https://colorado.posit.co/rspm/all/latest/src/contrib/PACKAGES.rds
curl: HTTP 404 https://bioconductor.org/packages/3.16/data/annotation/src/contrib/Archive/BiocVersion/BiocVersion_3.16.0.tar.gz
curl: (22) The requested URL returned error: 404 Not Found
curl: HTTP 404 https://bioconductor.org/packages/3.16/data/annotation/src/contrib/Archive/BiocVersion/BiocVersion_3.16.0.tar.gz
```

The manifest looks ok to me:

```
    "BiocVersion": {
      "Package": "BiocVersion",
      "Version": "3.16.0",
      "Source": "Bioconductor",
      "Repository": "https://bioconductor.org/packages/3.16/bioc",
```

I suspect there's some small mistake that will hopefully be obvious when I look at it on Monday.

R/bundlePackageRenv.R
```diff
@@ -0,0 +1,139 @@
...
+
+  pkg[manifestPackageColumns(pkg)]
+}
+
+biocPackages <- function(bundleDir) {
+  signal("evaluating", class = "rsconnect_biocRepos") # used for testing
+  availablePackages(biocRepos(bundleDir))
+}
+biocRepos <- function(bundleDir) {
+  repos <- getFromNamespace("renv_bioconductor_repos", "renv")(bundleDir)
```
hadley, line 128: @kevinushey do we also need to consider `options(BioC_mirror)` or does `renv_bioconductor_repos()` already do that?

R/writeManifest.R
jthomasmock, line 1: A side effect of renv unless I'm missing something is that the PROJECT libraries affect all subdirectories. Per the docs, isolating manifest files to the specific subdirectories is the intended workflow to avoid pulling in the entire directory of files: https://docs.posit.co/connect/user/git-backed/#creating-a-manifest-file-from-r

So in a more complex project, with a subdirectory for each app, every bundle will contain all dependencies, even if the specific bundle has no need for them. I'm imagining a project with a static report for scheduling and a shiny app. The static report doesn't need all the shiny app packages but the manifest for both bundles would have everything in PROJECT renv lock file?

In that case, maybe the user has to either accept all dependencies or opt out of using renv lockfile.

R/bundlePackageRenv.R
kevinushey: `renv` basically just calls `BiocManager:::.repositories()` so that that should respect BioC_mirror IIUC.

hadley: I'll probably switch this to use the `git_branch` `DESCRIPTION` (with some error/fallback) as it sounds like it's ok to rely on that: https://github.com/Bioconductor/BiocManager/issues/164#issuecomment-1519088086

R/writeManifest.R
edavidaja: I think perhaps it would make more sense to encourage users to  adopt a "one lockfile per deployable thing" in a project rather than having a "project" level lockfile?

hadley: To be clear, this change will only recognise lockfiles in the deployed directory, not in the parent.

hadley:
BioC install problem was caused by colorado only having 4.1 installed. Works on rstudioservices.

R/bundlePackagePackrat.R
```diff
@@ -0,0 +1,200 @@
...
+    # Package was installed from source, but packrat guessed it was installed
+    # from a known repo. We need to check that it's not actually a dev version
+    if (isDevVersion(record, availablePackages)) {
+      repository <- NA_character_
+      source <- NA_character_
+    } else {
+      repository <- findRepoUrl(pkg, availablePackages)
+      source <- findRepoName(repository, repos)
+    }
+  } else if (source %in% c("CRAN", "Bioconductor")) {
```
toph-allen, line 84:
```suggestion
  } else if (source == "CustomCRANLikeRepository" && isDevVersion(record, availablePackages)) {
    # Package was installed from source, but packrat guessed it was installed
    # from a known repo.
      repository <- NA_character_
      source <- NA_character_
  } else if (source %in% c("CRAN", "Bioconductor")) {
```

To me, this makes the overall conditional flow of this block easier to follow -- the nested `else` is the same as the final `else` in the outer set of conditionals. Feel free to take or leave, this is definitely a nitpick and a matter of taste.

toph-allen (changes requested):
In general, I think this change makes sense, especially because `packrat` is itself using a vendored version of `renv` to snapshot dependencies.

It seems like a lot of the complexity here is within the blocks that standardize the different ways `packrat`, `renv`, and `pak` record package metadata, as well as assumptions they make about packages that aren't just structural differences (and how these have evolved over time). I appreciate having links to the relevant source in the comments. (The one that stood out to me was the check for `isDevVersion()`).

The Bitbucket name capitalization is the only required change I noticed. But I only noticed it because I spent long enough with Bitbucket when I was implementing support for restoring those packages in Connect. We'll probably want to do some real-world tire-kicking of those sorts of scenarios in case there are other small things like that we missed. Either that or build out more examples in tests.

R/bundlePackageRenv.R
```diff
@@ -0,0 +1,142 @@
...
+        pkg$Source <- NA_character_
+        pkg$Repository <- NA_character_
+      } else {
+        pkg$Source <- "CRAN"
+        pkg$Repository <- findRepoUrl(pkg$Package, availablePackages)
+      }
+    } else {
+      # $Repository comes from DESCRIPTION and is set by repo, so can be
+      # anything. So we must look up from the package name
+      pkg$Repository <- findRepoUrl(pkg$Package, availablePackages)
```
aronatkins, line 103: nit: during restore, renv first favors a matching repository name before falling back to "any" repository. not sure if we want that same pattern here or want to rely `available.packages` order like we do for Packrat.

Maybe this is: prefer the Repository named in renv.lock and prefer available.packages for snapshots.

Also worth considering: Should we prefer the renv.lock repository list or the list in the current R session?

https://github.com/rstudio/renv/blob/81fecb07fcc78cabb61b5309875e59484727b090/R/retrieve.R#L875-L880

aronatkins (approved):
Some questions about package+repository handling, but this is probably far enough along to merge and get folks to start actively using it.

DESCRIPTION
```diff
@@ -28,13 +28,15 @@ Imports:
     lifecycle,
     openssl (>= 2.0.0),
     packrat (>= 0.6),
+    renv,
```
kevinushey, line 31: Not sure if you need it for this PR, but if you'd like to depend on `renv` internals without risking breakage from new `renv` releases, you can also use `renv::embed()` to embed a local copy of `renv` here.

R/appDependencies.R
```diff
@@ -8,6 +8,27 @@
...
+#' uses:
+#'
+#' 1. If `renv.lock` is present, it will use the versions and sources defined in
+#'    that file. If you're using the lockfile for some other purpose and
+#'    don't want it to affect deployment, add `renv.lock` to `.rscignore`.
+#'
+#' 2. Otherwise, rsconnect will call `renv::snapshot()` to find all packages
+#'    used by your code. If you'd instead prefer to only use the packages
+#'    declared in a `DESCRIPTION` file, run
+#'    `renv::settings$snapshot.type("explicit")` to activate renv's "explicit"
```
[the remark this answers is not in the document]
kevinushey: `renv::snapshot()` does accept a `type` argument so this would be straightforward to wire up, at least.

R/bundlePackagePackrat.R
```diff
@@ -0,0 +1,200 @@
...
+  list(Source = source, Repository = repository)
+}
+
+findRepoName <- function(repository, repos) {
+  idx <- match(repository, repos)
+  names(repos)[idx]
+}
+
+findRepoUrl <- function(pkg, availablePackages) {
+  if (pkg %in% rownames(availablePackages)) {
```
kevinushey, line 103: Nit: might be slightly safer to check `[, "Package"]` in case there were duplicate row names that got unique-ified?

kevinushey: (or can we already guarantee that this wouldn't occur? I'm thinking of things like where `MASS` and recommended packages are available in two parts of the repository, e.g. in `Recommend/4.3.0` or something)

R/bundlePackagePackrat.R
```diff
@@ -0,0 +1,200 @@
...
+
+isDevVersion <- function(record, availablePackages) {
+  if (!record$Package %in% rownames(availablePackages)) {
+    return(FALSE)
+  }
+
+  local_version <- record$Version
+  repo_version <- availablePackages[record$Package, "Version"]
+
+  package_version(local_version) > package_version(repo_version)
```
kevinushey, line 121: Should we also detect whether the trailing version component is "large"? In case a local version is 1.0.0-9000 but CRAN has 1.0.1?

R/bundlePackageRenv.R
```diff
@@ -0,0 +1,142 @@
...
+  defer(options(old))
+
+  renv::snapshot(bundleDir, prompt = FALSE)
+  defer(removeRenv(bundleDir))
+
+  parseRenvDependencies(bundleDir, snapshot = TRUE)
+}
+
+parseRenvDependencies <- function(bundleDir, snapshot = FALSE) {
+  renv <- jsonlite::read_json(renvLockFile(bundleDir))
```
kevinushey, line 16: I wonder if I should expose something like `renv::lockfile$read()` so you can read an `renv` lockfile and let `renv` handle parsing of the Repositories field

R/bundlePackageRenv.R
```diff
@@ -0,0 +1,142 @@
...
+  )
+  if (nrow(deps) == 0) {
+    return(data.frame())
+  }
+
+  if (snapshot) {
+    # Can use system libraries
+    deps$description <- lapply(deps$Package, function(pkg) {
+      dcf <- read.dcf(system.file("DESCRIPTION", package = pkg))
+      as.list(as.data.frame(dcf))
```
kevinushey, line 35:
```suggestion
      as.list(as.data.frame(dcf, stringsAsFactors = FALSE))
```

R/bundlePackageRenv.R
```diff
@@ -0,0 +1,142 @@
...
+    defer(options(old))
+
+    # Generate a library from the lockfile
+    lib_dir <- dirCreate(file.path(bundleDir, "renv_library"))
+    renv::restore(bundleDir, library = lib_dir, prompt = FALSE)
+    defer(unlink(lib_dir, recursive = TRUE))
+
+    deps$description <- lapply(deps$Package, function(pkg) {
+      dcf <- read.dcf(file.path(lib_dir, pkg, "DESCRIPTION"))
+      as.list(as.data.frame(dcf))
```
kevinushey, line 48:
```suggestion
      as.list(as.data.frame(dcf, stringsAsFactors = FALSE))
```

R/utils.R
```diff
@@ -132,6 +132,19 @@ hasPrefix <- function(x, prefix) {
   substring(x, 1, nchar(prefix)) == prefix
 }
 
+# Lightweight equivalent of withr::defer()
+defer <- function(expr, env = caller_env(), after = FALSE) {
+  thunk <- as.call(list(function() expr))
+  do.call(on.exit, list(thunk, TRUE, after), envir = env)
```
kevinushey, line 138: Nit: The `after` argument requires R (>= 3.5.0); can we take a dependency on newer R here?

kevinushey (approved):
Overall LGTM!

DESCRIPTION
aronatkins: Related thought: Do we need a version constraint on renv?

hadley: We don't need to embed because we're only using exported functions so far, but will keep it in mind for the future. 

I've add a requirement for the latest version from CRAN.

R/appDependencies.R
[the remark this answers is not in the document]
hadley: My sense is that once we enable support for renv lockfile, we'll get quite a few feature requests so we might as well just let them accumulate for a while to get a sense for what people actually want, rather than doing too much prospectively.

R/appDependencies.R
```diff
@@ -8,6 +8,27 @@
...
+#'    mode.
+#'
+#' 3. Dependency resolution using renv is a new feature in rsconnect 1.0.0, and
```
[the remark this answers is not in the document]
hadley: I have; deprecation messages updated in #830.

R/bundleFiles.R
```diff
@@ -179,7 +179,7 @@ ignoreBundleFiles <- function(dir, contents) {
     # rsconnect packages
     "rsconnect", "rsconnect-python", "manifest.json",
     # packrat + renv,
-    "renv", "renv.lock", "packrat",
+    "renv", "packrat",
```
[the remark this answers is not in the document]
hadley: Yes, but I don't think that should matter because the settings only really affect interactive usage, and the consequence of those settings will be encoded in the lockfile.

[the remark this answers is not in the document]
hadley: Yes; that's necessary so that we can process it in `bundlePackages()`. Do you foresee a problem with including this in the deployed bundle? If so, I can delete. (But I think you could also `renv::restore()` directly from that file without having to convert the manifest back into a packrat spec).

R/bundlePackagePackrat.R
hadley: I like it, thanks!

R/bundlePackagePackrat.R
hadley: I hadn't thought about that problem, thanks!

R/bundlePackagePackrat.R
hadley: That's not a version used by every package, so I worry that would dilute the impact of this change.

R/bundlePackageRenv.R
hadley: https://github.com/rstudio/renv/issues/1220

R/utils.R
hadley: We already have an implicit dependency on 3.5 via rlang, but I've made the explicit in the `DESCRIPTION`.

R/bundleFiles.R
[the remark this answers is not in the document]
aronatkins: The recommendation to use `renv::settings$snapshot.type("explicit")` presents a challenge, then, if we expect folks to set that ahead of each deploy operation but do so in a way that does not modify their existing renv project settings. What is your suggested workflow?

hadley:
I think this is done, and I've finally managed to knock of the CI failure. Unless anyone spots any problems, I'll aim to merge tomorrow.

R/bundleFiles.R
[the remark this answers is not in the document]
aronatkins: I am torn on this one.

On the one hand, including only the `manifest.json` in the bundle is quite nice because it means we have a single source-of-truth. Folks have been confused by the presence of both `manifest.json` and `packrat.lock` in our bundles, as well as the relationship between the two (and duplication of information).

On the other hand, I like including the `renv.lock`, as it gives us a future option of directly using that file without reconstructing it from the manifest.

Mostly, my struggle is that I had thought about this change as "manifest-only, no packrat.lock" rather than "manifest plus renv.lock; except when we create the renv.lock on the fly".

Let's go forward with the (sometimes) inclusion, but probably document the change.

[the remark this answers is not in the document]
aronatkins: A generated lock file isn't on the set of "bundle files" that was computed much earlier in the workflow... I'm OK either way, but if we do include a generated lock file, let's include it only when we have tracked dependencies (e.g. not for static).

toph-allen (approved):
Looked through the changes since my last review, and they all look good.

PR #794, opened 2023-03-22
Provide better tools for working with environment variables

### LINKED ISSUE #667: Better tools for environment variables

hadley:
Fixes #667


R/deployApp.R
```diff
@@ -365,6 +377,11 @@ deployApp <- function(appDir = getwd(),
     )
     taskComplete(quiet, "Visibility updated")
   }
+  if (length(target$envVars) > 0) {
+    taskStart(quiet, "Updating environment variables {envVars}...")
+    client$setEnvVars(application$guid, target$envVars)
```
aronatkins, line 382: Document that environment variables are applied _BEFORE_ the deployment. Should the deploy fail, the env var updates will still be in effect.

R/deployApp.R
```diff
@@ -151,6 +158,7 @@ deployApp <- function(appDir = getwd(),
                       appSourceDoc = NULL,
                       appName = NULL,
                       appTitle = NULL,
+                      envVars = NULL,
```
aronatkins, line 161: Any reason why `envVars` is not the final argument? Placing it here breaks any calls which assume argument order.

hadley: In a function like this, I think it's reasonable to assume that folks will name arguments rather than relying on position (especially at position 7), and this allows us to make the importance of this variable more obvious by listing it earlier in the function signature.

If you're worried about this silently changing existing code we code at `...` to the function signature (forcing all argument after the `...` to be named) and use `rlang::check_dots_empty()` to give an informative error.

R/deployApp.R
```diff
@@ -48,6 +48,13 @@
 #' @param appTitle Free-form descriptive title of application. Optional; if
 #'   supplied, will often be displayed in favor of the name. If ommitted,
 #'   on second and subsequent deploys, the title will be unchanged.
+#' @param envVars A character vector giving the names of environment variables
+#'   to upload to the server. The values of the environment variables are sent
+#'   over an encrypted connection and are not stored in the bundle, making
+#'   this a safe way to send private data to Connect.
```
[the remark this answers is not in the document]
hadley: Ok, I've taken a stab at this.

R/envvars.R
```diff
@@ -0,0 +1,75 @@
...
+#' @description
+#' * `listAccountEnvVars()` lists the environment variables used by
+#'   every application published to the specified account.
+#' * `updateAccountEnvVars()` updates the specified environment variables with
+#'   their current values for every app that uses them.
+#'
+#' Secure environment variable are currently only supported by Posit Connect
+#' so other server types will generate an error.
+#'
+#' @inheritParams deployApp
```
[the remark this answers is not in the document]
hadley: See line 9 😄

R/deployApp.R
```diff
@@ -48,6 +48,22 @@
 #' @param appTitle Free-form descriptive title of application. Optional; if
 #'   supplied, will often be displayed in favor of the name. If ommitted,
 #'   on second and subsequent deploys, the title will be unchanged.
+#' @param envVars A character vector giving the names of environment variables
+#'   whose values should be synchronised with the server (currently supported by
+#'   Connect only). The values of the environment variables are sent over an
+#'   encrypted connection and are not stored in the bundle, making this a safe
+#'   way to send private data to Connect.
```
aronatkins, line 55: We should eventually create example workflows:

1. named environment variables that are assigned on each deploy.
2. named environment variables that are assigned only on first deploy.
3. modifying the set of environment variables over time (across multiple deploys).

aronatkins (approved):

PR #796, opened 2023-03-24
Remove no-longer-needed readDcf

hadley:
Fix in base R in https://github.com/wch/r-source/commit/ef5d697064e10e33070b4024f06fdd7628c80bd4


aronatkins (approved):
Probably no NEWS need, yes? Especially since it was fixed in a so-long-ago version of R (probably 3.1.x or 3.2.x)?

PR #800, opened 2023-03-24
Display existing app url during deployment

hadley:


R/deployApp.R
```diff
@@ -342,7 +342,7 @@ deployApp <- function(appDir = getwd(),
       }
     )
     if (application$id == target$appId) {
-      taskComplete(quiet, "Found application")
+      taskComplete(quiet, "Found application {.url {application$url}}")
```
aronatkins, line 345: Should we also provide the URL when creating content? It's probably too early to present a URL to where the content will be served, but maybe a dashboard URL.

Talking out loud. No action needed.

aronatkins (approved):

PR #822, opened 2023-04-27
lint R code files for reticulate environment setup code

edavidaja:


hadley:
I want to move away from the linting system, because IMO static code analysis is fundamentally fragile. I think it would be better to think about how we could detect this dynamically, once we have the dry run system working.

kevinushey:
> I want to move away from the linting system, because IMO static code analysis is fundamentally fragile. I think it would be better to think about how we could detect this dynamically, once we have the dry run system working.

Wouldn't static analysis be necessary to diagnose things like absolute paths / case-sensitive path issues?

hadley:
@kevinushey yeah, I think that has to stay as static analysis.

Issue #824, opened 2023-04-28
Add support for deploying client-rendered static outputs to Posit Cloud


mslynch:
This is an upcoming feature in Cloud that rsconnect should support.

Issue #821, opened 2023-04-26
Version the deployment dcf record


hadley:
Should do it at as a matter of principle, and it will be useful for the upcoming app ID -> content ID change for cloud.

Issue #819, opened 2023-04-20
Support appId when publishing to Posit Cloud


mbaynton:
`deployApp()` includes the parameter `appId` whose intended usage is to allow publishing to a specific existing output on the server. This functionality needs to be given some love when the destination is Posit Cloud:

 * Posit Cloud has a notion of an "application", but its ID is intentionally difficult for a user to track down because we prefer instead to
    expose related construct "content" to users. Work happening right now to support a history of prior revisions to some types of published content in cloud also makes the relationship between a Cloud "content" and  a Cloud "application" not one-to-one -- each time you publish a new revision, you use the same content ID but get a new application ID. For all these reasons, to allow the user to publish to an existing output in Cloud without a .dcf deployment history, they should be providing the content ID, not the application ID.
  * As a result of the above, in the case where a deployment history is available, we're already written rsconnect code that switches to persisting the content ID rather than the application ID. I suggest this work merges before tackling this issue, and then we fix remaining problem(s) with manually passing a Cloud content ID:
  1. Manually passing a content ID (e.g. `deployApp(appId='lucid:content:123', ...)`) fails with
      ```
      Error in (function (..., row.names = NULL, check.rows = FALSE, check.names = TRUE,  : 
      arguments imply differing number of rows: 1, 0
     ```
  2. Perhaps we should _always_ consider any value passed to `appId` to be a content ID if the server is `posit.cloud`, even if not made explicit with the `lucid:content:`?

kippandrew:
I'd like to propose a third option that is similar to open 2:

3. Consider deprecating the `appId` parameter in favor of a new `contentId` parameter. I think this would work well for Posit Connect and Posit Cloud, both of which have the concept of `contentId` though there are differences (see my comment here: https://github.com/rstudio/rsconnect/issues/808#issuecomment-1521680429). However, shinyapps.io doesn't really have the same concept, but maybe it would be acceptable to have `contentId` mean `appId` when used with shinyapps.io.

mslynch:
We decided on passing the unqualified content id in appId (`deployApp(appId='123'...)`) – this will work when we implement static publishing. We'll also ensure backwards compatibility with old dcf files using application id through https://github.com/rstudio/rsconnect/issues/821.

mslynch:
Reopening – the error `arguments imply differing number of rows: 1, 0` is a regression on main, unrelated to any content id/app id changes. cc @hadley

hadley:
I'll take a look this afternoon.

PR #825, opened 2023-04-28
Cloud static outputs

mslynch:
- Add support for deploying client-rendered static outputs to Posit Cloud
- Save Cloud deployments with appId as the Cloud content id (instead of application id)
- Introduce versioning to dcf files (needed to support backwards compatibility for old Cloud dcf files saved with appId as the application id

Resolves https://github.com/rstudio/rsconnect/issues/824.
Resolves https://github.com/rstudio/rsconnect/issues/821.

### DIFF
--- a/R/client-cloud.R
+++ b/R/client-cloud.R
@@ -66,15 +66,39 @@ cloudClient <- function(service, authInfo) {
       listRequest(service, authInfo, path, query, "applications")
     },
 
-    getApplication = function(applicationId) {
-      path <- paste("/applications/", applicationId, sep = "")
-      application <- GET(service, authInfo, path)
+    getApplication = function(applicationId, dcfVersion) {
+      if (!(is.na(dcfVersion) || is.null(dcfVersion))) {
+        # On version >=1, applicationId refers to the id of the output, not the application.
+        path <- paste("/outputs/", applicationId, sep = "")
+        output <- GET(service, authInfo, path)
 
-      output_id <- application$content_id
-      path <- paste("/content/", output_id, sep = "")
+        path <- paste("/applications/", output$source_id, sep = "")
+        application <- GET(service, authInfo, path)
+      } else {
+        # backwards compatibility for data saved with the application's id
+        # TODO: remove support for this case
+        path <- paste("/applications/", applicationId, sep = "")
+        application <- GET(service, authInfo, path)
 
-      applications_output <- GET(service, authInfo, path)
-      application$url <- applications_output$url
+        output_id <- ifelse(is.null(application$output_id), application$content_id, application$output_id)
+
+        path <- paste("/outputs/", output_id, sep = "")
+        output <- GET(service, authInfo, path)
+      }
+
+      # if the output is trashed or archived, restore it to the active state
+      if (output$state == 'trashed' || output$state == 'archived') {
+        json <- list()
+        json$state <- 'active'
+        PATCH_JSON(service, authInfo, paste("/outputs/", output$id, sep = ""), json)
+      }
+
+      # Each redeployment of a static output creates a new application. Since
+      # those applications can be deleted, it's more reliable to reference
+      # outputs by their own id instead of the applications'.
+      application$content_id <- output$id
+      application$url <- output$url
+      application$name <- output$name
       application
     },
 
@@ -97,13 +121,13 @@ cloudClient <- function(service, authInfo) {
       GET(service, authInfo, path, query)
     },
 
-    createApplication = function(name, title, template, accountId) {
+    createApplication = function(name, title, template, accountId, appMode) {
       json <- list()
       json$name <- name
+      json$application_type <- ifelse(appMode == "static", "static", "connect")
 
       currentApplicationId <- Sys.getenv("LUCID_APPLICATION_ID")
       if (currentApplicationId != "") {
-        print("Found application...")
         path <- paste("/applications/", currentApplicationId, sep = "")
         current_application <- GET(service, authInfo, path)
         project_id <- current_application$content_id
@@ -120,6 +144,7 @@ cloudClient <- function(service, authInfo) {
       output <- POST_JSON(service, authInfo, "/outputs", json)
       path <- paste("/applications/", output$source_id, sep = "")
       application <- GET(service, authInfo, path)
+      application$content_id <- output$id
       # this swaps the "application url" for the "content url". So we end up redirecting to the right spot after deployment.
       application$url <- output$url
       application
@@ -159,8 +184,14 @@ cloudClient <- function(service, authInfo) {
       )
     },
 
-    deployApplication = function(applicationId, bundleId = NULL) {
-      path <- paste("/applications/", applicationId, "/deploy", sep = "")
+    createRevision = function(application) {
+        path <- paste("/outputs/", application$content_id, "/revisions", sep = "")
+        revision <- POST_JSON(service, authInfo, path, data.frame())
+        revision$application_id
+    },
+
+    deployApplication = function(application, bundleId = NULL) {
+      path <- paste("/applications/", application$id, "/deploy", sep = "")
       json <- list()
       if (length(bundleId) > 0 && nzchar(bundleId))
         json$bundle <- as.numeric(bundleId)
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -345,29 +345,33 @@ deployApp <- function(appDir = getwd(),
     showCookies(serverInfo(accountDetails$server)$url)
   }
 
+  isNewApplication <- FALSE
   if (is.null(target$appId)) {
     taskStart(quiet, "Creating application on server...")
     application <- client$createApplication(
       target$appName,
       target$appTitle,
       "shiny",
-      accountDetails$accountId
+      accountDetails$accountId,
+      appMetadata$appMode
     )
     taskComplete(quiet, "Created application with id {.val {application$id}}")
+    isNewApplication <- TRUE
   } else {
     application <- taskStart(quiet, "Looking up application with id {.val {target$appId}}...")
     application <- tryCatch(
-      client$getApplication(target$appId),
+      {
+        application <- client$getApplication(target$appId, target$version)
+        taskComplete(quiet, "Found application {.url {application$url}}")
+        application
+      },
       rsconnect_http_404 = function(err) {
-        applicationDeleted(client, target, recordPath)
+        application <- applicationDeleted(client, target, recordPath, appMetadata)
+        taskComplete(quiet, "Created application with id {.val {application$id}}")
+        isNewApplication <- TRUE
+        application
       }
     )
-    if (application$id == target$appId) {
-      taskComplete(quiet, "Found application {.url {application$url}}")
-    } else {
-      taskComplete(quiet, "Created application with id {.val {application$id}}")
-    }
-
   }
   saveDeployment(
     recordPath,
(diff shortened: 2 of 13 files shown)


R/client-cloud.R
```diff
@@ -66,15 +66,39 @@ cloudClient <- function(service, authInfo) {
       listRequest(service, authInfo, path, query, "applications")
     },
 
-    getApplication = function(applicationId) {
-      path <- paste("/applications/", applicationId, sep = "")
-      application <- GET(service, authInfo, path)
+    getApplication = function(applicationId, dcfVersion) {
+      if (!(is.na(dcfVersion) || is.null(dcfVersion))) {
+        # On version >=1, applicationId refers to the id of the output, not the application.
+        path <- paste("/outputs/", applicationId, sep = "")
```
omar-rs, line 72: Shouldn't we be checking the dcf version here?  Basically, what the comment is saying should be turned into an `if` statement.

R/client-cloud.R
```diff
@@ -66,15 +66,39 @@ cloudClient <- function(service, authInfo) {
       listRequest(service, authInfo, path, query, "applications")
     },
 
-    getApplication = function(applicationId) {
-      path <- paste("/applications/", applicationId, sep = "")
-      application <- GET(service, authInfo, path)
+    getApplication = function(applicationId, dcfVersion) {
```
omar-rs, line 69: Would be clearer if we renamed the function arg `applicationId` to just `id`, since it could be either that of the application or the output?

R/client-cloud.R
mslynch: These if statements are equivalent, since there are no versions under 1.

R/deployApp.R
```diff
@@ -345,29 +345,33 @@ deployApp <- function(appDir = getwd(),
...
+      accountDetails$accountId,
+      appMetadata$appMode
     )
     taskComplete(quiet, "Created application with id {.val {application$id}}")
+    isNewApplication <- TRUE
   } else {
     application <- taskStart(quiet, "Looking up application with id {.val {target$appId}}...")
     application <- tryCatch(
-      client$getApplication(target$appId),
+      {
```
hadley, line 363: I like how this change puts the messaging closer to the corresponding action 😄

R/deployApp.R
```diff
@@ -434,7 +441,7 @@ deployApp <- function(appDir = getwd(),
   if (!quiet) {
     cli::cli_rule("Deploying to server")
   }
-  task <- client$deployApplication(application$id, bundle$id)
```
hadley, line 437: Can you explain this change? Had I broken the existing code?

R/deploymentTarget.R
```diff
@@ -76,13 +77,17 @@ deploymentTargetForApp <- function(appId,
   accountDetails <- findAccount(account, server)
   application <- getApplication(accountDetails$name, accountDetails$server, appId)
 
+  resultAppId <- ifelse(is.null(application$content_id), application$id, application$content_id)
+
   createDeploymentTarget(
```
hadley, line 82: I think it's now worth naming all these arguments too, since otherwise I can't easily tell what you've changed here.

hadley (commented):
Looking good! Do you also want to add a news bullet briefly describing these changes?

R/client-cloud.R
```diff
@@ -66,15 +66,39 @@ cloudClient <- function(service, authInfo) {
       listRequest(service, authInfo, path, query, "applications")
     },
 
-    getApplication = function(applicationId) {
-      path <- paste("/applications/", applicationId, sep = "")
-      application <- GET(service, authInfo, path)
+    getApplication = function(outputOrApplicationId, dcfVersion) {
+      if (is.na(dcfVersion) || is.null(dcfVersion)) {
```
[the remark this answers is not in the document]
hadley: Actually I think I have this the wrong way around — `dcfVersion` comes from subsetting `deployments()`  which returns a data frame, so can only be `NA`, not `NULL`.

R/deploymentTarget.R
hadley: Actually, I'll make that change so you can just bring in my code.

tests/testthat/test-deployments.R
```diff
@@ -111,9 +111,10 @@ test_that("saveDeployment appends to global history", {
       envVars = NULL,
       account = "foo",
       username = "foo",
-      server = "bar"
+      server = "bar",
+      version = dcfVersion
     ),
-    application = list(),
+    application = list(id = 1),
```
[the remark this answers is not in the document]
hadley: See #827.

R/deployments.R
```diff
@@ -143,7 +143,10 @@ deploymentRecord <- function(name,
   c(standard, metadata)
 }
 
+dcfVersion <- "1"
```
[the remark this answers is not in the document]
aronatkins: The meaning of one of the fields is changing, yes? What does this mean when we attempt to use those DCF files with an old `rsconnect` release?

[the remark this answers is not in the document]
hadley: Ooh, I think I get what you mean now — we're adding a `version` field that will allow us to do more versioning in the future, but there's no intent to add any backward incompatible changes at this time. The current goal is to make it possible to tell if `applicationId` means `appId` or `contentId` in existing deployments to cloud; `contentId` is substantially easier to work with and more meaningful on the server side, but we don't want to break existing deployment records that stored `appId`. 

We considered renaming `applicationId` more generally (since rsconnect is not just about applications any more), but given that it's mostly an internal feature used by the IDE, we didn't think it was worth it.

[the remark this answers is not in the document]
aronatkins: The situation that concerned me:

1. Deploy (new) using CRAN rsconnect (or any version prior to this change).
2. Re-deploy using this version of rsconnect (which upgrades the DCF file).
3. Re-deploy using CRAN rsconnect.

Is the deployment DCF value of `applicationId` constant or does it change during this re-deploy sequence?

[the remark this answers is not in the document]
mslynch: @aronatkins The value of `appId` in the DCF file would change when deploying using this version of rsconnect (from the application id to the content id), so that scenario would break at step 3.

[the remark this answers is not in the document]
aronatkins: @hadley - You had mentioned trying to avoid this type of backwards incompatibility with the upcoming release.. has your opinion changed?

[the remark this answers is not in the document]
aronatkins: The reasoning: There's enough changing in rsconnect that the upcoming release will cause issues for some folks despite our efforts to avoid that. Folks will probably downgrade to the previous CRAN release until we can patch. We wanted to avoid situations where the downgrade caused its own problems.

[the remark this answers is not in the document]
hadley: I _think_ this is ok because of its relatively limited scope — my other proposed change would've broken every single piece of metadata that you'd saved, which seemed a bit extreme.

R/client-cloud.R
```diff
@@ -66,20 +66,46 @@ cloudClient <- function(service, authInfo) {
       listRequest(service, authInfo, path, query, "applications")
     },
 
-    getApplication = function(applicationId) {
-      path <- paste("/applications/", applicationId, sep = "")
-      application <- GET(service, authInfo, path)
+    getApplication = function(outputOrApplicationId, dcfVersion) {
+      if (is.na(dcfVersion)) {
```
mbaynton, line 70: I'm passively curious why versioning ended up being favored over the kinda duck-typing strategy where `outputOrApplicationId` was a lucid resource identifier.

Is this PR intended to address https://github.com/rstudio/rsconnect/issues/819?

mbaynton: I tested my understanding of the acceptance criteria for #819 using this branch, but get various errors or misbehaviors (new output in Cloud) depending on whether I use `deployDoc` or `deployApp` and whether an `rsconnect` directory is present.

I'm not sure fixing #819 is in scope for this PR, but I'm unsure this PR really resolves it. Does it work for you? Should we reopen that one?

mslynch: Ah, I misunderstood the problem... I had interpreted it as being related to the in-progress work for this issue, but it's actually a regression on main.

R/client-cloud.R
```diff
@@ -66,20 +66,46 @@ cloudClient <- function(service, authInfo) {
       listRequest(service, authInfo, path, query, "applications")
     },
 
-    getApplication = function(applicationId) {
-      path <- paste("/applications/", applicationId, sep = "")
-      application <- GET(service, authInfo, path)
+    getApplication = function(outputOrApplicationId, deploymentRecordVersion) {
+      if (is.na(deploymentRecordVersion)) {
+        # In pre-versioned dcf files, contentOrAppId is the id of the application.
```
toph-allen, line 71: Supposed to be `outputOrApplicationId` in the comment?

toph-allen (commented):
From the ongoing conversations and `NEWS.md`, I expected to find DCF versioning in `deployments.R`, but it looks like it was removed when `main` was merged into this branch in https://github.com/rstudio/rsconnect/pull/825/commits/215fcbba4b042d41db6992dbe916a0fa62fa97b8. However, tests are still passing; I'm not sure if it's intentional or not.

toph-allen (commented):
I can't actually see where `deploymentRecordVersion` is set. It looks like there used to be a variable named `dcfVersion` in `deployments.R`, but that change was removed in one of the commits that merged `main` into this branch.

I might just be overlooking something or searching for the wrong string, or misunderstanding the current state of things looking at the conversation. :)

R/deploymentTarget.R
```diff
@@ -94,7 +94,7 @@ createDeploymentTarget <- function(appName,
                                    username,
                                    account,
                                    server,
-                                   version = 1) {
+                                   version = deploymentRecordVersion) {
```
mslynch, line 97: > I can't actually see where `deploymentRecordVersion` is set.

@toph-allen That's right here.

omar-rs (approved):

hadley (approved):
Looks good to me!   

The only last thing you might want to do is to note any manual testing you did in a new `.Rmd` in `tests/manual`, so it's easier to remember in the future.

R/deployApp.R
```diff
@@ -355,23 +355,29 @@ deployApp <- function(appDir = getwd(),
...
     taskComplete(quiet, "Created application with id {.val {application$id}}")
   } else {
     application <- taskStart(quiet, "Looking up application with id {.val {target$appId}}...")
     application <- tryCatch(
-      client$getApplication(target$appId),
+      {
+        application <- client$getApplication(target$appId, target$version)
+        taskComplete(quiet, "Found application {.url {application$url}}")
+
+        if (application$type == "static") {
```
aronatkins, line 369: This "static" notion is specific to Posit Cloud; does it belong here or with the call to `uploadCloudBundle` (where we already have cloud-specific logic)?

Related: Do we also need to call `createRevision` in the 404 path just below? `applicationDeleted` calls `createApplication` -- could you explain why this path explicitly creates a revision while the creation path does not?

aronatkins (approved):
The wiring from `application$id` to `application$application_id` and changing the meaning of `application$id` is bound to throw us off in the future...

R/client-cloud.R
```diff
@@ -66,20 +66,46 @@ cloudClient <- function(service, authInfo) {
       listRequest(service, authInfo, path, query, "applications")
     },
 
-    getApplication = function(applicationId) {
-      path <- paste("/applications/", applicationId, sep = "")
-      application <- GET(service, authInfo, path)
+    getApplication = function(outputOrApplicationId, deploymentRecordVersion) {
+      if (is.na(deploymentRecordVersion)) {
+        # In pre-versioned dcf files, outputOrApplicationId is the id of the application.
+        # TODO: consider removing support for this case a year after the release of 0.8.30
```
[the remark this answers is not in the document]
mslynch: It's never quite "safe" to remove this block – as long as someone has an old has an old deployment DCF file, this would break for them on removal. We don't have visibility into that, but I'm assuming we don't want to support this case forever.

PR #827, opened 2023-05-01
Deployment record improvements

### LINKED ISSUE #820: Re-deploying old app fails because of `envVars`
```
Error in `deploymentRecord()` at rsconnect/R/deployments.R:93:2:
! `envVars` must be a character vector or `NULL`, not `NA`.
```

### LINKED ISSUE #821: Version the deployment dcf record

hadley:
* Improve `envVars` handling. Fixes #820.
* Add support for `version`. Fixes #821.
* Name arguments to `createDeploymentTarget()` calls

### DIFF
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -83,7 +83,7 @@ deploymentFields <- c(
   "bundleId", "url", "envVars", "version"
 )
 
-deploymentVersion <- 1L
+deploymentRecordVersion <- 1L
 
 saveDeployment <- function(recordDir,
                            target,
@@ -128,7 +128,7 @@ deploymentRecord <- function(name,
                              appId = NULL,
                              bundleId = NULL,
                              url = NULL,
-                             version = deploymentVersion,
+                             version = deploymentRecordVersion,
                              metadata = list()) {
 
   check_character(envVars, allow_null = TRUE)
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -70,6 +70,7 @@ deployments <- function(appPath = ".",
     ok <- ok & okServer
   }
 
+  deployments$envVars[is.na(deployments$envVars)] <- ""
   if (is.character(deployments$envVars)) {
     deployments$envVars <- strsplit(deployments$envVars, ", ")
   }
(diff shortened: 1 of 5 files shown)


R/deployments.R
```diff
@@ -79,9 +80,11 @@ deployments <- function(appPath = ".",
 
 deploymentFields <- c(
   "name", "title", "username", "account", "server", "hostUrl", "appId",
-  "bundleId", "url", "envVars"
+  "bundleId", "url", "envVars", "version"
 )
 
+deploymentVersion <- 1L
```
hadley, line 86: Maybe this should be `deploymentRecordVersion`?

aronatkins: I like that better.

tests/testthat/test-deployments.R
```diff
@@ -69,6 +69,14 @@ test_that("can read/write metadata", {
   expect_equal(out$meta2, "two")
 })
 
+test_that("can read/write version", {
```
aronatkins, line 72: Where do we test reading a DCF file that lacks the `envVars` and `version` fields?

We should test reading single files with/without these fields as well as a collection of files where some of them lack these fields.

aronatkins (approved):
The `envVars` and `version` fields are new and should not cause much difficulty if folks roll-back to an old release; old rsconnect will parse those fields, but may remove them on write.

Should we add something to NEWS about the consequences to stored `envVars` if folks upgrade then downgrade?

https://github.com/rstudio/rsconnect/blob/v0.8.29/R/deployments.R

tests/testthat/test-deployments.R
```diff
@@ -80,10 +88,15 @@ test_that("can read/write env vars", {
 
 test_that("can read env vars when none set", {
   app <- local_temp_app()
-  addTestDeployment(app, "test1", envVars = NA_character_)
+  path <- addTestDeployment(app, "test1", envVars = NULL)
+  deps <- deployments(app, excludeOrphaned = FALSE)
+  expect_equal(deps$envVars, list(character()))
+  expect_false("envVars" %in% rownames(read.dcf(path)))
```
hadley, line 94: @aronatkins this is where we test what happens when we deliberately create a deployment without `envVars` — we test that `envVars` is not written to the DCF, but it's still successfully read. I'll add something similar for `version`.

mslynch (approved):

tests/testthat/test-deployments.R
aronatkins: Thanks. I missed that this test was confirming the processing of an "old" DCF. Would you mind adding a comment / renaming the test to remind us that this is testing DCF compatibility upon upgrade?

PR #833, opened 2023-05-03
Sketch dryRun() implementation

### LINKED ISSUE #725: Deployment dry run
Make it easier to diagnose issues locally by running the app in an environment that simulates what you'll get on the server.

hadley:
To fix #725

### DIFF
--- a/R/dryRun.R
+++ b/R/dryRun.R
@@ -1,0 +1,153 @@
+#' Perform a deployment "dry run"
+#'
+#' @description
+#' `r lifecycle::badge("experimental")`
+#'
+#' `dryRun()` runs your app locally, attempting to simulate what will happen
+#' when you deploy it on another machine. This isn't a 100% reliable way of
+#' discovering problems, but it offers a much faster iteration cycle, so where
+#' it does reveal a problem, it will typically make identifying and fixing it
+#' much faster.
+#'
+#' This function is still experimental, so please let us know your experiences
+#' and where we could do better:
+#' <https://github.com/rstudio/rsconnect/issues/new>
+#'
+#' ## Where it helps
+#'
+#' `dryRun()` was motivated by the two most common problems when deploying
+#' your app:
+#'
+#' * The server doesn't install all the packages your app needs to work.
+#'   `dryRun()` solves this by using `renv::restore()` to create a project
+#'   specific library that uses only the packages that are explicitly used
+#'   by your project.
+#'
+#' * The server doesn't have environment variables you need. `dryRun()`
+#'   solves this by removing any environment variables that you've
+#'   set in `~/.Renviron`, except for those that you declare in `envVars`.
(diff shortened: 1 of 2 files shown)


R/dryRun.R
```diff
@@ -0,0 +1,175 @@
...
+    wd = bundleDir,
+    show = TRUE
+  )
+  invisible()
+}
+
+appRunner <- function(appMode) {
+  switch(appMode,
+    "rmd-static" = ,
+    "rmd-shiny" = function(primaryDoc) {
```
aronatkins, line 136: `rmd-shiny` should use `rmarkdown::run(file = NULL, dir = getwd())`

R/dryRun.R
```diff
@@ -0,0 +1,175 @@
...
+    "quarto-shiny" = function(primaryDoc) {
+      quarto::quarto_render(primaryDoc, quiet = TRUE)
+    },
+    "shiny" = function(primaryDoc) {
+      shiny::runApp()
+    },
+    "api" = function(primaryDoc) {
+      plumber::pr_run(plumber::pr("plumber.R"))
+    },
+    cli::cli_abort("Content type {appMode} not currently supported")
```
aronatkins, line 149: Rendered content and "static" -- we could start an HTML server against the `bundleDir` so they could see exactly what files they are about to deploy. Maybe that would help cases where folks forget to include some images/CSS.

The result of this would be that _every_ content type starts some type of HTTP server that they would interact with for validation.

aronatkins (approved):

R/dryRun.R
hadley: I already set the working directory above, but I'll take the other changes.

aronatkins: The `::run(..)` signature is necessary to have R Markdown do its "run the whole directory" magic.. it's not setting the working directory. I think that's this logic: https://github.com/rstudio/rmarkdown/blob/main/R/shiny.R#L83-L107

hadley: I just meant the `dir` argument

aronatkins: OK. Once you give `file = NULL`, the `dir` value is necessary, as its default value of `dirname(file)` is meaningless.

R/dryRun.R
```diff
@@ -0,0 +1,175 @@
...
+#'
+#' * `dryRun()` currently offers no way to diagnose problems with
+#'   mismatched R/Python/Quarto/pandoc versions.
+#'
+#' * `dryRun()` doesn't help much with paths. There are two common problems
+#'   it can't help with: using an absolute path and using the wrong case.
+#'   Both of these will work locally but fail on the server. [lint()]
+#'   uses an alternative technique (static analysis) to detect many of these
+#'   cases.
+#'
```
andpatt, line 55: Another limitation is that:

* `dryRun()` doesn't help if the server does not have the system dependencies that are present 
  on the machine on which you're running this dryRun().

PR #834, opened 2023-05-03
Support quarto sites in `deploySite()`

### LINKED ISSUE #813: deploySite does not work for non-RMarkdown sites
`deploySite()` looks specifically for static files and directories generated with rmarkdown, so it will not work for sites generated by Quarto because the static artifacts are slightly different.

hadley:
Fixes #813

I did a bunch of manual tests as described in `tests/manual/dpeloySite.Rmd`. 

This also revealed (again) that `recordDir` isn't a real directory, so we shouldn't test for its existence in `deployApp()`.


R/deploySite.R
```diff
@@ -102,3 +91,65 @@ deploySite <- function(siteDir = getwd(),
...
+    call = error_call
+  )
+
+  config <- quarto::quarto_inspect(path)$config
+
+  list(
+    render = function() quarto::quarto_render(path, quiet = quiet, as_job = FALSE),
+    name = basename(normalizePath(path)),
+    title = config$website$title %||% config$book$title %||% config$title,
+    output_dir = outputDir(path, config$project$`output-dir`)
```
aronatkins, line 108: Quarto projects that are not websites or books do not have a declared output directory; they render into the working directory.

For example, use the `_quarto.yml` alongside an `index.qmd` and `second.qmd`:

```
project:
  title: "project"
```

Running `quarto render` produces `index.html`, `second.html` and `*_files` directories for each. All are written into the working directory, not a separate output directory.

Is this configuration a "site"? It's definitely a "project". How is this project best deployed? What happens if I `render="local"`?

In `quarto::quarto_publish_site`, it looks like we publish the entire input directory, including the source files, even when we have rendered the content.

R/deploySite.R
```diff
@@ -35,61 +38,47 @@ deploySite <- function(siteDir = getwd(),
...
+  quiet <- identical(match.arg(logLevel), "quiet")
+  if (isQuarto) {
+    site <- quartoSite(siteDir, quiet = quiet)
+  } else {
+    site <- rmarkdownSite(siteDir, quiet = quiet)
   }
 
   # render locally if requested
-  render <- match.arg(render)
+  render <- arg_match(render)
```
aronatkins, line 53: note: Client-side rendering by rsconnect is supported by `deploySite` but not `deployDoc`. Rendering is performed external to rsconnect for single file deploys.

R/deploySite.R
```diff
@@ -102,3 +91,65 @@ deploySite <- function(siteDir = getwd(),
...
+  siteGenerator <- rmarkdown::site_generator(siteDir)
+  if (is.null(siteGenerator)) {
+    cli::cli_abort(
+      "No `_site.yml` found in {.path {siteDir}.}",
+      call = error_call
+    )
+  }
+
+  list(
+    render = function() {
```
aronatkins, line 130: How different is this from calling `rmarkdown::render_site`?

The `render_site` function returns the name of the output file, which is generally `output/index.html` or `./index.html`, and the output directory can be derived from it, leaving a render function that returns its output location looking something like:

```r
render = function() {
    result <- rmarkdown::render_site(...)
    return(dirname(result))
}
```

aronatkins (approved):
Confirmed that `quarto::quarto_publish_site` calls `rsconnect::deployApp` not `rsconnect::deploySite`, so there are not circular calls.

Should we update the Quarto R package to call this function, it will need to be version-constrained.

R/deploySite.R
hadley: I don't know, I just copied the existing code 😄. If you think it's likely to work, I'm happy to switch (and I'll re-run the appropriate tests).

R/deploySite.R
aronatkins: I think of this project as a website without all the automatic treatment that comes with being a Quarto website.

I'm also OK with including the entire working directory post-rendering; mostly, I wanted you to know that this type of project has no `output-dir` in the output from `quarto inspect`.

R/deploySite.R
aronatkins: It's quite similar to `render_site`, but I don't think it's worth the bother. Mostly, I was leaning into the idea that the `render()` could return the `output-dir` rather than computing it separately (since it is only necessary when rendering).

R/deploySite.R
hadley: Oh got it; I'll add a fallback.

R/deploySite.R
hadley: If I have time, I think I should implement this for `deployDoc()` too, to bring closer to `quarto_publish_doc()`.

PR #838, opened 2023-05-08
Don't allow user to change visibility on cloud

### LINKED ISSUE #806: Fix visibility management when published to Posit Cloud
`deployApp` contains an optional `appVisibility` argument. If the app is being deployed to shinyapps.io, this works as expected to control whether the app is initially public or private.

On posit cloud, although truly public content is not currently supported, content _can_ be set to public or private, and this affects the set of Cloud users who can see it. For now we should probably update `rsconnect` to set `/content/{id}`'s `visibility` when the server is `posit.cloud`

hadley:
Fixes #806

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # rsconnect (development version)
 
+* `deployApp()` now gives an actionable error if you attempt to set
+  visiblity of an app deployed to posit.cloud (#838).
+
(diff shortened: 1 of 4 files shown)


mbaynton (approved):

hadley:
Hmmmmm, probably should just to be safe, but I doubt it will affect many people.

PR #839, opened 2023-05-08
Add manual tests for deployment

hadley:
@aronatkins is there anything else that you commonly do that I should add to this list?

### DIFF
--- a/tests/manual/publishing-dialog.Rmd
+++ b/tests/manual/publishing-dialog.Rmd
@@ -33,3 +33,46 @@ Success:
 
 1. No errors in console
 1. `nrow(rsconnect::accounts("shinyapps.io"))` returns 1 or greater.
+
+# Apps
+
+1. Create a new shiny app by clicking "New Project" | "New Directory" | "Shiny app"
+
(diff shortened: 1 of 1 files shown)


tests/manual/publishing-dialog.Rmd
```diff
@@ -33,3 +33,46 @@ Success:
 
 1. No errors in console
 1. `nrow(rsconnect::accounts("shinyapps.io"))` returns 1 or greater.
+
+# Apps
```
aronatkins, line 37: The combinatorics can quickly explode..

1. having renv.lock
2. without renv.lock
3. environment that lacks some referenced package (error)

It's probably not worth enumerating every "it's an error" case, but the "missing package" error felt common enough to call out.

aronatkins (approved):
1️⃣ ⚫ ⚫ ⚫  3️⃣ 💰

tests/manual/publishing-dialog.Rmd
hadley: FWIW those are in `dependencies.Rmd`

PR #844, opened 2023-05-08
Make `local_temp_app()` usage more consistent

hadley:
And make it a little easier to use with `...`


tests/testthat/test-bundleFiles.R
```diff
@@ -146,23 +149,23 @@ test_that("ignores python virtual envs", {
 # explodeFiles ------------------------------------------------------------
 
 test_that("returns relative paths", {
-  dir <- withr::local_tempdir()
+  dir <- local_temp_app()
```
aronatkins, line 152: Do you expect that this will become:

```r
dir <- local_temp_app(
  "x/a" = "",
  "x/b" = "",
  "x/c" = "'
)
```

`local_temp_app` would need to change to `dir.create(dirname(files))`, but that's in reach, if you want it.

The challenge would be the cases where you want to create empty directories as part of the test set-up. Maybe:

```r
dir <- local_temp_app(
    "packrat/" = ""
)
```

None of this is must-take. Just musing on ways to simply describe "create this set of files / directories" without requiring every test to have its own creation code.

hadley: Yeah, I had contemplated that too. If you were thinking the same thing, I think it's a sign that it's worth doing.

Issue #852, opened 2023-05-23
Deploys to shinyapps.io fail


mslynch:
Recent changes (still unreleased) have broken deployments to shinyapps.io, which now fail with the error:

```
> rsconnect::deployApp("/Users/matthew/code/mine/shiny-deploy-test", server='shinyapps.io')
── Preparing for deployment ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
✔ Deploying "shinydeploytest17" to "server: shinyapps.io / username: 9kimvo-matthew-lynch"
ℹ Creating application on server...
✔ Created application with id 1865337
ℹ Bundling 1 file: app.R
ℹ Capturing R dependencies with renv
✔ Found 31 dependencies
✔ Created 17,592b bundle
ℹ Uploading bundle...
Error in `POST()` at rsconnect/R/http.R:242:2:
! <https://api.staging.posit.cloud/v1/bundles> failed with HTTP status 400
Request Validation Error: : field 'application': is required
```

PR #853, opened 2023-05-23
fix shinyapps.io deploys

mslynch:
We changed our call to `uploadCloudBundle` to pass in `application$application_id` to support including the output id in `application$id`, but `application$application_id` wasn't populated for shinyapps.

Resolves https://github.com/rstudio/rsconnect/issues/852.

### DIFF
--- a/R/client-cloud.R
+++ b/R/client-cloud.R
@@ -146,11 +146,11 @@ cloudClient <- function(service, authInfo) {
       output <- POST_JSON(service, authInfo, "/outputs", json)
       path <- paste0("/applications/", output$source_id)
       application <- GET(service, authInfo, path)
-      application$application_id <- application$id
-      application$id <- output$id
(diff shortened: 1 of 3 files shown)


omar-rs (approved):

hadley (approved):
Does `client-connect.R` need a similar fix? I presume not, but it's hard to tell from the code.

I wonder if it's worthwhile to explicitly construct a named list from each `createApplication()` "method" so we can see exactly what fields are needed?

mslynch:
> Does `client-connect.R` need a similar fix? I presume not, but it's hard to tell from the code.

Nope, only shinyapps and cloud follow this path.

> I wonder if it's worthwhile to explicitly construct a named list from each `createApplication()` "method" so we can see exactly what fields are needed?

I like that idea – pushed a commit to implement.

hadley:
That looks much better 😄

omar-rs (approved):

PR #865, opened 2023-06-21
provide the guid when creating content

### LINKED ISSUE #859: Failure to add environment variables to content
When publishing a new content item with environment variables, the initial API call to deploy the content succeeds, but the secondary call to add environment variables fails with a 404 error.

I saw this both with a local dev Connect environment and with a proper instance. Logs from the local dev example attached.

```
Completed deployed execution
── Deployment complete ─────────────────────────────────────────────────────────────────────────────────────────────
✔ Successfully deployed to <http://localhost:3939/content/6441c48e-771b-4db6-ab58-fda27a036386/>
> rsconnect::deployDoc("index.Rmd", appName = "rmd-with-secret", envVars = c(SECRET = "forty-two"))
ℹ Discovering document dependencies...
✔ Document dependencies discovered
── Preparing for deployment ────────────────────────────────────────────────────────────────────────────────────────
Found multiple accounts.
Which one do you want to use?
1: server: localhost / username: toph
2: server: rsc.radixu.com / username: toph
Selection: 1
✔ Deploying "rmd-with-secret" to "server: localhost / username: toph"
✔ Created application with id 525
ℹ Updating environment variables forty-two...
Error in `PATCH()`:
! <http://localhost:3939/__api__> failed with HTTP status 404
404 page not found
```

aronatkins:
fixes #859

### DIFF
--- a/R/client-connect.R
+++ b/R/client-connect.R
@@ -59,6 +59,7 @@ connectClient <- function(service, authInfo) {
       application <- POST_JSON(service, authInfo, "/applications", details)
       list(
         id = application$id,
+        guid = application$guid,
         url = application$url
       )
     },
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -361,7 +361,7 @@ deployApp <- function(appDir = getwd(),
     )
     taskComplete(quiet, "Created application with id {.val {application$id}}")
   } else {
-    application <- taskStart(quiet, "Looking up application with id {.val {target$appId}}...")
+    taskStart(quiet, "Looking up application with id {.val {target$appId}}...")
     application <- tryCatch(
       {
         application <- client$getApplication(target$appId, target$version)


R/deployApp.R
```diff
@@ -361,7 +361,7 @@ deployApp <- function(appDir = getwd(),
     )
     taskComplete(quiet, "Created application with id {.val {application$id}}")
   } else {
-    application <- taskStart(quiet, "Looking up application with id {.val {target$appId}}...")
+    taskStart(quiet, "Looking up application with id {.val {target$appId}}...")
```
hadley, line 364: I don't see `guid` used here, but I assume it is further down.

R/client-connect.R
```diff
@@ -59,6 +59,7 @@ connectClient <- function(service, authInfo) {
       application <- POST_JSON(service, authInfo, "/applications", details)
       list(
         id = application$id,
+        guid = application$guid,
```
hadley, line 62: Need same change for other clients?

hadley (approved):

R/deployApp.R
aronatkins: This was an unrelated mistake; we double-assigned `application` but the first assignment (from `taskStart`) was incorrect and immediately overwritten by the next statement. Found this while trying to trace the definition of `application`.

R/client-connect.R
aronatkins: I don't think so. Only connect has `guid`. See the set of fields that are "kept" from API list responses in `applications()`.

PR #872, opened 2023-06-23
rely on README.md, not Rmd

aronatkins:
Any reason to preserve the R Markdown -> GitHub Markdown conversion?


hadley:
Probably not that important now, but it does make it easy to add actual running examples. But maybe that's not terribly feasiable for rsconnect. I don't have strong feelings either way.

aronatkins:
The Markdown and R Markdown versions have occasionally fallen out-of-sync, thus this PR. I'm open to other approaches which keep the README in sync.

hadley:
I guess I don't worry too much about the `README.md` getting out of sync because I know one of the items on the release checklist is to run `devtools::build_readme()`, so there is a guaranteed sync point prior to release.

aronatkins:
We have been following https://github.com/rstudio/rsconnect/blob/main/RELEASE.md, which doesn't include that step (unfortunately). It should.

hadley:
Ah we should switch much of that to `usethis::use_release_issue()`

aronatkins:
Updating our release checklist to `devtools::build_readme()` rather than shifting to the Markdown version.

Issue #871, opened 2023-06-23
Republishing to posit.cloud fails after updating rsconnect


ronblum:
1. Install `rsconnect` 0.8.29.
2. <kbd>File -> New Project -> New Directory -> Shiny Application</kbd>
3. Provide directory name and click <kbd>Create Project</kbd>.
4. Publish to posit.cloud.
5. Update `rsconnect` to dev version.
6. Republish.

Republishing fails in the UI, with the message "No output found for id=" etc.

<img width="280" alt="no_output_found" src="https://github.com/rstudio/rsconnect/assets/31009952/a112f310-63b7-472e-8f24-4b3873bfbad0">

1. Fails for R Markdown as well, not just Shiny app.
2. Works with rsc.radixu.com.
3. Republishing with the same version of `rsconnect` works.
4. Tested with
    - Posit Workbench 2023.06.0+421.pro1 and R 3.6.3 on Ubuntu 20.04.5 via Chrome on MacOS 13.4.
    - RStudio Desktop 2023.08.0-daily+46 and R 4.3.0 on MacOS 13.4.

aronatkins:
@mslynch - This is likely related to the posit.cloud identifier changes -- could you investigate if this upgrade problem is genuine?

mslynch:
This is genuine. The problem is that the IDE looks up applications using the appId, but we don't have the context to know whether it's the id of the content or the application. We may have to walk back on our DCF versioning approach and discover which it is through API calls. The alternative would be to make IDE changes to accommodate this, but I'd prefer not to go that way.

cc @omar-rs @hadley

ronblum:
Verified with the development version (0.8.29.9000) of `rsconnect` using RStudio Server 2023.06.0+421.pro1 on Ubuntu 20.04.5 via Chrome on MacOS 13.4.

PR #873, opened 2023-06-23
allow RStudio to call getApplication for Posit Cloud without knowing whether the id is for application or content

mslynch:
Resolves https://github.com/rstudio/rsconnect/issues/871.

### DIFF
--- a/R/client-cloud.R
+++ b/R/client-cloud.R
@@ -100,7 +100,7 @@ cloudClient <- function(service, authInfo) {
 
             list(application = application, output = output)
           }
-        tryCatch(
+        applicationAndOutput <- tryCatch(
           {
             path <- paste0("/outputs/", outputOrApplicationId)
             output <- GET(service, authInfo, path)
@@ -113,6 +113,8 @@ cloudClient <- function(service, authInfo) {
           rsconnect_http_403 = handleError,
           rsconnect_http_404 = handleError
         )
+        application <- applicationAndOutput$application
+        output <- applicationAndOutput$output
       } else {
         # from dcf version >= 1, outputOrApplicationId is the id of the output.
         path <- paste0("/outputs/", outputOrApplicationId)
--- a/R/client-cloud.R
+++ b/R/client-cloud.R
@@ -90,16 +90,16 @@ cloudClient <- function(service, authInfo) {
         output <- GET(service, authInfo, path)
       } else if (deploymentRecordVersion == "unknown") {
         handleError <- function(err) {
-            path <- paste0("/applications/", outputOrApplicationId)
-            application <- GET(service, authInfo, path)
+          path <- paste0("/applications/", outputOrApplicationId)
+          application <- GET(service, authInfo, path)
 
-            output_id <- application$output_id %||% application$content_id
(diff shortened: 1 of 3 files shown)


R/client-cloud.R
```diff
@@ -86,6 +88,31 @@ cloudClient <- function(service, authInfo) {
 
         path <- paste0("/outputs/", output_id)
         output <- GET(service, authInfo, path)
+      } else if (deploymentRecordVersion == "unknown") {
```
mslynch, line 91: Adding this condition means the other conditions aren't strictly necessary. I'm on the fence about whether we should keep the DCF versioning and version handling logic.

R/client-cloud.R
```diff
@@ -86,6 +88,31 @@ cloudClient <- function(service, authInfo) {
...
+        applicationAndOutput <- tryCatch(
```
omar-rs, line 92: Is there a possibility of finding an output with an application id (if it's very old, with a low id number) or vice versa?  If so, should we do additional sanity checks on the application type, or other expected properties?

mslynch: I checked our production data and found that there are no collisions between output ids and output application ids. As long as that's true, we're safe. The user would also have to have permissions for the application, which makes this case even more unlikely. I don't think there's any sanity checks we can do otherwise.

omar-rs (approved):

R/client-cloud.R
hadley: I feel like it's probably worth it, because it does save an API call, and we can eventually get the right logic into the IDE too.

R/client-cloud.R
```diff
@@ -86,6 +88,33 @@ cloudClient <- function(service, authInfo) {
...
+            path <- paste0("/applications/", outputOrApplicationId)
+            application <- GET(service, authInfo, path)
+
+            output_id <- application$output_id %||% application$content_id
+
+            path <- paste0("/outputs/", output_id)
+            output <- GET(service, authInfo, path)
+
+            list(application = application, output = output)
+          }
```
hadley, line 102: Indenting?

R/client-cloud.R
```diff
@@ -86,6 +88,33 @@ cloudClient <- function(service, authInfo) {
...
+            path <- paste0("/applications/", output$source_id)
+            application <- GET(service, authInfo, path)
+
+            list(application = application, output = output)
+          },
+          rsconnect_http_403 = handleError,
+          rsconnect_http_404 = handleError
+        )
+        application <- applicationAndOutput$application
+        output <- applicationAndOutput$output
```
hadley, line 117:
```suggestion
        tryCatch(
          {
            path <- paste0("/outputs/", outputOrApplicationId)
            output <- GET(service, authInfo, path)

            path <- paste0("/applications/", output$source_id)
            application <- GET(service, authInfo, path)

            list(application = application, output = output)
          },
          rsconnect_http_403 = handleError,
          rsconnect_http_404 = handleError
        )
```

?

hadley (approved):

omar-rs (approved):

PR #879, opened 2023-06-26
allow multiple servers with same URL

### LINKED ISSUE #878: error when registering account
When multiple `servers()` have the same URL:

https://github.com/rstudio/rsconnect/blob/f3725d77260f412b4be0a1228486b63a5d8d5656/R/ide.R#L121-L122

```
<error/rlang_error>
Error in `serverInfo()`:
! `server` must be a single string, not a character vector.
---
Backtrace:
     ▆
  1. └─.rs.rpc.get_user_from_token(...)
  2.   └─rsconnect:::getUserFromRawToken(...)
  3.     └─rsconnect:::waitForAuthedUser(server, token = token, private_key = privateKey)
  4.       ├─base::tryCatch(...)
  5.       │ └─base (local) tryCatchList(expr, classes, parentenv, handlers)
  6.       │   └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
  7.       │     └─base (local) doTryCatch(return(expr), name, parentenv, handler)
  8.       └─rsconnect:::getAuthedUser(...)
  9.         └─rsconnect:::clientForAccount(account)
 10.           └─rsconnect::serverInfo(name = account$server)
 11.             └─rsconnect:::findServer(name, local = FALSE)
 12.               └─rsconnect:::check_string(server, call = error_call)
 13.                 └─rsconnect:::stop_input_type(...)
 14.                   └─rlang::abort(message, ..., call = call, arg = arg)
```

aronatkins:
additionally, use the "ensured" URL when finding a local server.

fixes #878

### DIFF
--- a/R/ide.R
+++ b/R/ide.R
@@ -18,7 +18,7 @@ validateServerUrl <- function(url, certificate = NULL) {
 # registered server is returned.
 findAndRegisterLocalServer <- function(url) {
   # helper to find a server given its URL
-  findServerByUrl <- function(name) {
+  findServerByUrl <- function(url) {
     allServers <- rsconnect::servers(local = TRUE)
     match <- allServers[allServers$url == url, , drop = FALSE]
     if (nrow(match) == 0)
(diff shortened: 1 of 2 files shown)


hadley (approved):
Worth a test IMO

aronatkins:
> Worth a test IMO

Done. I had assumed that adding tests would be too involved for this change, since most of our HTTP tests were only against the lowest-level HTTP helpers. I think this does what we need.

hadley:
I didn't realise there would be so much HTTP mocking involved, but overall doesn't seem too bad.

PR #883, opened 2023-06-27
Consistent format for package record

### LINKED ISSUE #881: manifest schema is different if using packrat
If I enable packrat:
```
options(rsconnect.packrat = TRUE)
```

And generate a manifest for an R Markdown doc:
```
rsconnect::writeManifest(appFiles=c("mydoc.Rmd"))
```

The manifest it generates has a different schema than if I do not enable packrat. In particular, the `description` of each package is an array rather than an object.

### With packrat:
```
  "packages": {
    "R6": {
      "Package": "R6",
      "Version": "2.5.1",
      "Source": "CRAN",
      "Repository": "http://rspm/default/__linux__/focal/latest",
      "description": ["Package: R6", ... etc ]
```

### Without packrate:
```
  "packages": {
    "R6": {
      "Package": "R6",
      "Version": "2.5.1",
      "Source": "CRAN",
      "Repository": "http://rspm/default/latest",
      "description": {
        "Package": "R6",
```

hadley:
Fixes #881

### DIFF
--- a/R/bundlePackage.R
+++ b/R/bundlePackage.R
@@ -23,6 +23,9 @@ bundlePackages <- function(bundleDir,
     out
   })
   names(packages_list) <- deps$Package
+
+  packages_list$Package <- NULL
+  packages_list$Version <- NULL
   packages_list
 }
 
--- a/R/bundlePackage.R
+++ b/R/bundlePackage.R
@@ -93,3 +93,7 @@ availablePackages <- function(repos) {
     filters = getOption("available_packages_filters", default = "duplicates")
   )
 }
+
+package_record <- function(name, lib_dir = NULL) {
+  readLines(system.file("DESCRIPTION", package = name, lib.loc = lib_dir))
+}
--- a/R/bundlePackageRenv.R
+++ b/R/bundlePackageRenv.R
@@ -5,8 +5,8 @@ snapshotRenvDependencies <- function(bundleDir,
 
   old <- options(
     renv.verbose = FALSE,
-    renv.consent = TRUE,
-    pkgType = "source")
+    renv.consent = TRUE
+  )
   defer(options(old))
 
   renv::snapshot(bundleDir, prompt = FALSE)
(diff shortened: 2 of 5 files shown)


R/bundlePackageRenv.R
```diff
@@ -5,8 +5,8 @@ snapshotRenvDependencies <- function(bundleDir,
 
   old <- options(
     renv.verbose = FALSE,
-    renv.consent = TRUE,
-    pkgType = "source")
```
hadley, line 9: Not sure this is correct; need to discuss.

hadley: Talked this over with @kevinushey, and I think the new behaviour is more correct — this should reduce snapshot time in the worst case where you have an existing lockfile and your library is out of sync.

R/bundlePackage.R
```diff
@@ -93,3 +93,7 @@ availablePackages <- function(repos) {
     filters = getOption("available_packages_filters", default = "duplicates")
   )
 }
+
+package_record <- function(name, lib_dir = NULL) {
```
aronatkins, line 97: Is it worth a test confirming the shape of the resulting JSON?

aronatkins (approved):

aronatkins:
One other question: Do we want to remove the additional `Package=R6` and `Version=2.5.1` fields? They are not produced by the CRAN release.

hadley:
@aronatkins I have a vague recollection we added those so that we could eventually drop the `description` entries altogether.

aronatkins:
> I have a vague recollection we added those so that we could eventually drop the `description` entries altogether.

Hm. I don't recall. Maybe that was when we were thinking of always including the `renv.lock` and using that to drive the restore? Today, the `description` is needed to recalculate cache location since it captures dependencies.

hadley:
Ok, if it doesn't help, then we might as well drop them.

aronatkins (approved):

PR #886, opened 2023-06-28
Always record quarto version in manifest

### LINKED ISSUE #885: deployDoc and deploySite do not include the quarto version
Note, this testing was done when deploying to Posit Cloud

If I run `rsconnect::deployDoc` or `rsconnect::deploySite`, the resulting manifest does not include the quarto version used. However, if I use `rsconnect::writeManifest` the resulting manifest does include the quarto version.

For cloud to do the right thing, it will need to know the version of quarto.

hadley:
Fixes #885

This will now send the `quarto` field to shinyapps.io. Is that ok?

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -108,7 +108,6 @@ createAppManifest <- function(appDir,
                               users = NULL,
                               pythonConfig = NULL,
                               retainPackratDirectory = TRUE,
-                              isCloudServer = FALSE,
                               image = NULL,
                               verbose = FALSE,
                               quiet = FALSE) {
@@ -198,9 +197,8 @@ createAppManifest <- function(appDir,
   }
 
   # indicate whether this is a quarto app/doc
-  if (!is.null(appMetadata$quartoInfo) && !isCloudServer) {
-    manifest$quarto <- appMetadata$quartoInfo
-  }
+  manifest$quarto <- appMetadata$quartoInfo
+
   # if there is python info for reticulate or Quarto, attach it
   if (!is.null(python)) {
     manifest$python <- python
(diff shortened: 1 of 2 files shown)


R/bundle.R
```diff
@@ -198,9 +197,8 @@ createAppManifest <- function(appDir,
   }
 
   # indicate whether this is a quarto app/doc
-  if (!is.null(appMetadata$quartoInfo) && !isCloudServer) {
-    manifest$quarto <- appMetadata$quartoInfo
-  }
+  manifest$quarto <- appMetadata$quartoInfo
```
aronatkins, line 200: When `appMetadata$quartoInfo` is null, we do not write a Quarto object, correct? The null assignment is interpreted as a no-op?

aronatkins (approved):
Let's get confirmation from the hosted folks before merging, but I love that the manifest is (more? completely?) service agnostic.

The exception might be with how we infer app-mode (`appMetadata(coerceStaticRmd=TRUE)`), but that will eventually get relaxed, too.

R/bundle.R
hadley: Right:

``` r
x <- list()
x$x <- NULL
x
#> list()
```

<sup>Created on 2023-06-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

aronatkins: Thanks. I didn't look to see how `manifest` was constructed.

aronatkins:
Chatted with @samperman - merging this to make their testing easier; they don't expect Quarto fields in any currently-supported content.

PR #887, opened 2023-06-28
Turn parallel testing on

hadley:
(There's a bug in processx that can leave test processes run if you abort the tests during setup; that's fixed by `pak::pak("r-lib/processx")`).


hadley:
It's the testing processes started by testthat; webfakes should automatically cleanup already.

Issue #901, opened 2023-07-06
Redeploys to shinyapps.io fail


mslynch:
Happens on `main` only.

```
── Preparing for deployment ────────────────────────────────────────────────────
✔ Re-deploying "20230706_redeploy_shiny_R_430_new_rsconnect" to "server: shinyapps.io / username: disk"
ℹ Looking up application with id "9417386"...
✔ Found application <https://disk.shinyapps.io/20230706_redeploy_shiny_R_430_new_rsconnect/>
ℹ Bundling 1 file: 'app.R'
ℹ Capturing R dependencies with renv
✔ Found 31 dependencies
✔ Created 17,524b bundle
ℹ Uploading bundle...
Error in `POST()`:
! <https://api.shinyapps.io/v1/bundles> failed with HTTP status 400
Request Validation Error: : field 'application': is required
Backtrace:
     ▆
  1. └─rsconnect::deployApp(...)
  2.   └─rsconnect:::uploadCloudBundle(...)
  3.     └─client$createBundle(...)
  4.       └─rsconnect:::POST_JSON(service, authInfo, "/bundles", json)
  5.         └─rsconnect:::POST(...)
  6.           └─rsconnect:::httpRequestWithBody(...)
  7.             └─rsconnect:::handleResponse(httpResponse, error_call = error_call)
  8.               └─rsconnect (local) reportError(json$error)
  9.                 └─cli::cli_abort(...)
 10.                   └─rlang::abort(...)
Execution halted
```

dskard:
This works for me manually now:

```R
# install github version of rsconnect
install.packages("remotes")
remotes::install_github("rstudio/rsconnect")

# TODO: create your own shiny app and change the directory name here
app_dir <- "20230707_1_redeploy_shiny_R_430_github_rsconnect"
app_name <- app_dir

# TODO: fill in your own account, token, secret details
shinyapps_account <- "..."
shinyapps_server <- "shinyapps.io"
shinyapps_token <- "..."
shinyapps_secret <- "..."

# setup a shinyapps.io account to publish to
rsconnect::setAccountInfo(
  name = shinyapps_account,
  token = shinyapps_token,
  secret = shinyapps_secret,
  server = shinyapps_server
)

# first deploy
rsconnect::deployApp(
  app_dir,
  appName = app_name,
  account = shinyapps_account,
  server = shinyapps_server
)

# second deploy
rsconnect::deployApp(
  app_dir,
  appName = app_name,
[...]

PR #902, opened 2023-07-06
include application_id in client-shinyapps getApplication result

mslynch:
We're looking for `application$application_id` when we deploy the bundle. This is currently present on the `createApplication` return value, but not for `getApplication`.

Resolves https://github.com/rstudio/rsconnect/issues/901.


omar-rs (approved):

aronatkins (approved):
Is there an opportunity for tests?

mslynch:
> Is there an opportunity for tests?

I'm sure there is... shinyapps.io doesn't have any unit tests to build off of and adding something meaningful would take more time than I'm willing to spend right now. I created https://github.com/rstudio/rsconnect/issues/903.

PR #922, opened 2023-07-20
Make bundlePackages() more robust

hadley:
This ensures that we work better with both dev renv and with recommended packages that are bundled with R itself.

### DIFF
--- a/R/bundlePackageRenv.R
+++ b/R/bundlePackageRenv.R
@@ -46,7 +46,12 @@ parseRenvDependencies <- function(bundleDir, snapshot = FALSE) {
     renv::restore(bundleDir, library = lib_dir, prompt = FALSE)
     defer(unlink(lib_dir, recursive = TRUE))
 
-    deps$description <- lapply(deps$Package, package_record, lib_dir = lib_dir)
+    deps$description <- lapply(
+      deps$Package,
+      package_record,
+      # Ensure we fall back to system libraries
+      lib_dir = c(lib_dir, .libPaths())
+    )
   }
 
   deps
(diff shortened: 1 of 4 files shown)


R/bundlePackageRenv.R
```diff
@@ -46,7 +46,12 @@ parseRenvDependencies <- function(bundleDir, snapshot = FALSE) {
     renv::restore(bundleDir, library = lib_dir, prompt = FALSE)
     defer(unlink(lib_dir, recursive = TRUE))
 
-    deps$description <- lapply(deps$Package, package_record, lib_dir = lib_dir)
+    deps$description <- lapply(
+      deps$Package,
+      package_record,
+      # Ensure we fall back to system libraries
```
aronatkins, line 52: Aside: In an renv project, this is the set of libraries managed by renv.. This would have made yesterday's renv changes unnecessary, as the incoming library within the deploy pane included the two renv paths and my system library.

In other words: When deploying an renv project, the `renv::restore` run by rsconnect might be unnecessary since the incoming `.libPaths()` should already have all the dependencies available.

Do you recall why we added the restore?

aronatkins (approved):

kevinushey (approved):

R/bundlePackageRenv.R
hadley: It ensures that we get the `DESCRIPTION` for the versions declared in your lockfile, rather than the versions you currently have installed in your library.

But maybe we should check `renv::status()$sychronised` and only restore if that's `FALSE`? That seems like a nice optimisation.

aronatkins: Feels like that optimization can come later. Feels like we can just merge as-is, then.

hadley: I'll make a note.

PR #933, opened 2023-07-26
don't restrict shinyapps.io from previous deployment check in deploymentTarget

mslynch:


aronatkins (commented):
Once you prove that this works, NEWS and test, please.

tests/testthat/test-deploymentTarget.R
```diff
@@ -187,42 +187,46 @@ test_that("default title is the empty string", {
...
-      url = "http://example.com/test",
-      stringsAsFactors = FALSE
-    ),
-    shouldUpdateApp = function(...) TRUE
-  )
-
-  app_dir <- withr::local_tempdir()
-  target <- deploymentTarget(app_dir, appName = "my_app")
-  expect_equal(target$appId, 123)
+  for (server in c("example.com", "shinyapps.io")) {
```
aronatkins, line 190: I think this is best done with a helper function and two invocations of that test function -- that will help us understand which condition is failing. Something like:

```r
confirm_existing_application_use <- function(server) {
    # innards of loop
}
test_that("can find existing application on server & use it", {
  confirm_existing_application_use("example.com")
})
test_that("can find existing application on shinyapps.io & use it", {
  confirm_existing_application_use("shinyapps.io")
})
```

@hadley - do you have different recommendations?

aronatkins (approved):

mbaynton (approved):
This tests out properly for me. Running interactively I am prompted whether to redeploy to the application that was found, running noninteractively it assumes that's what I want, which is consistent with prior behavior.

aronatkins (approved):

PR #937, opened 2023-07-27
prefer RSCONNECT_PACKRAT over rsconnect.packrat

### LINKED ISSUE #935: rsconnect.packrat does not apply to push-button deployment
From a project that does not use renv (a very simple Shiny application will suffice):

```r
options(rsconnect.packrat = TRUE)
```

Deploy the project using push-button deployment. The "Deploy" pane will show:

```
ℹ Capturing R dependencies with renv
```

Deploying the same project using the console:

```r
reconnect::deployApp()
```

we see the output:

```
ℹ Capturing R dependencies with packrat
```

It appears that not all R options make their way to the deploy R session.

### LINKED ISSUE #936: rsconnect.packrat does not bypass renv.lock use
When deploying from an renv project, the `rsconnect.packrat` option does not bypass use of the `renv.lock`.

```r
options(rsconnect.packrat = TRUE)
rsconnect::deployApp()
#> ...
#> ℹ Capturing R dependencies from renv.lock
```

The `rsconnect.packrat` option is used as a feature-switch only when the deployment does not include an `renv.lock`.

aronatkins:
Environment variables are visible in the RStudio IDE R session used for push-button deployment. Not all R options are mirrored into that session.

These settings indicate that packrat should be used, even in the presence of an renv.lock. When using packrat, the renv directory and renv.lock file are removed from the bundle directory, as would happen prior to rsconnect 1.0.0.

Fixes #935
Fixes #936

```r
# Unset any existing options before testing.
Sys.unset("RSCONNECT_PACKRAT")
options(reconnect.packrat = NULL)

# Use packrat; applies to push-button and console deploys.
# Ignores renv.lock if it exists.
Sys.setenv(RSCONNECT_PACKRAT = "TRUE")

# Use packrat; applies to console deploys only. Discouraged.
# Ignores renv.lock if it exists.
options(rsconnect.packrat = TRUE)

# If both are set, the environment variable setting wins.
# This example will use renv because RSCONNECT_PACKRAT is set.
Sys.setenv(RSCONNECT_PACKRAT = "FALSE")
options(rsconnect.packrat = TRUE)
```

### DIFF
--- a/R/bundlePackage.R
+++ b/R/bundlePackage.R
@@ -29,12 +29,30 @@ bundlePackages <- function(bundleDir,
   packages_list
 }
 
+usePackrat <- function() {
+  # Use RSCONNECT_PACKRAT when it has any value; fall-back to rsconnect.packrat when the environment
+  # variable is unset.
+  env_value <- Sys.getenv("RSCONNECT_PACKRAT", unset = NA)
+  if (is.na(env_value)) {
+    return(truthy(getOption("rsconnect.packrat", default = FALSE)))
+  }
+  
+  return(truthy(env_value))
+}
+
 computePackageDependencies <- function(bundleDir,
                                        extraPackages = character(),
                                        quiet = FALSE,
                                        verbose = FALSE) {
 
-  if (file.exists(renvLockFile(bundleDir))) {
+  if (usePackrat()) {
+    taskStart(quiet, "Capturing R dependencies with packrat")
+    # Remove renv.lock so the packrat call to renv::dependencies does not report an implicit renv
+    # dependency. Mirrors rsconnect before 1.0.0, which did not include renv.lock in bundles.
+    # https://github.com/rstudio/rsconnect/blob/v0.8.29/R/bundle.R#L96
+    removeRenv(bundleDir)
+    deps <- snapshotPackratDependencies(bundleDir, extraPackages, verbose = verbose)
+  } else if (file.exists(renvLockFile(bundleDir))) {
     # This ignores extraPackages; if you're using a lockfile it's your
     # responsibility to install any other packages you need
     taskStart(quiet, "Capturing R dependencies from renv.lock")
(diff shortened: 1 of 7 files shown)


R/bundlePackage.R
```diff
@@ -29,26 +29,41 @@ bundlePackages <- function(bundleDir,
...
+usePackrat <- function() {
+  # Use RSCONNECT_PACKRAT when it has any value; fall-back to rsconnect.packrat when the environment
+  # variable is unset.
+  env_value <- Sys.getenv("RSCONNECT_PACKRAT", unset = NA)
+  if (is.na(env_value)) {
+    return(truthy(getOption("rsconnect.packrat", default = FALSE)))
+  }
+
+  return(truthy(env_value))
+}
```
hadley, line 41:
```suggestion
  value <- Sys.getenv("RSCONNECT_PACKRAT", unset = NA)
  if (is.na(value)) {
    value <- getOption("rsconnect.packrat", default = FALSE)
  }

  truthy(value)
}
```

That makes the logic a bit more clear to me.

R/utils.R
```diff
@@ -161,3 +161,16 @@ toJSON <- function(x, ...) {
...
+    default
+  else if (is.character(value))
+    value %in% c("TRUE", "True", "true", "T", "1")
+  else if (is.symbol(value))
+    as.character(value) %in% c("TRUE", "True", "true", "T", "1")
+  else if (is.na(value))
+    default
+  else
+    as.logical(value)
+}
```
hadley, line 176:
```suggestion
truthy <- function(value, default = FALSE) {
  if (!is.atomic(value) || length(value) != 1 || is.na(value))
    default
  else if (is.character(value))
    value %in% c("TRUE", "True", "true", "T", "1")
  else
    as.logical(value)
}
```

From https://github.com/rstudio/renv/pull/1609

tests/testthat/test-bundlePackage.R
```diff
@@ -1,18 +1,41 @@
...
+  expect_named(pkgs, c("foreign", "MASS"), ignore.order = TRUE)
+  expect_named(pkgs$foreign, c("Source", "Repository", "description"))
   expect_named(pkgs$MASS, c("Source", "Repository", "description"))
 
   # No renv lockfile left behind
   expect_equal(list.files(app_dir), "foo.R")
 })
 
-test_that("can snapshot deps with packrat", {
+test_that("can snapshot deps with packrat (option)", {
```
hadley, line 15: IMO you don't need this test because you test `usePackrat()` below

hadley (approved):

tests/testthat/test-bundlePackage.R
aronatkins: I'm going to keep both variations because `usePackrat` is an implementation detail and the setting/env var are incoming state.. and this is a reasonable point in the stack to run these functional tests. The `usePackrat` tests are used to check other state permutations; we won't re-check all those situations with renv/packrat analysis.

PR #949, opened 2023-08-11
app mode inferred only using top-level files

### LINKED ISSUE #942: Deployment fails because bundle is wrongly identified as API by `inferAppMode()`
## Context 

We are using a golem-style app-layout and an additional plumber API in `inst/plumber/api/plumber.R`, i.e.

```
.
├── app.R
├── DESCRIPTION
├── inst
│   └── plumber
|         └── api
|              └── plumber.R  <----  The culprit
├── R
│   ├── server.R
│   └── ui.R
...

```

## Undesired Behavior

`rsconnect:::inferAppMode()` infers that the project has `appmode: api`, not `appmode: shiny`. This function is called when using `rsconnect::writeManifest()`, `rsconnect::bundleApp()` (which is called by `rsconnect::deployApp()`), as well as when using the deploy button in the RStudio-Interface. Thus, `rsconnect::deployApp()` and the deployButton will wrongly try to deploy the project as an API instead of as a shiny app.

For example: 

```r
> rsconnect::writeManifest()
...
> jsonlite::read_json("manifest.json")$metadata$appmode
# [1] "api"
```

Creating a `manifest.json` in the directory and simply changing `appmode` to `shiny` in that file doesn't solve the issue,  because `deployApp()` calls `bundleApp()`, [which  generates its own `manifest.json` straight in a temporary bundle directory](https://github.com/rstudio/rsconnect/blob/2f07fde4c8a2bf96cb89d07eaf4c8780e372a12d/R/deployApp.R#L600), so the `manifest.json` file in the project directory is ignored.

## The root of the Problem 

`rsconnect:::inferAppMode()` checks for evidence of a plumber API (and returns) before it checks for evidence of a Shiny App [(as can be seen here)](https://github.com/rstudio/rsconnect/blob/2f07fde4c8a2bf96cb89d07eaf4c8780e372a12d/R/appMetadata.R#L132).

The behavior was introduced in rsconnect version 1.0.0, because 

- Prior to 1.0.0, `inferAppMode()` was called with a non-recursive list of the files in the directory, which was obtained via [`bundleFiles()`](https://github.com/rstudio/rsconnect/blob/9ce427edf03b6d243a4074264dbc950242b0a295/R/deployApp.R#L263C27-L263C27). Thus, only the top level was searched for evidence of the different app modes and `inst/api/plumber.R` would not have been part of the list. 
- Since 1.0.0, the function gets a recursive list of all files in the bundle, which is obtained via `listDeploymentFiles()`. Thus, in our case, the list now includes the api file in one of the subdirectories 'inst/api/plumber.R`
- This is due to a change in the internal function `listBundleFIles()`, which now calls a function that returns a recursive list of files ([old version in 0.8.29](https://github.com/rstudio/rsconnect/blob/9ce427edf03b6d243a4074264dbc950242b0a295/R/bundle.R#L204) vs. [new version in 1.0.0](https://github.com/rstudio/rsconnect/blob/1dc159ae02792dcc63c00a4b2669d2e156020bbc/R/bundleFiles.R#L108))
- This is despite a [comment stating that it only checks files in the root dir](https://github.com/rstudio/rsconnect/blob/db2bcba0d82698aca6bc6aff4f639cc531060b4b/R/appMetadata.R#L119)

## Workaround: setting `appPrimaryDoc = "app.R"` 

In the community forums, [someone suggested to use the argument `appPrimaryDoc = "app.R"`](https://community.rstudio.com/t/rsconnect-deployapp-bundle-wrongly-identified-as-api/170863/2?u=teebusch).

Unfortunately, this  throws an error:

```r
> rsconnect::writeManifest(appPrimaryDoc = "app.R")`

Error in `checkAppLayout()`:
! Project must not contain both app.R and a single-file Shiny app.
```

This error is thrown by a conditional in `rsconnect::checkAppLayout(appDir, appPrimaryDoc)` [(here)](https://github.com/rstudio/rsconnect/blob/2f07fde4c8a2bf96cb89d07eaf4c8780e372a12d/R/appMetadata.R#L88), a function that checks  whether the app directory has any of  the expected  layouts. This function is called by both, `writeManifest()` and `bundleApp()` via `rsconnect:::appMetadata()` and  it explicitly prevents the user from setting `appPrimaryDoc` to be `app.R`: 

```r
appFilesBase <- tolower(list.files(appDir))
...
primaryIsRScript <- identical(tolower(tools::file_ext(appPrimaryDoc)), "r")
if (primaryIsRScript && "app.r" %in% appFilesBase) {
  cli::cli_abort("Project must not contain both {.file app.R} and a single-file Shiny app.")
}
```

A solution is to add an additional condition to test, whether `appPrimaryDoc == app.R`. Ideally, this would have to be fixed in `{rsconnect}` itself, but one can temporarily inject a fix like this: 

```r
checkAppLayout_modified <- function (appDir, appPrimaryDoc = NULL) 
{
  ...
  if (primaryIsRScript && "app.r" %in% appFilesBase && tolower(appPrimaryDoc) != "app.r") {
    cli::cli_abort("Project must not contain both {.file app.R} and a single-file Shiny app.")
  }
  ...
}

assignInNamespace("checkAppLayout", checkAppLayout_modified, ns = "rsconnect")
rsconnect::writeManifest(appPrimaryDoc = "app.R")
```
(see below for full functional code)

This allows me to publish to rsconnect succesfully, since  the  error isn't thrown anymore and the generated `manifest.json` now has `appmode: shiny`.  However, the solution requires me to modify a function in the `rsconnect` namespace and it still doesn't allow me to use the deploy Button in RStudio, because there I cannot supply the `appPrimaryDoc` argument. (Also, this solution won't work if I want to set `appPrimaryDoc = "plumber.R"` to publish the API.)

It looks like currently there is no straightforward solution to use the project structure we have with the latest version of rsconnect. 

## Conclusion

Generally, I think, the preference of [an API over an App by `deployApp()`](https://github.com/rstudio/rsconnect/blob/2f07fde4c8a2bf96cb89d07eaf4c8780e372a12d/R/appMetadata.R#L132) is unexpected and should be removed. If I wanted to deploy the API that's in the package, I'd expect to use `deployApi()` to do so.

Similarly, when clicking the deploy Button in `app.R` I'd expect it to deploy the app that's in that file, whereas clicking the deploy button in `plumber.R` should deploy the API. 

It might be a solution to revert to the old behavior of searching only the root directory to infer the app mode. Another option would be to use the user-supplied "manifest.json" in `deployApp()`.

---

## Complete functional code of my 'workaround'

```r
# A solution is to prevent the error by checking whether `tolower(appPrimaryDoc) != "app.r"`

checkAppLayout_original <- rsconnect:::checkAppLayout  # for restoring it later

checkAppLayout_modified <- function (appDir, appPrimaryDoc = NULL) 
{
  cli::cli_alert_warning("Using custom function injected into the rsconnect namespace.")
  appFilesBase <- tolower(list.files(appDir))
  wwwFiles <- tolower(list.files(file.path(appDir, "www/")))
  primaryIsRScript <- identical(tolower(tools::file_ext(appPrimaryDoc)), 
                                "r")
  
  if (primaryIsRScript && "app.r" %in% appFilesBase && tolower(appPrimaryDoc) != "app.r") {
    cli::cli_abort("Project must not contain both {.file app.R} and a single-file Shiny app.")
  }
  satisfiedLayouts <- c(shinyAndUi = all(c("server.r", "ui.r") %in% 
                                           appFilesBase), shinyAndIndex = "server.r" %in% appFilesBase && 
                          "index.html" %in% wwwFiles, app = primaryIsRScript || 
                          any("app.r" %in% appFilesBase), Rmd = any(grepl(glob2rx("*.rmd"), 
                                                                          appFilesBase)), Qmd = any(grepl(glob2rx("*.qmd"), appFilesBase)), 
                        static = any(grepl("(?:html?|pdf)$", appFilesBase)), 
                        plumber = any(c("entrypoint.r", "plumber.r") %in% appFilesBase))
  if (any(satisfiedLayouts)) {
    return()
  }
  cli::cli_abort(c("Cancelling deployment: invalid project layout.", 
                   i = "Expecting one of the following publication types:", 
                   ` ` = "1. A Shiny app with `app.R` or `server.R` + `ui.R`", 
                   ` ` = "2. R Markdown (`.Rmd`) or Quarto (`.qmd`) documents.", 
                   ` ` = "3. A website containing `.html` and/or `.pdf` files.", 
                   ` ` = "4. A plumber API with `plumber.R` or `entrypoint.R`."))
}

assignInNamespace("checkAppLayout", checkAppLayout_modified, ns = "rsconnect")

rsconnect::writeManifest(appPrimaryDoc = "app.R")

appmode <- jsonlite::read_json("manifest.json")$metadata$appmode
if (appmode == "shiny") {
  rsconnect::deployApp(appPrimaryDoc = "app.R")
} else {
  cli::cli_abort("Something went wrong, `appmode` is `{ appmode }`.")
}
```

aronatkins:
Update: Rerooted on #950 to have its CI adjustments.

fixes #942

### DIFF
--- a/tests/testthat/test-appMetadata.R
+++ b/tests/testthat/test-appMetadata.R
@@ -33,6 +33,13 @@ test_that("handles special case of appPrimaryDoc as R file", {
   expect_equal(metadata$appMode, "shiny")
 })
 
+# https://github.com/rstudio/rsconnect/issues/942
+test_that("files beneath the root are not ignored when determining app-mode", {
+  dir <- local_temp_app(list("app.R" = "", "plumber/api/plumber.R" = ""))
+  metadata <- appMetadata(dir)
+  expect_equal(metadata$appMode, "shiny")
+})
+
 # checkLayout -------------------------------------------------------------
 
 test_that("checkLayout() errors if primary doc & app.R", {
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -46,7 +46,7 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: any::rcmdcheck
+          extra-packages: any::rcmdcheck, Biobase=?ignore, BiocManager=?ignore
           needs: check
 
       - uses: r-lib/actions/check-r-package@v2
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,8 @@
 * The `renv.lock` is ignored when the `RSCONNECT_PACKRAT` environment variable
   or the `rsconnect.packrat` option is set. (#936)
 
+* The content type is inferred by analyzing the set of top-level files. (#942)
+
 # rsconnect 1.0.1
 
 * `deployDoc()` includes `.Rprofile`, `requirements.txt` and `renv.lock` when
--- a/R/appMetadata.R
+++ b/R/appMetadata.R
@@ -36,7 +36,7 @@ appMetadata <- function(appDir,
     # Inference only uses top-level files
     rootFiles <- appFiles[dirname(appFiles) == "."]
     appMode <- inferAppMode(
-      file.path(appDir, appFiles),
+      file.path(appDir, rootFiles),
       usesQuarto = quarto,
       isShinyappsServer = isShinyappsServer
     )
@@ -117,7 +117,7 @@ checkAppLayout <- function(appDir, appPrimaryDoc = NULL) {
 }
 
 # infer the mode of the application from files in the root dir
-inferAppMode <- function(absoluteAppFiles,
+inferAppMode <- function(absoluteRootFiles,
                          usesQuarto = NA,
                          isShinyappsServer = FALSE) {
 
@@ -127,24 +127,24 @@ inferAppMode <- function(absoluteAppFiles,
   }
 
   # plumber API
-  plumberFiles <- matchingNames(absoluteAppFiles, "^(plumber|entrypoint).r$")
+  plumberFiles <- matchingNames(absoluteRootFiles, "^(plumber|entrypoint).r$")
   if (length(plumberFiles) > 0) {
     return("api")
   }
 
   # Shiny application using single-file app.R style.
-  appR <- matchingNames(absoluteAppFiles, "^app.r$")
+  appR <- matchingNames(absoluteRootFiles, "^app.r$")
   if (length(appR) > 0) {
     return("shiny")
   }
 
-  rmdFiles <- matchingNames(absoluteAppFiles, "\\.rmd$")
-  qmdFiles <- matchingNames(absoluteAppFiles, "\\.qmd$")
+  rmdFiles <- matchingNames(absoluteRootFiles, "\\.rmd$")
+  qmdFiles <- matchingNames(absoluteRootFiles, "\\.qmd$")
 
   if (is.na(usesQuarto)) {
     # Can't use _quarto.yml alone because it causes deployment failures for
     # static content: https://github.com/rstudio/rstudio/issues/11444
-    quartoYml <- matchingNames(absoluteAppFiles, "^_quarto.y(a)?ml$")
+    quartoYml <- matchingNames(absoluteRootFiles, "^_quarto.y(a)?ml$")
 
     usesQuarto <- length(qmdFiles) > 0 ||
       (length(quartoYml) > 0 && length(rmdFiles > 0))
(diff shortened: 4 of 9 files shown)


tests/testthat/test-appMetadata.R
```diff
@@ -33,6 +33,13 @@ test_that("handles special case of appPrimaryDoc as R file", {
   expect_equal(metadata$appMode, "shiny")
 })
 
+# https://github.com/rstudio/rsconnect/issues/942
+test_that("files beneath the root are not ignored when determining app-mode", {
+  dir <- local_temp_app(list("app.R" = "", "plumber/api/plumber.R" = ""))
+  metadata <- appMetadata(dir)
+  expect_equal(metadata$appMode, "shiny")
```
aronatkins, line 40: This test incorrectly produced "api" before this fix.

aronatkins:
Test failures from https://github.com/rstudio/rsconnect/pull/950; will update once that merges down.

tests/testthat/test-appMetadata.R
```diff
@@ -33,6 +33,13 @@ test_that("handles special case of appPrimaryDoc as R file", {
   expect_equal(metadata$appMode, "shiny")
 })
 
+# https://github.com/rstudio/rsconnect/issues/942
+test_that("files beneath the root are not ignored when determining app-mode", {
```
hadley, line 37:
```suggestion
test_that("files beneath the root are ignored when determining app-mode", {
```

hadley (approved):

Issue #954, opened 2023-08-16
Failure ('test-linters.R:17:3'): The linter identifies files not matching in case sensitivity --


tillea:
Hi,
when testing the Debian package of rsconnect I get:

    == Failed tests ================================================================
    -- Failure ('test-linters.R:17:3'): The linter identifies files not matching in case sensitivity --
    as.integer(filepath.capitalization$indices) (`actual`) not equal to 31 (`expected`).
      `actual`:
    `expected`: 31
    [ FAIL 1 | WARN 1 | SKIP 116 | PASS 548 ]

You can see a full build log in the [CI of Debian Gitlab instance](https://salsa.debian.org/r-pkg-team/r-cran-rsconnect/-/jobs/4562989#L550)

Kind regards, Andreas.

aronatkins:
The last passing build appears to be have from rsconnect-0.8.29: https://salsa.debian.org/r-pkg-team/r-cran-rsconnect/-/jobs/3806435

aronatkins:
> I guess this is not a safe point to start with.

Understood - there were linting changes in the 1.0.0 rsconnect release. I know the PRs and commits that included those changes. I mostly wanted to record that we were previously in a "passing" state.

Most likely, it was this work (https://github.com/rstudio/rsconnect/pull/695).

Is there a `Dockerfile` or something similar that I could use to recreate this failure?

tillea:
Am Wed, Aug 16, 2023 at 07:02:49AM -0700 schrieb Aron Atkins:
> Understood - there were linting changes in the 1.0.0 rsconnect release. I know the PRs and commits that included those changes. I mostly wanted to record that we were previously in a "passing" state.

Understood,
 
> Is there a `Dockerfile` or something similar that I could use to recreate this failure?

I'm pretty sure there are some docker files of Debian unstable (this is what you need to seek for) out there.  We are running [this script](https://salsa.debian.org/r-pkg-team/r-cran-rsconnect/-/blob/master/debian/tests/run-unit-test) for testing.  (Note: You probably should skip the copying stuff and just do

        LC_ALL=C R --no-save < $testfile

aronatkins:
The copy and gzip lines:

```
cp -a /usr/share/doc/$debname/tests/* $AUTOPKGTEST_TMP
gunzip testthat/test*.gz # do not unzip everything - some files need to remain compressed
```

[...]

PR #956, opened 2023-08-16
avoid zero-length files

aronatkins:
Debian discards zero-length files. We may add additional zero-length files within our test hierarchy in the future, but avoid them for now.

See #954

@tillea - this will help avoid your immediate problem, but the test hierarchy really should have its file and directory structure preserved.

### DIFF
--- a/tests/testthat/quarto-website-r-py/requirements.txt
+++ b/tests/testthat/quarto-website-r-py/requirements.txt
@@ -1,0 +1,2 @@
+# stub requirements.txt
+# Its presence indicates the need for Python dependencies.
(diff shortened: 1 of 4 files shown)


aronatkins:
The set of files was discovered by:

```
find tests -size 0
#> tests/testthat/quarto-website-r-py/requirements.txt
#> tests/testthat/shiny-app-in-subdir/my-app/ui.r
#> tests/testthat/shiny-app-in-subdir/my-app/server.R
#> tests/testthat/shinyapp-with-absolute-paths/subdir/Genetics.txt
#> tests/testthat/shinyapp-with-absolute-paths/data/College.txt
```

PR #969, opened 2023-08-22
renv dependency limit mirrors the rsconnect bundle limit

### LINKED ISSUE #968: error reporting large directories
Set-up:

```r
# create a new project
usethis::create_project("~/Desktop/many-files")
# within the project, create a shiny application
writeLines(c(
        "shinyApp(ui = fluidPage('so many files.'), server = function(input, output){})"
    ), "app.R")
# create many (tiny) files
dir.create("so-many-files")
for (each in seq_len(1001)) { writeLines(character(0), file.path("so-many-files", paste0(each, ".txt")))}
```

Error:

``` r
rsconnect::writeManifest()
#> ℹ Capturing R dependencies with renv
#> Error in dirname(root): a character vector argument expected
```

<sup>Created on 2023-08-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

aronatkins:
additionally, do not let renv infer the root.

fixes #968

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # rsconnect (development version)
 
+* Fixed analysis of directories that were smaller than the
+  `rsconnect.max.bundle.files=10000` limit but larger than the
+  `renv.config.dependencies.limit=1000` limit. (#968)
+
 # rsconnect 1.0.2
 
 * Fixed redeployments to shinyapps.io where `appName` is provided, but no local
--- a/R/bundlePackageRenv.R
+++ b/R/bundlePackageRenv.R
@@ -9,9 +9,18 @@ snapshotRenvDependencies <- function(bundleDir,
   )
   defer(options(old))
 
+  dependenciesLimit <- getOption("renv.config.dependencies.limit")
+  if (is.null(dependenciesLimit)) {
+    maxFiles <- getOption("rsconnect.max.bundle.files", 10000)
+    oldlim <- options(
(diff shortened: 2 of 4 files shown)


kevinushey (approved):
LGTM!

aronatkins:
Note: We probably uncovered the same `root = NULL` problem in packrat and worked around it by providing the project path as the root directory.
https://github.com/rstudio/packrat/blob/main/R/dependencies.R#L182

aronatkins:
@kevinushey - Any theories about why the Windows release tests are failing? It appears to be enumerating files, which isn't present on the other OSes.

kevinushey:
Maybe we need `progress = FALSE` in the call to `dependencies()`?

kevinushey:
For reference, `renv` has a progress reporter that only prints progress if the associated action takes a "long" time (more than 1 second). It's started here:

https://github.com/rstudio/renv/blob/ac7126d7b2856a39012d6079689861b18c07c5b7/R/dependencies.R#L477-L481

And see the progress helpers here:

https://github.com/rstudio/renv/blob/ac7126d7b2856a39012d6079689861b18c07c5b7/R/progress.R#L2-L32

aronatkins:
Thanks. It's probably `progress`. There is no option to control that in CI-only, correct?

aronatkins:
Also: We made the same `progress = FALSE` change in the packrat call to `renv::dependencies()`... same lessons. :)

PR #973, opened 2023-08-23
ignore .env virtual environments; allow .env files

### LINKED ISSUE #972: My .env file is not visible when deploying app on shiny server
I could not deploy my app on shiny server:
 - please update to the latest available on CRAN
 so I updated rsconnect package and now my .env file is not visible: It is no longer among the files I can select

aronatkins:
fixes #972


aronatkins:
related: https://github.com/rstudio/rsconnect-python/pull/207
note: we may still need to check for venv on Windows differently.

kevinushey (approved):
LGTM!

Issue #974, opened 2023-08-23
Ensure manifest data is encoded as UTF-8


mslynch:
shinyapps.io and Posit Cloud presume `manifest.json` to be valid JSON encoded in UTF-8. However, some of the data in this file comes from `DESCRIPTION` files, which may not be encoded in UTF-8. If we see that `DESCRIPTION`'s `Encoding` field is not UTF-8, we should attempt to convert it.

hadley:
I think it's very rare for the `Encoding` to be non-UTF-8.

mslynch:
> I think it's very rare for the `Encoding` to be non-UTF-8.

That's my impression as well – and it should be ever rarer still for `DESCRIPTION` to also contain characters that are only valid in latin-1. But we had a customer try to use a package like this (`relaimpo`).

hadley:
And particularly annoying in that case is that the only field that is non-ASCII is `Packaged`, which is definitely not very important.

tyatposit:
validated.  worked fine for me install/call `relaimpo` and deploy to both cloud and shinyapps which sounds like was a package with a DESCRIPTION encoded as UTF-8.

- created new shinyapp
- `install.packages('relaimpo')`
- called `library(relaimpo)`
- deployed the app to both shinyapps and cloud
- app built and loaded up without issue

PR #979, opened 2023-08-24
handle non-utf8 description encodings

mslynch:
Resolves https://github.com/rstudio/rsconnect/issues/974.

### DIFF
--- a/R/bundlePackage.R
+++ b/R/bundlePackage.R
@@ -115,5 +115,5 @@ availablePackages <- function(repos) {
 
 package_record <- function(name, lib_dir = NULL) {
   record <- packageDescription(name, lib.loc = lib_dir, encoding = "UTF-8")
-  lapply(record, identity)
+  unclass(record)
 }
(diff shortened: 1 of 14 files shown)


hadley (approved):
Probably worth a small unit test?

mslynch:
@hadley looks like this breaks on `oldrel-4`. Any suggestions for a path forward?

kevinushey:
Can you just use `packageDescription()`? That tries to handle Encoding internally in a similar way.

mslynch:
`packageDescription()` works beautifully, thanks!

R/bundlePackage.R
```diff
@@ -114,6 +114,6 @@ availablePackages <- function(repos) {
 }
 
 package_record <- function(name, lib_dir = NULL) {
-  path <- system.file("DESCRIPTION", package = name, lib.loc = lib_dir)
-  as.list(as.data.frame(read.dcf(path)))
+  record <- packageDescription(name, lib.loc = lib_dir, encoding = "UTF-8")
+  lapply(record, identity)
```
hadley, line 118: Could that just be `as.list(record)`?

mslynch: That gives the error `Error: No method asJSON S3 class: packageDescription`. Not sure if there's a better way around this.

Issue #961, opened 2023-08-17
Add space parameter to deploy*() functions for specifying a destination when publishing to Posit Cloud


kippandrew:
Posit Cloud supports publishing content to any space within the target account. However, because we don't provide a way to target a space when publishing via the command line, content is published in the user's default space called "Your Workspace". We would like to allow users to optionally specify a space when deploying content to Posit Cloud. The space parameter could accept a space by name or ID. It may be useful to have a function to let users discover spaces they have access to.

cc @hadley @aronatkins @omar-rs @mslynch @Alex-Chisholm

PR #983, opened 2023-08-28
add space parameter for Posit Cloud

mslynch:
Resolves https://github.com/rstudio/rsconnect/issues/961.

cc @scubastevew


R/client-cloud.R
```diff
@@ -161,23 +161,27 @@ cloudClient <- function(service, authInfo) {
...
-      # in case the source cloud project is a temporary copy, there is no
-      # content id. The output will be published without a space id.
-      if (!is.null(currentProjectId)) {
-        json$project <- currentProjectId
+      if (is.null(spaceId)) {
+        currentProjectId <- getCurrentProjectId(service, authInfo)
+        # in case the source cloud project is a temporary copy, there is no
+        # content id. The output will be published without a space id.
+        if (!is.null(currentProjectId)) {
+          json$project <- currentProjectId
```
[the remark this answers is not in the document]
mslynch: > This change does not add the `json$project` when provided an incoming `spaceId`. Is it needed in this case?

Thanks, it is needed here – pushed a commit to change this logic. If we're executing inside a Posit Cloud project, we always want to link the output to that output through `project`. For the space, we want to update it if there's a specific space passed in through `deployApp`. If there's no space when we create the application, it should be set to the space of the project.

R/client-cloud.R
```diff
@@ -230,12 +234,21 @@ cloudClient <- function(service, authInfo) {
         revision$application_id
     },
 
-    deployApplication = function(application, bundleId = NULL) {
+    deployApplication = function(application, bundleId = NULL, spaceId = NULL) {
+      outputPatchData <- list()
+
       currentProjectId <- getCurrentProjectId(service, authInfo)
       if (!is.null(currentProjectId)) {
+        outputPatchData$project <- currentProjectId
```
aronatkins, line 242: This is different logic than is used when creating; intentional? In this case, we could end up using `currentProjectId` from the environment and the `spaceId` argument. When creating, the `spaceId` argument is only used when there is no `currentProjectId`.

tests/testthat/test-client-cloud.R
```diff
@@ -303,6 +303,50 @@ test_that("Create application", {
...
+        application_id <- strtoi(substr(methodAndPath, match[2], end))
+
+        list(
+          "id" = application_id,
+          "content_id" = 1
+        )
+      })
+  ))
+
+  restoreOpt <- options(rsconnect.http = mockServer$impl)
```
aronatkins, line 333: You should be able to:
```r
withr::local_options(rsconnect.http = mockServer$impl)
```

That should replace the `options(); defer()` two-step.

aronatkins (approved):

R/client-cloud.R
mslynch: I think the logic as-written makes sense (at least to me). When you deploy, it should definitely set the space based on the space of the environment project. But on redeploys, it would probably be more confusing for projects to move between spaces without providing the argument.

@m-- Any other thoughts?

kippandrew: I think that makes sense to me.

mslynch:
> The https://github.com/rstudio/rsconnect/issues/961 issue also suggests creating a function to list the available spaces; is that arriving, as well?

Probably, likely in another PR.

Issue #988, opened 2023-08-31
Set content_category on static outputs and revisions


omar-rs:
Creating static outputs and revisions in Posit Cloud supports setting `content_category` in the API payload.

PR #993, opened 2023-09-05
set content category for outputs and revisions

mslynch:
Resolves https://github.com/rstudio/rsconnect/issues/988.

### DIFF
--- a/R/client-cloud.R
+++ b/R/client-cloud.R
@@ -161,8 +161,13 @@ cloudClient <- function(service, authInfo) {
       GET(service, authInfo, path, query)
     },
 
+<<<<<<< Updated upstream
     createApplication = function(name, title, template, accountId, appMode, contentCategory = NULL) {
       json <- list()
+=======
+    createApplication = function(name, title, template, accountId, appMode, contentCategory) {
+      json <- list(content_)
+>>>>>>> Stashed changes
       json$name <- name
       json$application_type <- if (appMode %in% c("rmd-static", "quarto-static", "static")) "static" else "connect"
       if (appMode %in% c("rmd-static", "quarto-static")) {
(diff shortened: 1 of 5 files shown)


aronatkins (commented):
Does this need any documentation or NEWS update?

mslynch:
> Does this need any documentation or NEWS update?

I think it's probably not notable enough. We use the content category from the bundle manifest already; this just gets that information to us sooner.

R/client-cloud.R
```diff
@@ -224,9 +228,10 @@ cloudClient <- function(service, authInfo) {
       )
     },
 
-    createRevision = function(application) {
+    createRevision = function(application, contentCategory) {
         path <- paste0("/outputs/", application$id, "/revisions")
-        revision <- POST_JSON(service, authInfo, path, data.frame())
+        json <- list(content_category = contentCategory)
+        revision <- POST_JSON(service, authInfo, path, contentCategory)
```
hadley, line 234: Shouldn't line 234 use the `json` object you just created?

R/client-cloud.R
```diff
@@ -180,6 +180,10 @@ cloudClient <- function(service, authInfo) {
         json$space <- currentProject$space_id
       }
 
+      if (!is.null(contentCategory)) {
```
hadley, line 183: FWIW I don't think this check is strictly necessary since `contentCategory` is `NULL`, assigning it into a list won't do anything.

R/client-cloud.R
```diff
@@ -161,7 +161,7 @@ cloudClient <- function(service, authInfo) {
       GET(service, authInfo, path, query)
     },
 
-    createApplication = function(name, title, template, accountId, appMode) {
+    createApplication = function(name, title, template, accountId, appMode, contentCategory) {
```
hadley, line 164:
```suggestion
    createApplication = function(name, title, template, accountId, appMode, contentCategory = NULL) {
```

? (Would simplify a couple of the tests)

hadley (approved):

PR #1006, opened 2023-09-27
handle unknown and Local package source

aronatkins:
This shifts from using a local "Source" with whatever "Repository" comes from renv to using NA for both. An error now occurs because the package did not come from a well-known repository.

fixes #1004

```r
# Create a new project and launch a separate RStudio session.
usethis::create_project("~/Desktop/mixed-content")

# In that new project ...
unlink("R", recursive = TRUE)
writeLines(c(
    "library(shiny)",
    "shinyApp('single file Shiny deploy', function(input, output) {})"
), "app.R")

renv::init()
renv::install(c("rsconnect"), prompt = FALSE)
renv::install("~/Downloads/shiny_1.7.5.tar.gz", repos = NULL, type = "source", prompt = FALSE)
renv::snapshot(prompt = FALSE)

rsconnect::writeManifest()
jsonlite::fromJSON("manifest.json")$packages$shiny[c("Source","Repository")]
#> $Source
#> [1] "Local"
#>
#> $Repository
#> [1] "CRAN"

renv::install("rstudio/rsconnect@aron-issue-1004", prompt = FALSE)

# restart R session ...

rsconnect::writeManifest()
#> ℹ Capturing R dependencies from renv.lock
#> ✔ Found 32 dependencies
#> Error in `createAppManifest()` at rstudio-rsconnect-8856765/R/writeManifest.R:61:2:
#> ! All packages must be installed from a reproducible location.
#> ✖ Can't re-install packages installed from source: shiny.
#> ℹ See `rsconnect::appDependencies()` for more details.
#> Run `rlang::last_trace()` to see where the error occurred.
```

Issue #1004, opened 2023-09-26
Changes in renv/rsconnect - now getting "Local" packages in our CI/CD pipeline


slodge-work:
Not entirely sure this is an issue - but this seems like the best place to get advice/answers - so please forgive me asking here.

For many of our apps, reports and apis, our repos and CI/CD pipelines work by:

- each repo contains a package and one or more apps
- the package is version 9.9.9 in dev/git 
- the build pipeline:
  - stamps an incrementing version number on top
  - uses devtools to build the package (just a zip)
  - uses devtools to install the local package from disk
  - uploads the package to either our test CRAN or to our real CRAN
  - runs rsconnect writeManifest and deploy steps to deploy to either our test Connect or to our prod Connect
 
This pipeline's been working pretty well for a couple of years and is very stable.

Right now I'm looking at a breaking change in our pipelines following upgrading rsconnect to 1.x. What I'm seeing is that the currently built package is being picked up in the auto-generated manifest.json as: 
```
    "R.Brandy": {
      "Source": "Local",
      "Repository": "OurCo.IMS",
      "description": {
        "Package": "R.Brandy",
        "Type": "Package",
        "Title": "Shiny support for OurCo look and feel including the Unified\nDesign System",
        "Authors@R": "...",
        "Description": "R package with functions exposing Shiny Unified Design.",
        "Version": "0.1.209",
        "Maintainer": "...",
        "License": "file LICENSE",
        "Encoding": "UTF-8",
        "Repository": "OurCo.IMS",
        "Imports": "htmltools, shiny, dplyr, rmarkdown, purrr, magrittr,\ndownloadthis, rlang, jsonlite, stringr, rstudioapi, ggplot2,\nbase64enc, tibble, readr",
        "RoxygenNote": "7.2.1",
        "NeedsCompilation": "no",
        "Packaged": "2023-09-26 15:18:59 UTC; azdagent_azpcontainer",
        "Author": "Support [aut, cre]",
        "Built": "R 4.2.0; ; 2023-09-26 15:19:03 UTC; unix",
        "RemoteType": "local",
        "RemoteUrl": "/__w/1/a/package/R.Brandy_0.1.209.tar.gz"
      }
    },
```

The problem here are the line:
```
      "Source": "Local",      
      "Repository": "OurCo.IMS",
```

These cause the Connect deployment step to fail with lots of messages like...

```
2023/09/26 15:21:19.810629517 curl: (6) Could not resolve host: OurCo.IMS
[...]

PR #1009, opened 2023-10-02
search repositories for unknown and local packages

aronatkins:
Split from #1006; we will probably not take this path.


aronatkins:
Closing per discussion in #1004; created this PR to record the idea.

PR #1010, opened 2023-10-02
tests against "real" projects are skipped on CRAN

aronatkins:
Avoid (revdep) test situations where a test project has a dependency that may be installed from source. We encountered this first with curl, but could occur with any package required by content (e.g. Shiny, Plumber).

The now-skipped tests were identified by installing shiny, plumber, rmarkdown, and https://github.com/jeroen/curl from source and then running tests. Each failing test (caused by renv snapshot failures related to from-source installs) was marked for CRAN skipping.

### DIFF
--- a/tests/testthat/test-appDependencies.R
+++ b/tests/testthat/test-appDependencies.R
@@ -1,4 +1,6 @@
 test_that("appDependencies includes implicit deps", {
+  skip_on_cran()
+
   withr::local_options(renv.verbose = TRUE)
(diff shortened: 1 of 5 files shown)


aronatkins:
I don't love that we are adding `skip_on_cran()` to more and more tests, but don't see an alternative.

aronatkins:
Chatted with @kevinushey - we're going to skip-tests for now.

Issue #1013, opened 2023-10-05
Unable to replace existing shinyapps.io already published


rasanderson:
If a shiny application has already been published on shinyapps.io it is no longer possible to replace it. Using version 1.0.1 of rsconnect a message box pops up in RStudio asking the user to confirm that the existing app will be replaced. After confirming, the following error is displayed:

```
Creating application on server...
Error in `POST()`:
<[https://api.shinyapps.io/v1/applications/]> failed with HTTP status
409
Application exists with name: expt_design
Backtrace:
1. └─rsconnect::deployApp(...)
2. └─client$createApplication(...)
3. └─rsconnect:::POST_JSON(service, authInfo, "/applications/", json)
4. └─rsconnect:::POST(...)
5. └─rsconnect:::httpRequestWithBody(...)
6. └─rsconnect:::handleResponse(httpResponse, error_call = error_call)
7. └─rsconnect (local) reportError(json$error)
8. └─cli::cli_abort(...)
9. └─rlang::abort(...)
Execution halted
```

After downgrading rsconnect to version 0.8.29 and retrying an error message is displayed on Ubuntu 20.04, but not Windows 10. In Windows 10 downgrading to 0.8.29 provides a successful workaround for the problem. But in Ubuntu 20.04 the system detects that rsconnect is out-of-date and refuses to process it. Error message in Ubuntu 20.04 is:

```
Preparing to deploy document...Error: HTTP 426
GET https://api.shinyapps.io/v1/applications/?filter=account_id:668772&filter=type:shiny&filter=name:expt_design&count=100&offset=0
The rsconnect package is out of date.
 Please update to the latest available on CRAN.
Execution halted
```

<details>
The current environment shown by sessionInfo() in Ubuntu 20.04 is below:

```
sessionInfo()
R version 4.1.2 (2021-11-01)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.6 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C               LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8    
[...]

PR #1015, opened 2023-10-10
Return stub deployments when matching records do not exist.

### LINKED ISSUE #985: showProperties() continue giving a ! Couldn't find any deployments matching supplied criteria
I opened an issue with this problem here: https://github.com/rstudio/rsconnect/issues/980. Thanks @aronatkins for investigate and fix an issue related with `showProperties()` here: https://github.com/rstudio/rsconnect/pull/984. I have installed the development version of `rsconnect` and I tried to use this function again, but unfortunately the same problem continue. ¿Someone with the same problem?, ¿any idea about the reason or way to fix it?
Thanks in advance.

### LINKED ISSUE #989: showLogs() does not run if it does not find the deployment
After the update to 1.0.2 I cannot run `rsconnect::showLogs()` on a different machine that the one I use to develop and deploy.
showLogs() says it doesn't find any deployment.
This was not a problem before the last version change. Before the change, only the account (token) and the app details was enough to run showLogs().

A patch to this problem is to copy the app folder from the machine I use to deploy into the machine I use to run showLogs().
It's not a very clean solution, but at least showLogs() now finds the deployment and works again.

It would be nice if showLogs() could run anywhere independently of the app/deployment folder, as it was the case before.

Thanks, bye.

aronatkins:
The stub deployment is subsequently used to look-up the application by name. This lets showLogs, configureApp, setProperty, and unsetProperty interact with shinyapps.io when the client lacks a deployment record.

Does not address issues with publishing, such as #1013.

Fixes #985
Fixes #989

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # rsconnect (development version)
 
+* `showLogs()`, `configureApp()`, `setProperty()`, and `unsetProperty()`
+  search for the application by name when there are no matching deployment
+  records. (#985, #989)
+
 # rsconnect 1.1.1
 
 * Added `space` parameter to deploy directly to a space in Posit Cloud.
--- a/R/applications.R
+++ b/R/applications.R
@@ -121,6 +121,19 @@ applications <- function(account = NULL, server = NULL) {
   return(res)
 }
 
+# Use the API to filter applications by name.
+getAppByName <- function(client, accountInfo, name) {
+  # NOTE: returns a list with 0 or 1 elements
+  app <- client$listApplications(accountInfo$accountId, filters = list(name = name))
+  if (length(app)) {
+    return(app[[1]])
+  }
+
+  stop("No application found. Specify the application's directory, name, ",
+       "and/or associated account.", call. = FALSE)
+}
+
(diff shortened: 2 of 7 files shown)


toph-allen (approved):
Checking my understanding: This works because the ShinyApps.io API works with just the info contained in the stub deployment.

aronatkins:
> Checking my understanding: This works because the ShinyApps.io API works with just the info contained in the stub deployment.

The `getAppByName` function only needs the server, account, and name to locate an application. That's enough to let these calls interact with shinyapps.io -- a full deployment record is unnecessary.

We likely need something similar for the re-deploy workflow, but that will probably need to create a full deployment record. The functions referenced in this PR were the only ones which used `findDeployment()`, which let me create a narrower solution.

toph-allen:
### Verification

Deployed a test app to ShinyApps.io and deleted the `rsconnect` subdirectory containing deployment records.

- On the previous version of `rsconnect` I'd see an error when running `rsconnect::showProperties()`.
- On the version installed from GitHub containing this fix, I'm prompted to pick an account, and when I pick my ShinyApps.io account, I see the app's properties.

PR #1018, opened 2023-10-13
serialize multi-value metadata entries as comma-separated values

### LINKED ISSUE #1017: quarto: engines metadata is vector, not singular
Quarto content can report multiple engines, which arrives as the `metadata` argument to `deployApp()` and others.

That metadata is eventually serialized into a deployment record. The presence of multiple engines causes rsconnect to write two records into a single DCF file.

``` r
record <- rsconnect:::deploymentRecord(
    name = "the-name",
    title = "the-title",
    username = "the-username",
    account = "the-account",
    server = "the-server",
    metadata = list(quarto_engines = c("knitr","markdown"))
)
t <- tempfile()
rsconnect:::writeDeploymentRecord(record, t)
read.dcf(t)
#>      name       title       username       account       server       hostUrl
#> [1,] "the-name" "the-title" "the-username" "the-account" "the-server" ""     
#> [2,] "the-name" "the-title" "the-username" "the-account" "the-server" ""     
#>      appId bundleId url version quarto_engines
#> [1,] ""    ""       ""  "1"     "knitr"       
#> [2,] ""    ""       ""  "1"     "markdown"
```

We serialize `envVars` to a comma-separated list, but take no similar action for other multi-value fields.

https://github.com/rstudio/rsconnect/blob/525289dd91cd806c64cf79233c643d2f6bf2409a/R/deployments.R#L73-L76

https://github.com/rstudio/rsconnect/blob/525289dd91cd806c64cf79233c643d2f6bf2409a/R/deployments.R#L141

aronatkins:
fixes #1017

``` r
record <- rsconnect:::deploymentRecord(
    name = "the-name",
    title = "the-title",
    username = "the-username",
    account = "the-account",
    server = "the-server",
    metadata = list(quarto_engines = c("knitr","markdown"))
)
t <- tempfile()
rsconnect:::writeDeploymentRecord(record, t)
read.dcf(t)
#>      name       title       username       account       server       hostUrl
#> [1,] "the-name" "the-title" "the-username" "the-account" "the-server" ""     
#>      appId bundleId url version quarto_engines   
#> [1,] ""    ""       ""  "1"     "knitr, markdown"
```

This change does not adjust deployment records reads.

### DIFF
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -145,6 +145,9 @@ deploymentRecord <- function(name,
     url = url %||% "",
     version = version
   )
+  # convert any multi-value metadata entries into comma-separated values
+  # this prevents write.dcf from writing multiple records into one file.
+  metadata <- lapply(metadata, function(v) paste0(v, collapse = ", "))
   c(standard, metadata)
 }
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,9 @@
 * `rpubsUpload()` correctly records the initial RPubs destination, allowing
   republishing. (#976)
 
+* `deployApp()` and friends record multi-value `metadata` entries as
+  comma-separated values. (#1017)
+
 # rsconnect 1.1.1
 
 * Added `space` parameter to deploy directly to a space in Posit Cloud.
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -112,6 +112,9 @@
 #' @param metadata Additional metadata fields to save with the deployment
 #'   record. These fields will be returned on subsequent calls to
 #'   [deployments()].
+#'
+#'   Multi-value fields are recorded as comma-separated values and returned in
+#'   that form. Custom value serialization is the responsibility of the caller.
 #' @param forceUpdate What should happen if there's no deployment record for
 #'   the app, but there's an app with the same name on the server? If `TRUE`,
 #'   will always update the previously-deployed app. If `FALSE`, will ask
--- a/man/deployApp.Rd
+++ b/man/deployApp.Rd
@@ -144,7 +144,10 @@ potentially problematic code?}
 
 \item{metadata}{Additional metadata fields to save with the deployment
 record. These fields will be returned on subsequent calls to
-\code{\link[=deployments]{deployments()}}.}
+\code{\link[=deployments]{deployments()}}.
+
+Multi-value fields are recorded as comma-separated values and returned in
+that form. Custom value serialization is the responsibility of the caller.}
 
 \item{forceUpdate}{What should happen if there's no deployment record for
 the app, but there's an app with the same name on the server? If \code{TRUE},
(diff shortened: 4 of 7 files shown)


R/deployments.R
```diff
@@ -145,6 +145,9 @@ deploymentRecord <- function(name,
     url = url %||% "",
     version = version
   )
+  # convert any multi-value metadata entries into comma-separated values
+  # this prevents write.dcf from writing multiple records into one file.
+  metadata <- lapply(metadata, function(v) paste0(v, collapse = ", "))
```
hadley, line 150: I don't see an equivalent change in the parsing code, so I assume that already existed?

hadley: Also can now simplify line 141?

aronatkins: I'm rewriting only `metadata` values, as those are not a well-specified part of the `rsconnect::deployApp()` contract. Additionally, I'm leaving the rewritten result as-is because I didn't want to try to grok the expectations of the caller.

I could be convinced otherwise (on both of your points).

hadley: Ooooh, got it.

It does feel asymmetric to me to handle writing but not reading. I think it would be more symmetric to either error if metadata contains anything other that single strings on write, or to also parse comma separated values on read.

aronatkins: I agree (symmetry is preferred), but the IDE and the Quarto R package already provide vectors of engines...

hadley: In that case, the current code seems like the least worst solution to me 😄

aronatkins: > In that case, the current code seems like the least worst solution to me 😄

@hadley - to confirm: the change in this PR is the least-worst solution, not the current `main` implementation, correct?

toph-allen (approved):
I see the changes, but don't have a strong feeling about them only because I don't know how what reads the `quarto_engines` fields from DCF files. Seems fine, but if we read the values (e.g. to compare them) we should probably deserialize properly or guarantee ordering? Maybe I'm overthinking, and perhaps we're just writing the values so that users can use them if they want.

aronatkins:
> I don't know how what reads the quarto_engines fields from DCF files.

This is the real challenge. The IDE passes some fields as `metadata` to use when reconstructing the UI on subsequent deployments. Things like additional and ignored files.

https://github.com/rstudio/rstudio/blob/c8e95aa32165dc189a4718a838f9fc49b7e6f821/src/cpp/session/modules/SessionRSConnect.cpp#L293-L301

The Quarto metadata is not used for that purpose; it modifies how rsconnect treats the deployment. Quarto details are not used by the IDE on subsequent deployments.

All other information provided in `metadata` is passed-through.

The Quarto details probably should have been provided as explicit arguments to the rsconnect functions, but that would have introduced other difficulties,  since that API change would have needed to propagate to CRAN, etc, etc. Reasons.

We could strip the Quarto fields before writing the deployment record, but it also feels like their presence is useful.

aronatkins:
One additional note that came up during a discussion: We always use the incoming `metadata` object when considering the Quarto information. We never use the metadata from the deployment record.

Issue #1019, opened 2023-10-16
Behavior when both appname and appid is specified?


soetang:
We have had problems understanding why our deployment pipelines sometimes created new apps even if the app already exists on Posit connect.  Turns out that if you specify both appName and appId it seems like the appId is ignored if the user is not the current owner of the "old" application. 

This seems to be the opposite of the documentation for appid (https://rstudio.github.io/rsconnect/reference/deployApp.html): 
´Use this to deploy to an exact known application, ignoring all existing deployment records and appName.´

Dont know if this a bug or the documentation needs an update. Ideally i would also like a warning if both are specified so that, we are made aware of the behavior chosen.

Removing appname as an parameter for deployApp solved the problem for us, but others might be dealing with the same issues.

I suspect the issue comes from the call to appDeployments in R/deploymentTarget.R.  Where providing the account for the accountFilter is the cause of the issue. (I am not familiar with the codebase so might be wrong here).
 
```
 appDeployments <- deployments(
    appPath = recordPath,
    nameFilter = appName,
    accountFilter = account,
    serverFilter = server
  )
```

soetang:
I would be happy to make a pull request if needed? Just want to clarify if I am correct first?

aronatkins:
The development version of rsconnect uses `appId` as the primary way to identify existing content; could you test that redeployment targets the correct application in your environment?

```r
remotes::install_github("rstudio/rsconnect")
```

Issue #1007, opened 2023-09-27
Collaborator source RMD publishing creates a new app instead of a new bundle


dbkegley:
rsconnect: `1.1.0`
Connect: `2023.10.0-dev+216-g70573d7869`

1. Add Admin and Publisher users as connections in the IDE
1. Deploy an RMD (with source) as the Admin from the IDE
1. Add the Publisher as a collaborator in the Connect dashboard
1. In the IDE, remove the Admin user connection
1. Re-deploy the same RMD (with source) as the Publisher from the IDE

<img width="589" alt="Screenshot 2023-09-27 at 12 12 25 PM" src="https://github.com/rstudio/rsconnect/assets/5727001/e7919131-f141-4af3-91c7-01b2c121f14c">

actual:
A new App is created (owned by the Publisher)

expected: 
A new bundle should be created for the existing App (owned by the Admin)

see also: ~https://docs.posit.co/tutorials/user/publishing-with-collaboration-on-rsc/~ (obsolete)
see also: https://docs.posit.co/connect/user/publishing/#publishing-collaboration

dbkegley:
We have also seen this bug in `Version: 1.0.1` so it may have been introduced initially with the `1.0.0` refactoring

toph-allen:
You can work around this issue by using the `appID` argument to `rsconnect::deployApp()`, e.g.

```r
rsconnect::deployApp(appDir = ".", appId = "a2859a31-4326-4125-9deb-39f31e5d4fbc")
```

aronatkins:
The v0.8.29 release falls into this clause when determining the deployment target:
https://github.com/rstudio/rsconnect/blob/9ce427edf03b6d243a4074264dbc950242b0a295/R/deployApp.R#L681-L686

The IDE calls `deployApp()` using the content name from the previous deployment, but uses the `account` associated with the (local) collaborator account. I have not uncovered how the IDE determines which account to provide when calling `deployApp()`.
https://github.com/rstudio/rstudio/blob/afbabb533c00add040c124126438177457ba5592/src/cpp/session/modules/SessionRSConnect.cpp#L280

aronatkins:
When the IDE recognizes that the target deployment does not correspond with an account that is configured, it attempts to discover a local account on the same server. As a result, `rsconnect::deployApp()` can be called with the original content name and server but a different account.

[...]

Issue #981, opened 2023-08-28
Unable to publish collaboratively


ischmidt20:
I would like to have multiple different people publish the same Shiny app to the same endpoint on Posit Connect. I have followed the guidelines [here](https://support.posit.co/hc/en-us/articles/360024007153-Tips-for-Collaborative-Publishing-on-RStudio-Connect-) precisely, yet every time someone other than the original publisher tries to publish, a new app is created.

In our repo, the `.dcf` file is located in `rsconnect/[posit url]/[username]/ShellApp,dcf`.

When the second user attempts to publish, this location is available as an option in the dropdown, and the resulting window looks as it should based on the video: "publish from" contains the second user's username, "publish to account" shows the original user.

However, upon publishing, a new app is created. Note that the original app has id 67.

```
i Creating application on server...
v Created application with id 78
```

I have tried moving the `.dcf` file to different locations, first outside of the username directory and then outside of the posit URL directory as well. Neither worked. I have also tried removing the `account` and `username` rows from the file, as discussed [here](https://community.rstudio.com/t/rstudio-connect-collaborative-publishing-troubleshooting-when-two-shiny-app-collaborators-fail-to-deploy-to-the-same-endpoint/42495/7). That also fails.

The only thing that "works" is a manual call to `deployApp` with the `appId` argument specified. This then creates a new dcf file at `rsconnect/[posit url]/[second username]/ShellApp,dcf` and all future deployments from the second user work. However, this is not ideal considering that manually calling `deployApp` requires specifying all of the other arguments that would normally be contained in the dcf (e.g. which files to include/exclude).

Clearly, `rsconnect` is either not reading or not using the `appId` value when reading a `.dcf` from another user. I haven't tested whether other settings fail to transfer over as well.

I am using the latest version of `rsconnect` on CRAN

ischmidt20:
OK, I think I've developed a workaround, similar to what I described above, but is still not ideal.

Second user pulls repo with latest copy of settings `.dcf` file from the first user.

Second user manually deploys app via `rsconnect::deployApp` and *manually specifies* `appId` (can be obtained from `.dcf` file or from browsing server itself).

This generates a new `.dcf` file for second user. Second user can then copy any additional settings from first user's `.dcf` file into the new `.dcf` file, such as excluded files.

Quit Rstudio and restart. (Not 100% sure that this is necessary).

[...]

PR #1021, opened 2023-10-23
republishing and collaborating

aronatkins:
Revisit how the deployment target is selected.

Use 0.8.29 implementation when reviewing this work. In particular, that version of `deploymentTarget()` did not offer any separation between the "owning" account and the "deploying" account. Details recorded in the in-memory deployment record were not necessarily consistent with the on-disk deployment record. Additionally, resolving that deployment record to an actual remote application occurred much later, in `applicationForTarget()`.
https://github.com/rstudio/rsconnect/blob/v0.8.29/R/deployApp.R

The implementation available in 1.0.0 until now did a good job consolidating this logic, but lost track of the fact that we may have a local user attempting to re-use deployment records created by some other user.

This change sees us use the following prioritized techniques to locate a deployment record:

1. When `appId` is provided, it must be used to locate the target content. A target (local) account is required to make this determination. When a local deployment record is not available, the application is loaded from the server. The target deployment record does not need to be owned by the local target account.
2. When `appName` is provided without `appId`, it must be used to locate the target content or create a new deployment target.
3. When neither `appName` nor `appId` is provided, the user is asked to choose from the available deployment records (which must correspond to a locally available account).
4. When there are no deployment records, a generated name is used along with a target account to remotely probe for that name (owned by the target account) before creating a new item.

Some collaboration instructions from the Connect user guide: https://docs.posit.co/connect/user/publishing/#publishing-collaboration

Related issues: #1019, #1013, #1007, #981 along with multiple internal reports.

This change may fix each of those issues, but they need more review to be sure.

### DIFF
--- a/R/applications.R
+++ b/R/applications.R
@@ -128,9 +128,7 @@ getAppByName <- function(client, accountInfo, name) {
   if (length(app)) {
     return(app[[1]])
   }
-
-  stop("No application found. Specify the application's directory, name, ",
-       "and/or associated account.", call. = FALSE)
+  return(NULL)
 }
 
 # Use the API to list all applications then filter the results client-side.
@@ -240,6 +238,10 @@ showLogs <- function(appPath = getwd(), appFile = NULL, appName = NULL,
   accountDetails <- accountInfo(deployment$account, deployment$server)
   client <- clientForAccount(accountDetails)
   application <- getAppByName(client, accountDetails, deployment$name)
+  if (is.null(application)) {
+    stop("No application found. Specify the application's directory, name, ",
+         "and/or associated account.", call. = FALSE)
+  }
 
   if (streaming) {
     # streaming; poll for the entries directly
--- a/R/applications.R
+++ b/R/applications.R
@@ -121,14 +121,21 @@ applications <- function(account = NULL, server = NULL) {
   return(res)
 }
 
-# Use the API to filter applications by name.
+# Use the API to filter applications by name and error when it does not exist.
 getAppByName <- function(client, accountInfo, name) {
   # NOTE: returns a list with 0 or 1 elements
   app <- client$listApplications(accountInfo$accountId, filters = list(name = name))
   if (length(app)) {
     return(app[[1]])
   }
-  return(NULL)
+  cli::cli_abort(
+    c(
+      "No application found",
+      i = "Specify the application directory, name, and/or associated account."
+    ),
+    call = NULL,
+    class = "rsconnect_app_not_found"
+  )
 }
 
 # Use the API to list all applications then filter the results client-side.
@@ -238,10 +245,6 @@ showLogs <- function(appPath = getwd(), appFile = NULL, appName = NULL,
   accountDetails <- accountInfo(deployment$account, deployment$server)
   client <- clientForAccount(accountDetails)
   application <- getAppByName(client, accountDetails, deployment$name)
-  if (is.null(application)) {
-    stop("No application found. Specify the application's directory, name, ",
-         "and/or associated account.", call. = FALSE)
-  }
 
   if (streaming) {
     # streaming; poll for the entries directly
--- a/R/applications.R
+++ b/R/applications.R
@@ -122,7 +122,7 @@ applications <- function(account = NULL, server = NULL) {
 }
 
 # Use the API to filter applications by name and error when it does not exist.
-getAppByName <- function(client, accountInfo, name) {
+getAppByName <- function(client, accountInfo, name, error_call = caller_env()) {
   # NOTE: returns a list with 0 or 1 elements
   app <- client$listApplications(accountInfo$accountId, filters = list(name = name))
   if (length(app)) {
@@ -133,7 +133,7 @@ getAppByName <- function(client, accountInfo, name) {
       "No application found",
       i = "Specify the application directory, name, and/or associated account."
     ),
-    call = NULL,
+    call = error_call,
     class = "rsconnect_app_not_found"
   )
 }
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -325,34 +325,20 @@ deployApp <- function(appDir = getwd(),
     cli::cli_rule("Preparing for deployment")
   }
 
+  forceUpdate <- forceUpdate %||% getOption("rsconnect.force.update.apps") %||% fromIDE()
+
   # determine the deployment target and target account info
   recordPath <- findRecordPath(appDir, recordDir, appPrimaryDoc)
-  if (!is.null(appId) && is.null(appName)) {
-    # User has supplied only appId, so retrieve app data from server
-    # IDE supplies both appId and appName so should never hit this branch
-    target <- deploymentTargetForApp(
-      appId = appId,
-      appTitle = appTitle,
-      account = account,
-      server = server
-    )
-  } else {
-    forceUpdate <- forceUpdate %||% getOption("rsconnect.force.update.apps") %||%
-      fromIDE()
-
-    # Use name/account/server to look up existing deployment;
-    # create new deployment if no match found
-    target <- deploymentTarget(
-      recordPath = recordPath,
-      appId = appId,
-      appName = appName,
-      appTitle = appTitle,
-      envVars = envVars,
-      account = account,
-      server = server,
-      forceUpdate = forceUpdate
-    )
-  }
+  target <- deploymentTarget(
+    recordPath = recordPath,
+    appId = appId,
+    appName = appName,
+    appTitle = appTitle,
+    envVars = envVars,
+    account = account,
+    server = server,
+    forceUpdate = forceUpdate
+  )
   if (is.null(target$appId)) {
     dest <- accountLabel(target$username, target$server)
     taskComplete(quiet, "Deploying {.val {target$appName}} to {.val {dest}}")
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -414,7 +414,7 @@ deployApp <- function(appDir = getwd(),
   }
   saveDeployment(
     recordPath,
-    previous = deployment,
+    deployment = deployment,
     application = application,
     metadata = metadata
   )
@@ -467,7 +467,7 @@ deployApp <- function(appDir = getwd(),
 
     saveDeployment(
       recordPath,
-      previous = deployment,
+      deployment = deployment,
       application = application,
       bundleId = bundle$id,
       metadata = metadata
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -271,15 +271,6 @@ deployApp <- function(appDir = getwd(),
     recordDir <- appSourceDoc
   }
 
-  cat(paste("deployApp entry:",
-            "appDir:", appDir,
-            "appId:", appId,
-            "appName:", appName,
-            "appTitle:", appTitle,
-            "account:", account,
-            "server:", server,
-            "\n"))
-
   # set up logging helpers
   logLevel <- match.arg(logLevel)
   quiet <- identical(logLevel, "quiet")
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -277,9 +277,9 @@ deployApp <- function(appDir = getwd(),
             "appName:", appName,
             "appTitle:", appTitle,
             "account:", account,
-            "server:",server,
+            "server:", server,
             "\n"))
- 
+
   # set up logging helpers
   logLevel <- match.arg(logLevel)
   quiet <- identical(logLevel, "quiet")
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -329,7 +329,7 @@ deployApp <- function(appDir = getwd(),
 
   # determine the target deployment record and deploying account
   recordPath <- findRecordPath(appDir, recordDir, appPrimaryDoc)
-  target <- deploymentTarget(
+  target <- findDeploymentTarget(
     recordPath = recordPath,
     appId = appId,
     appName = appName,
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -344,10 +344,10 @@ deployApp <- function(appDir = getwd(),
 
   if (is.null(deployment$appId)) {
     dest <- accountLabel(accountDetails$name, accountDetails$server)
-    taskComplete(quiet, "Deploying {.val {deployment$appName}} using {.val {dest}}")
+    taskComplete(quiet, "Deploying {.val {deployment$name}} using {.val {dest}}")
   } else {
     dest <- accountLabel(accountDetails$name, accountDetails$server)
-    taskComplete(quiet, "Re-deploying {.val {deployment$appName}} using {.val {dest}}")
+    taskComplete(quiet, "Re-deploying {.val {deployment$name}} using {.val {dest}}")
   }
 
   # Run checks prior to first saveDeployment() to avoid errors that will always
@@ -383,8 +383,8 @@ deployApp <- function(appDir = getwd(),
   if (is.null(deployment$appId)) {
     taskStart(quiet, "Creating application on server...")
     application <- client$createApplication(
-      deployment$appName,
-      deployment$appTitle,
+      deployment$name,
+      deployment$title,
       "shiny",
       accountDetails$accountId,
       appMetadata$appMode,
@@ -441,7 +441,7 @@ deployApp <- function(appDir = getwd(),
 
     taskStart(quiet, "Bundling {length(appFiles)} file{?s}: {.file {appFiles}}")
     bundlePath <- bundleApp(
-      appName = deployment$appName,
+      appName = deployment$name,
       appDir = appDir,
       appFiles = appFiles,
       appMetadata = appMetadata,
@@ -589,7 +589,7 @@ applicationDeleted <- function(client, deployment, recordPath, appMetadata) {
 
   path <- deploymentConfigFile(
     recordPath,
-    deployment$appName,
+    deployment$name,
     deployment$account,
     deployment$server
   )
@@ -597,8 +597,8 @@ applicationDeleted <- function(client, deployment, recordPath, appMetadata) {
 
   accountDetails <- accountInfo(deployment$account, deployment$server)
   client$createApplication(
-    deployment$appName,
-    deployment$appTitle,
+    deployment$name,
+    deployment$title,
     "shiny",
     accountDetails$accountId,
     appMetadata$appMode
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -414,7 +414,7 @@ deployApp <- function(appDir = getwd(),
   }
   saveDeployment(
     recordPath,
-    target = deployment,
+    previous = deployment,
     application = application,
     metadata = metadata
   )
@@ -467,7 +467,7 @@ deployApp <- function(appDir = getwd(),
 
     saveDeployment(
       recordPath,
-      target = deployment,
+      previous = deployment,
       application = application,
       bundleId = bundle$id,
       metadata = metadata
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -271,6 +271,15 @@ deployApp <- function(appDir = getwd(),
     recordDir <- appSourceDoc
   }
 
+  cat(paste("deployApp entry:",
+            "appDir:", appDir,
+            "appId:", appId,
+            "appName:", appName,
+            "appTitle:", appTitle,
+            "account:", account,
+            "server:",server,
+            "\n"))
+ 
   # set up logging helpers
   logLevel <- match.arg(logLevel)
   quiet <- identical(logLevel, "quiet")
@@ -327,7 +336,7 @@ deployApp <- function(appDir = getwd(),
 
   forceUpdate <- forceUpdate %||% getOption("rsconnect.force.update.apps") %||% fromIDE()
 
-  # determine the deployment target and target account info
+  # determine the target deployment record and deploying account
   recordPath <- findRecordPath(appDir, recordDir, appPrimaryDoc)
   target <- deploymentTarget(
     recordPath = recordPath,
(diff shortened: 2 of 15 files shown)


aronatkins (commented):
Will need NEWS ahead of merge.

R/applications.R
```diff
@@ -128,9 +128,7 @@ getAppByName <- function(client, accountInfo, name) {
   if (length(app)) {
     return(app[[1]])
   }
-
-  stop("No application found. Specify the application's directory, name, ",
-       "and/or associated account.", call. = FALSE)
+  return(NULL)
```
hadley, line 131: Given that you have to repeat this error message four times below, it might be better to continue to throw the error then use `tryCatch()` in `deploymentTarget()`. If you do that, probably should convert to `cli::cli_abort()` and provide a class, e.g.

```R
cli::cli_abort(
   c(
     "No application found",
     i = "Specify the application's directory, name, and/or associated account"
   ),
   class = "rsconnect_app_not_found"
)
```

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
-deploymentTarget <- function(recordPath = ".",
-                             appId = NULL,
-                             appName = NULL,
-                             appTitle = NULL,
-                             envVars = NULL,
-                             account = NULL,
-                             server = NULL,
-                             forceUpdate = FALSE,
-                             error_call = caller_env()) {
+deploymentTarget <- function(
```
hadley, line 3: Spurious WS change?

I wonder if we should change this name — it's currently a noun, which makes me think it should mostly be about creating an object; but actually it contains quite a lot of logic for finding the deployment. So maybe `findDeployment()`?

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
+  envVars = NULL,
+  account = NULL,
+  server = NULL,
+  forceUpdate = FALSE,
+  error_call = caller_env()
+) {
 
-  appDeployments <- deployments(
+  if (!is.null(appId)) {
+    return(deploymentTargetFromAppId(
```
hadley, line 16: Then this becomes something like `findDeploymentWithAppId()`

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
+      appName = appName,
+      appTitle = appTitle,
+      envVars = envVars,
+      account = account,
+      server = server,
+      error_call = error_call
+    ))
+  }
+
+  if (!is.null(appName)) {
```
hadley, line 28:
```suggestion
  } else if (!is.null(appName)) {
```

? Too make it more clear we're grouping two early returns together.

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
+  return(deploymentTargetFromAppName(
+    recordPath = recordPath,
+    appName = appName,
+    appTitle = appTitle,
+    envVars = envVars,
+    account = accountDetails$name,
+    server = accountDetails$server,
+    forceUpdate = forceUpdate,
+    error_call = error_call
+  ))
```
hadley, line 74:
```suggestion
  deploymentTargetFromAppName(
    recordPath = recordPath,
    appName = appName,
    appTitle = appTitle,
    envVars = envVars,
    account = accountDetails$name,
    server = accountDetails$server,
    forceUpdate = forceUpdate,
    error_call = error_call
  )
```

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
-    deployment <- disambiguateDeployments(appDeployments, error_call = error_call)
-    updateDeploymentTarget(deployment, appTitle, envVars)
+  # Existing local deployment record.
+  if (nrow(appDeployments) == 1) {
+    deployment <- appDeployments[1, ]
+    deployment <- updateDeploymentTarget(deployment, appTitle, envVars)
+    return(list(
+      accountDetails = accountDetails,
+      deployment = deployment
+    ))
```
hadley, line 125:
```suggestion
    list(
      accountDetails = accountDetails,
      deployment = deployment
    )
```

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
   }
-}
 
-deploymentTargetForApp <- function(appId,
-                                   appTitle = NULL,
-                                   account = NULL,
-                                   server = NULL) {
-  accountDetails <- findAccount(account, server)
+  # No local deployment record. Get it from the server.
```
hadley, line 128: I think this would be more clear if you three if clauses corresponding to 0 matches, 1 match, more than 1 match.

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
-          appId <- thisApp$id
-          appName <- thisApp$name
-        } else {
-          appName <- uniqueName
-        }
-      }
-    }
+# Discover the deployment target given appId.
+#
+# When appId is provided, all other information is secondary. An appId is an indication from the
```
hadley, line 79: I found this comment very useful for understand the problem and where I went wrong previously.

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
-  createDeploymentTarget(
+  # Note: The account+server of this deployment record may
+  # not correspond to the original content creator.
+  deployment <- createDeploymentTarget(
     appName = application$name,
     appTitle = application$title %||% appTitle,
     appId = application$id,
-    envVars = NULL,
+    envVars = envVars,
     username = application$owner_username %||% accountDetails$name,
```
hadley, line 138: Should we change the arguments to `createDeploymentTarget()` to more clearly differentiate the username of the current user vs. the username of the account owner? Might need to be in a separate PR to avoid cluttering this one, but I think it's likely to make understanding the code easier in the future.

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
+  account = NULL,
+  server = NULL,
+  forceUpdate = FALSE,
+  error_call = caller_env()
+) {
+
+  appDeployments <- deployments(
+    appPath = recordPath,
+    nameFilter = appName,
+    accountFilter = account,
```
hadley, line 169: Why is `account` used here? i.e. what makes this path different from `deploymentTargetFromAppId()`?

tests/testthat/_snaps/deploymentTarget.md
```diff
@@ -3,7 +3,7 @@
     Code
       deploymentTarget()
     Condition
-      Error in `deploymentTarget()`:
+      Error in `accountInfo()`:
```
hadley, line 6: This indicates that `caller_env` probably isn't being passed correctly.

R/deployApp.R
```diff
@@ -271,6 +271,15 @@ deployApp <- function(appDir = getwd(),
     recordDir <- appSourceDoc
   }
 
+  cat(paste("deployApp entry:",
```
aronatkins, line 274: Reminder: Remove this code ahead of merge.

R/applications.R
aronatkins: Thanks. I'll play with this option but may also refactor into a helper to enforce the existence of an application, which removes the repetition of the error.

R/deployApp.R
```diff
@@ -418,15 +415,15 @@ deployApp <- function(appDir = getwd(),
...
+        application <- applicationDeleted(client, deployment, recordPath, appMetadata)
         taskComplete(quiet, "Created application with id {.val {application$id}}")
         application
       }
     )
   }
   saveDeployment(
     recordPath,
-    target = target,
+    target = deployment,
```
aronatkins, line 426: note: This is the remaining use of `target` that is a deployment record. Consider renaming the argument, though that means renaming the in-function variable, as well.

aronatkins: note: The 0.8.29 implementation wrote deployment records only when the target (local) account was the content owner, using username matching. This PR does not revert to that behavior; all deploys write to disk, though we try to maintain the account+server whenever possible.

R/deploymentTarget.R
aronatkins: Not a spurious whitespace change. I intentionally shifted to this style, but can revert if it's not a supported convention. With so many arguments and very long function names (e.g. the helpers called from here), the indentation felt a bit crazy.

We already have `findDeployment()`! 😜

Maybe `findDeploymentTarget()`, but I'll also try to explain why we need `findDeployment()` and the difference between the two "deployment discovery" functions.

R/deploymentTarget.R
aronatkins: My initial feeling is to keep this as-is; most of these deployment discovery functions are using early returns. Shifting to an if-else makes me want to group the remainder of the function in a massive "else" clause, which leads to more indentation.

It's a minor style preference.

R/deploymentTarget.R
aronatkins: Would you always favor removing the final return and prefer the implicit return? I like the fact that it's visually obvious that the two calls to `deploymentTargetFromAppName()` look the same -- with similar call signature and having their results immediately returned.

R/deploymentTarget.R
aronatkins: Thanks. It took _so_ much code-reading and reasoning about the previous workflow to see it. I'm hoping that being much more explicit will help us in the future.

R/deploymentTarget.R
aronatkins: I'm not sure I follow. We have:

```r
if (num > 1) {
  error
}

if (num == 1) {
  return(...)
}

# else no matches. do some work then
return(...)
```

I tend to favor this style because it avoids deep conditional clauses in many scenarios. What is the advantage of forcing this into an explicit if-else-if-else?

R/deploymentTarget.R
aronatkins: The deployment record has the account+server corresponding to local account record for the user responsible for creating the deployment record on disk. This is usually the creator, but not always (e.g. collaborator publish using `appId`). The username is associated with the creating user (unsure if we get this back from shinyapps.io).

I'm not sure how to best represent this possible inconsistency. Happy to file an issue for further discussion.

R/deploymentTarget.R
aronatkins: When given `appId`, that identifies a unique remote item. On one server, there cannot be two content items having the same identifier.

When using `appId`, we search across all deployment records, even when there is no local account for the account+server referenced by the deployment record.

When `appId` does not identify existing content, deployment fails.

When given `appName`, that identifies a remote item owned by the current user. Two people can use the same name to identify different items.

When using `appName`, the search is restricted to the incoming account+server. When there is no incoming account+server, the search is restricted to deployments which have a corresponding local account.

When `appName` does not identify existing content, new content is created.

tests/testthat/_snaps/deploymentTarget.md
aronatkins: Yes, absolutely. The problem is that `accountInfo()` is part of the public API and does not accept `caller_env`. Previously, we were calling `findAccount()` in some contexts. That does take `caller_env`, but does not return the full account detail. I wanted to avoid reasoning about what shape of account information was available, so I switched to always calling `accountInfo()`. The drawback is that we lose `caller_env`.

How would you resolve this complication?

R/deploymentTarget.R
hadley: We generally reserve the style you moved to for cases where the style you moved from has arguments that don't all fit on one line. 

Hmmm, I had forgotten about `findDeployment()`. Maybe this function is something more like `findAndUpdateDeployment()`? Or is there some difference between a deployment and a "deployment target" that I've forgotten?

R/deploymentTarget.R
hadley: :+1:

R/deploymentTarget.R
hadley: Sorry, I misread this because I was looking at the unified diff.

R/deploymentTarget.R
hadley: I like the separate if statements when the code blocks are long, but here where they're relatively short, I think the single if block makes it easier to verify that you've covered all the conditions:

```R
if (num > 1) {

}  else if (num == 1) {

} else } # num == 0

}
```

R/deploymentTarget.R
hadley: I was thinking maybe we could rename `username` to something like `owner_account` to make it clear that this is also an account name, but it's the account name for the original author.

R/deploymentTarget.R
hadley: Thanks, that's helpful! I wonder if it's worth putting in as a comment in `deploymentTarget()`?

tests/testthat/_snaps/deploymentTarget.md
hadley: I'd be tempted to push more of the logic out of `accountInfo()` and into `findAccount()`, then switch back to calling `findAccount()` everywhere. Then `accountInfo()` just becomes a very thin wrapper around `findAccount()`.

Alternatively, you could accept that `findAccount()` is mostly an internal facing function that users are unlikely to call directly, and just add an `error_call` argument to it.

R/deploymentTarget.R
```diff
@@ -1,93 +1,241 @@
...
+  # When the appName does not identify a target, see if it exists on the server. That content is
+  # conditionally used. A resolved account is required.
+  accountDetails <- accountInfo(account, server)
+  if (!isPositCloudServer(accountDetails$server)) {
+    client <- clientForAccount(accountDetails)
+    application <- getAppByName(client, accountDetails, appName)
+    if (!is.null(application)) {
+      uniqueName <- findUnique(appName, application$name)
+      if (shouldUpdateApp(application, uniqueName, forceUpdate)) {
+        deployment <- createDeploymentTarget(
```
hadley, line 206: I wonder if it would be useful to have an explicit function called that takes an application and an accountDetails and creates a deploymentTarget

R/deploymentTarget.R
aronatkins: This became `findDeploymentTargetByAppId()`

R/deploymentTarget.R
aronatkins: This became `findDeployment()`

tests/testthat/_snaps/deploymentTarget.md
aronatkins: I've created `findAccountInfo()` as a version of `accountInfo()` that accepts error context and lives alongside `findAccount()`, which continues to identify the target account name+server.

R/deploymentTarget.R
aronatkins: I'm going to leave this as-is for now, as each of the three `findDeploymentTarget*` functions follow the same pattern -- dispatch for some cases then do something slightly more complicated when there are no records.

R/deployApp.R
aronatkins: The `target` has been renamed.

tests/testthat/test-deploymentTarget.R
```diff
@@ -300,3 +472,79 @@ test_that("findUnique always returns unique name", {
...
+    envVars = c("VAR-NAME"),
+    appId = "1",
+    username = "alice.username",
+    account = "alice",
+    server = "example.com",
+    version = deploymentRecordVersion
+  ))
+})
+
+test_that("updateDeployment ignores NULL updates", {
```
aronatkins, line 526: Adding these tests let me see that we were mixing records from disk which used `name` and records created on-the-fly which used `appName`. We are now consistent.

R/deploymentTarget.R
```diff
@@ -1,106 +1,277 @@
...
-                             account = NULL,
-                             server = NULL,
-                             forceUpdate = FALSE,
-                             error_call = caller_env()) {
+# Discover the deployment target given the passed information.
+#
+# Returns a list containing a deployment record and the account details to use
+# when performing the deployment.
+#
+# When appId is provided, it must identify an existing application. The
```
hadley, line 6: I'd maybe spread these amongst the existing implementation?

hadley (approved):
Looks like a nice improvement to code clarity!

tests/testthat/test-deploymentTarget.R
```diff
@@ -83,46 +197,70 @@ test_that("succeeds if there's a single existing deployment", {
...
+  expect_equal(deployment$name, "test")
+  expect_equal(deployment$username, "ron")
+  expect_equal(deployment$account, "ron")
+  expect_equal(deployment$server, "example.com")
+  expect_equal(deployment$version, "999")
 })
 
-test_that("errors if single deployment and appId doesn't match", {
+test_that("appId is used even when name does not match", {
   local_temp_config()
```
toph-allen, line 233: Just occurs to me — do we / should we emit a warning when this happens? Or not worth it?

I guess if it works and does what the user wants, no need to confuse them (especially because of the somewhat confusing difference between `name` and `title`).

toph-allen (approved):
Looks good to me. A lot clearer for posterity too.

R/deploymentTarget.R
aronatkins: I believe this is already the case -- each of the `findDeploymentTargetBy*` helpers explains its responsibility. This is the roll-up.

R/applications.R
```diff
@@ -121,16 +121,21 @@ applications <- function(account = NULL, server = NULL) {
...
   }
-
-  stop("No application found. Specify the application's directory, name, ",
-       "and/or associated account.", call. = FALSE)
+  cli::cli_abort(
+    c(
+      "No application found",
+      i = "Specify the application directory, name, and/or associated account."
+    ),
+    call = NULL,
```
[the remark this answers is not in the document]
aronatkins: Thanks. Done.

R/deployments.R
```diff
@@ -85,28 +85,32 @@ deploymentFields <- c(
 
 deploymentRecordVersion <- 1L
 
+# Save a deployment record to disk using an incoming record (which may or may
+# not correspond to an existing on-disk deployment record). Created by
+# deploymentRecord() or by findDeploymentTarget(), and possibly loaded from
+# disk.
 saveDeployment <- function(recordDir,
-                           target,
+                           previous,
```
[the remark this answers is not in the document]
aronatkins: The in-function variable is called `deployment`, which is why I didn't initially use this name. It seems like a safe change, however.

R/deploymentTarget.R
```diff
@@ -1,106 +1,277 @@
...
+    deployment <- disambiguateDeployments(appDeployments, error_call = error_call)
+    deployment <- updateDeployment(deployment, appTitle, envVars)
+    accountDetails <- findAccountInfo(deployment$account, deployment$server, error_call = error_call)
+    return(list(
+      accountDetails = accountDetails,
+      deployment = deployment
+    ))
+  }
+
+  # When the appName does not identify a target, see if it exists on the server. That content is
```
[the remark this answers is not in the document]
aronatkins: Yes. I'll revise.

R/deploymentTarget.R
```diff
@@ -1,106 +1,277 @@
...
+# Returns a list containing a deployment record and the account details to use
+# when performing the deployment.
+#
+# When appId is provided, it must identify an existing application. The
+# application may have been created by some other user. That application may
+# or may not have an existing deployment record on disk.
+#
+# When using appId, a search across all deployment records occurs, even when
+# there is no local account+server referenced by the deployment record. This
+# lets us identify on-disk deployment records created by some collaborator.
```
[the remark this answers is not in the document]
aronatkins: Yes. I'll clarify.

R/deploymentTarget.R
```diff
@@ -1,106 +1,277 @@
...
+#
+# It is an error when appId does not identify an existing application.
+#
+# When appName is provided, it may identify an existing application owned by
+# the calling user (e.g. associated with a locally known account).
+#
+# When using appName, the search across deployment records is restricted to
+# the incoming account+server. When there is no incoming account+server, the
+# search is restricted to deployments which have a corresponding local
+# account.
```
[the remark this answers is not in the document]
aronatkins: I'll clarify the docs.

R/deploymentTarget.R
```diff
@@ -1,106 +1,277 @@
...
+      if (shouldUpdateApp(application, uniqueName, forceUpdate, error_call = error_call)) {
+        deployment <- createDeploymentFromApplication(application, accountDetails)
+        deployment <- updateDeployment(deployment, appTitle, envVars)
+        return(list(
+          accountDetails = accountDetails,
+          deployment = deployment
+        ))
+      } else {
+        appName <- uniqueName
+      }
```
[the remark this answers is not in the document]
aronatkins: When the content exists remotely with no local deployment record, we tell the person that we found that content and ask if they want to update that in-use name or create a new, distinct deployment.

The "forced" update skips this prompt and unconditionally updates the remote application.

R/deploymentTarget.R
```diff
@@ -110,8 +281,20 @@ createDeploymentTarget <- function(appName,
   )
 }
 
-updateDeploymentTarget <- function(previous, appTitle = NULL, envVars = NULL) {
-  createDeploymentTarget(
+createDeploymentFromApplication <- function(application, accountDetails) {
```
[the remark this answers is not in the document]
aronatkins: Yes. This is an application object.

tests/testthat/test-deploymentTarget.R
aronatkins: We do not emit a warning. We talked about emitting warnings earlier in the investigation, but that's because we thought that folks were making mistakes when calling `deployApp()`. In fact, `deployApp()` was making the wrong deployment choices.

ChaitaC:
Verified the fix for https://github.com/rstudio/rsconnect/issues/1013

Before the fix, when using CRAN Version 1.1.1
```
── Preparing for deployment ────────────────────────────────────────────────────
✔ Deploying "shinyappsIO" to "server: shinyapps.io / username: chaitatest"
Error in `POST()`:
! <https://api.shinyapps.io/v1/applications/> failed with HTTP status
  409
Application exists with name: shinyappsIO
Backtrace:
    ▆
 1. └─rsconnect::deployApp(...)
 2.   └─client$createApplication(...)
 3.     └─rsconnect:::POST_JSON(service, authInfo, "/applications/", json)
 4.       └─rsconnect:::POST(...)
 5.         └─rsconnect:::httpRequestWithBody(...)
 6.           └─rsconnect:::handleResponse(httpResponse, error_call = error_call)
 7.             └─rsconnect (local) reportError(json$error)
 8.               └─cli::cli_abort(...)
 9.                 └─rlang::abort(...)
Execution halted
```

Using the development version from GitHub, the existing app can be redeployed and replaced without any issue.
```
── Deployment complete ─────────────────────────────────────────────────────────
✔ Successfully deployed to <https://chaitatest.shinyapps.io/SHinyappsIO/>
Deployment completed: https://chaitatest.shinyapps.io/SHinyappsIO/
```

ChaitaC:
Verified https://github.com/rstudio/rsconnect/issues/981
Verified https://github.com/rstudio/rsconnect/issues/1007

Collaborator publishing works without any issues. A collaborator can update the content as long as he has an accurate rsconnect dir.

ChaitaC:
⚠️ If a user lacks the necessary permissions to collaborate but possesses the required rsconnect directory, the following error message is displayed:

Screenshot 2023-11-08 at 3 09 50 PM

✅ When a user possesses the appropriate collaborative privileges, the interface appears as follows:
Screenshot 2023-11-08 at 3 11 07 PM

ChaitaC:
@aronatkins the error message 39] in it does not make sense.  See for image https://github.com/rstudio/rsconnect/pull/1021#issuecomment-1802691309

this is an edge case though

aronatkins:
I have filed https://github.com/rstudio/rsconnect/issues/1026 to track the escape codes that are presented by the IDE when the target content has been deleted (or is not visible to the user).

Issue #1008, opened 2023-09-27
Content metadata updates on Connect don't propagate to DCF file


toph-allen:
rsconnect version: 1.0.1

## Description

When content metadata fields (`title`, `name`) are changed on Connect, via the GUI or API, these don't propagate back to the deployment record DCF file. This can cause confusing behavior in the IDE.

## Reproduction

Steps to reproduce:

1. Deploy content to Connect with `rsconnect` or push-button.
2. Change content's `title` on Connect.
3. Deploy an update to the same destination.

Expected behavior:

- The new title from Connect is populated into the DCF file.

Actual behavior:

- The metadata in the DCF file is not updated with new metadata from server.

## Notes

Other metadata is also not updated, e.g. `name` (which can be changed via the API)

The IDE lists available redeploy destinations using the `title` field from the DCF file, so these entries don't reflect changes to the title made in the Connect GUI or via the API. (This is the behavior when the `name`s in Connect and the DCF file match. If they don't match, the deployment workflow shows a text field pre-populated with the IDE project title, even though publishing doesn't update the title on Connect.)

toph-allen:
@hadley This is the issue I mentioned

aronatkins:
`rsconnect::syncAppMetadata()` may help here, but it may need improvements to update a subset of applications. As written, it updates all content owned by local accounts.

toph-allen:
I tested `rsconnect::syncAppMetadata` after changing the title on the server. It updated the deployment record immediately, but I had to restart the server before seeing the change in push-button publishing because of its caching behavior.

hadley:
@aronatkins I don't think `syncAppMetadata()` is as bad as you think it is, because it does only update deployments for the current app directory, which is likely to be a small number. 

@toph-allen are you suggesting that we update the metadata on deploy? Won't this generally be too late, since you'll already have been confused by the dialog? I'm not quite sure I fully understand the problem here.

aronatkins:
@hadley - When I said "subset of applications" I was speaking of its need to gain deployment targeting within a specific project directory. You're absolutely right - it's generally a small number. It does not support, for example, updating the deployment records originally created by a collaborator.

@toph-allen was testing `rsconnect::syncAppMetadata()` as an alternative to having `deployApp()` update the metadata automatically. With confirmation that it works, that is probably sufficient to close this issue.

[...]

PR #1023, opened 2023-10-27
generate title and name; name is normalized

### LINKED ISSUE #1022: Published content names are being set lowercase characters.
The the development version of `rsconnect` sets the name of the content to lowercase. 

For example, when publishing to posit.cloud, the content is named `EnLS3Mz3UZ` and the resulting published content is named `enls3mz3uz`. 

Steps to reproduce: 
1. Create an HTML file that is camel cased
2. Publish the HTML to posit.cloud or Posit Connect

rsconnect::deployDoc('EnLS3Mz3UZ.html', server='posit.cloud')
rsconnect::deployDoc('EnLS3Mz3UZ.html', server='<posit connect server>')

What the published content looks like with the development version of rsconnect:
Screenshot 2023-10-26 at 1 38 45 PM

What the published content looks like with the CRAN release of rsconnect.
Screenshot 2023-10-26 at 1 46 54 PM

aronatkins:
~When deploying from a directory named "Incredible Shiny Application", the generated name for Posit Connect is `Incredible_Shiny_Application` and for shinyapps.io is `incredible_shiny_application`. Both services require `a-zA-Z0-9-_` for names, but we lower-case only for shinyapps.io.~

~The name is derived from an incoming title, when available, and otherwise from the content path.~

~When deploying without a title, Connect uses the incoming name to seed the title. This behavior is not changing.~

When deploying from a directory named "Incredible Shiny Application", a generated title mirrors the directory name and name is generated from the title and normalized like `incredible_shiny_application`. 

A generated name is used only when one is not already provided.

Title changes on the server are now reflected in the deployment record.

fixes #1022
related to #1008

### DIFF
--- a/R/deploymentTarget.R
+++ b/R/deploymentTarget.R
@@ -91,7 +91,11 @@ findDeploymentTarget <- function(
   # Otherwise, identify a target account (given just one available or prompted
   # by the user), generate a name, and locate the deployment.
   accountDetails <- findAccountInfo(account, server, error_call = error_call)
-  appName <- generateAppName(appTitle, recordPath, accountDetails$name, unique = FALSE)
+  appName <- generateDefaultName(
+    recordPath = recordPath,
+    title = appTitle,
+    server = accountDetails$server
+  )
   return(findDeploymentTargetByAppName(
     recordPath = recordPath,
     appName = appName,
@@ -315,27 +319,56 @@ updateDeployment <- function(previous, appTitle = NULL, envVars = NULL) {
   )
 }
 
-defaultAppName <- function(recordPath, server = NULL) {
+normalizeName <- function(name, server = NULL) {
+  if (is.null(name)) {
+    return("")
+  }
+
+  if (isShinyappsServer(server)) {
+    name <- tolower(name)
+  }
+
+  # Replace non-alphanumerics with underscores, trim to length 64
+  name <- gsub("[^[:alnum:]_-]+", "_", name, perl = TRUE)
+  name <- gsub("_+", "_", name)
+  if (nchar(name) > 64) {
+    name <- substr(name, 1, 64)
+  }
+
+  name
+}
+
+titleFromPath <- function(recordPath) {
   if (isDocumentPath(recordPath)) {
-    name <- file_path_sans_ext(basename(recordPath))
-    if (name == "index") {
-      # parent directory will give more informative name
-      name <- basename(dirname(recordPath))
+    title <- file_path_sans_ext(basename(recordPath))
+    if (title == "index") {
+      # parent directory will give more informative name+title
+      title <- basename(dirname(recordPath))
     } else {
       # deploying a document
     }
   } else {
     # deploying a directory
-    name <- basename(recordPath)
+    title <- basename(recordPath)
   }
+  title
+}
 
-  if (isShinyappsServer(server)) {
-    # Replace non-alphanumerics with underscores, trim to length 64
-    name <- tolower(gsub("[^[:alnum:]_-]+", "_", name, perl = TRUE))
-    name <- gsub("_+", "_", name)
-    if (nchar(name) > 64) {
-      name <- substr(name, 1, 64)
-    }
+# Determine name given a file or directory path and (optional) title.
+#
+# Prefer generating the name from the incoming title when provided and
+# fall-back to one derived from the target filename.
+#
+# Name is guaranteed to conform to [a-zA-Z0-9_-]{0,64}. Minimum length is
+# enforced by the server.
+#
+# Names produced for Shinyapps.io deployments are lower-cased.
+generateDefaultName <- function(recordPath, title = NULL, server = NULL) {
+  name <- normalizeName(title, server = server)
+
+  if (nchar(name) < 3) {
+    title <- titleFromPath(recordPath)
+    name <- normalizeName(title, server = server)
   }
 
   name
--- a/R/deploymentTarget.R
+++ b/R/deploymentTarget.R
@@ -91,10 +91,12 @@ findDeploymentTarget <- function(
   # Otherwise, identify a target account (given just one available or prompted
   # by the user), generate a name, and locate the deployment.
   accountDetails <- findAccountInfo(account, server, error_call = error_call)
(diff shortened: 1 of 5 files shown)


aronatkins:
CI failures due to `r-project.org` DNS trouble. Will re-run once that clears.

hadley (approved):
Should we do something to `generateAppName()` to make it less likely to make this mistake again?

aronatkins:
> Should we do something to `generateAppName()` to make it less likely to make this mistake again?

I considered modifying `generateAppName()` but did not want to change its implementation in a breaking fashion, since it is a public function. We disconnected `generateAppName()` during the build-up to 1.0.0 but didn't mark it as deprecated at that point in time.

Open to suggestions.

hadley:
Maybe we should deprecate it. It is public, but I doubt many people are using it. There's no evidence of another CRAN package using it: https://github.com/search?q=org%3Acran%20generateAppName&type=code. But maybe it's used by the IDE?

hadley:
Gah, yes: https://github.com/rstudio/rstudio/blob/ad0e5854c7a0abed5728c476ec3611f458e52d82/src/cpp/session/modules/SessionRSConnect.R#L592-L594

aronatkins:
The IDE calls `generateAppName()` to create a name from an incoming title; that makes me want to shift back to using `generateAppName()` but remove its lower-casing... We could even pass in an (optional) incoming title and act more like the IDE when `deployApp()` is directly called.

aronatkins:
@hadley Given https://github.com/rstudio/rsconnect/issues/1022#issuecomment-1785343724, how do we want `deployApp()` to behave?

* generate lower-cased names, which leads to lower-case default titles. This is the behavior of rsconnect 0.8.29 and the current development version of rsconnect. Relies on `generateAppName()`, mirroring the IDE.
* generate mixed-case names, which leads to mixed-case default titles. This is the behavior of rsconnect 1.0.0 through the current CRAN version.
* generate lower-cased names and mixed-case titles. This leaves names compatible with rsconnect 0.8.29 (and preserves any existing look-ups), while letting rich file names be directly used as the content title.

We cannot modify `generateAppName()` because its generated (lower-cased) names are used by the IDE to detect a duplicate deployment.

The best answer would be to use the incoming filename to generate a title alongside the name, which needs less normalization, and then normalize only the name. I considered this option but abandoned it because it felt like too much of a change for what felt like a bug fix.

I'm leaning towards having `deployApp()` explicitly infer the title.

hadley:
Yeah, letting `deployApp()` explicitly infer the title sounds good to me.

aronatkins:
> Yeah, letting `deployApp()` explicitly infer the title sounds good to me.

OK. I'll take a stab at that and up date this PR.

R/deploymentTarget.R
```diff
@@ -91,7 +91,13 @@ findDeploymentTarget <- function(
   # Otherwise, identify a target account (given just one available or prompted
   # by the user), generate a name, and locate the deployment.
   accountDetails <- findAccountInfo(account, server, error_call = error_call)
-  appName <- generateAppName(appTitle, recordPath, accountDetails$name, unique = FALSE)
+  appTitle <- generateTitle(recordPath = recordPath, title = appTitle)
```
aronatkins, line 94: This has the side-effect of generating an `appTitle` for shinyapps.io content, which does not support title. Alternatively, we could generate a title only for Connect deployments.

image

The other drawback of this approach is that as soon as we have generated the title, we lose track of the fact that the title was not user-specified. This means that the generated title will always overwrite the deployment record title.

R/deployments.R
```diff
@@ -98,7 +98,7 @@ saveDeployment <- function(recordDir,
                            addToHistory = TRUE) {
   deployment <- deploymentRecord(
     name = deployment$name,
-    title = deployment$title,
+    title = application$title %||% deployment$title,
```
aronatkins, line 101: This helps make sure that a server-updated title is reflected on the client.

R/deploymentTarget.R
aronatkins: We may want to distinguish between "user provided title" and "generated title" when updating a discovered deployment record.

```r
title <- userTitle %||% deployment$title %||% generatedTitle
```

aronatkins: After more testing, we probably want to avoid generating titles for shinyapps.io.

The IDE deployment pane forces a normalized "title", which becomes the title and name:

image

After manually normalizing the title with "incredible-shiny-application" and completing this workflow, the deployment record name and title both receive the same value (as forced by the IDE).

aronatkins: Using the CRAN version of `rsconnect::deployApp()` leaves a shinyapps.io deployment record with `title=NULL`. The IDE deployment workflow creates deployment records with title and name both with the same value.

The IDE allows mixed-case names(titles) for shinyapps.io, but mixed-case is not produced by `generateAppName()`.

R/deployments.R
aronatkins: We may want to take this change even if we decide to abandon the other name-generation aspects of this PR.

toph-allen: How will automatically generating a title interact with this? Will auto-generated titles overwrite titles that have been changed on Connect?

toph-allen (commented):
I don't feel like I have a completely clear picture of some of the effects of this change. Open questions for me are around how these changes interact with titles that have been updated on Connect.

Prior to this update, the default title was an empty string. Now, if we aren't provided a title, we generate one, and use this to generate a name. Will this overwrite titles that have been changed on Connect?

How does this interact with deployment records? It looks like we get titles from deployment records ([see here](https://github.com/rstudio/rsconnect/blob/e2236c26aee9052ac9c4bb251dba4d3a0b9a85cb/R/deploymentTarget.R#L313)). Will this lead to a title that has been changed on Connect reverting to the previously-published title when it is updated? (Outside of the scope of these changes, but a question raised by them.)

The pre-1.0.0 behavior was to use a non-lowercased `appName`, right? Perhaps better to revert to that, and only use `appTitle` if provided.

aronatkins:
Discarding this PR; it has not been necessary.

PR #1027, opened 2023-11-21
add envVars to example in deployApp()

fh-mthomson:
Pulls out helpful example from https://github.com/rstudio/rsconnect/issues/859#issuecomment-1600949465

This workflow and messaging was _very_ clear and easy!

Additionally, it'd be helpful to mention `rsconnect` as an option to set env vars _prior_ to deployment in the [Posit Connect User Guide](https://docs.posit.co/connect/user/content-settings/index.html#content-vars), which currently only references the Connect API.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -198,7 +198,8 @@
 #' # find the Quarto binary
 #' deployApp("~/projects/quarto/site1")
 #'
-#' # deploy application with environment variables (e.g., `SECRET_PASSWORD=XYZ` is set via an ~/.Renviron file)
+#' # deploy application with environment variables
+#' # (e.g., `SECRET_PASSWORD=XYZ` is set via an ~/.Renviron file)
 #' rsconnect::deployApp(envVars = c("SECRET_PASSWORD"))
 #' }
 #' @seealso [applications()], [terminateApp()], and [restartApp()]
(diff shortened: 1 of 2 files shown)


hadley:
Except that's not quite right either because running `Sys.setenv()` in the console will mean that your secret is now in the `.Rhistory` file 😞

hadley:
@fh-mthomson do you use hashicorp vault? I've been trying to find someone who uses it to help me learn if there's something we should do to make it easier to use from R.

fh-mthomson:
I do! As a starting point, check out https://github.com/vimc/vaultr (h/t @richfitz for the great work!)

General feedback:
* Solid coverage of main [Vault auth methods](https://developer.hashicorp.com/vault/docs/auth): token, user/pass, approle, LDAP
  * LDAP was included in recent 1.2.0 release (https://github.com/vimc/vaultr/pull/46)
  * AWS auth has been stuck (https://github.com/vimc/vaultr/issues/28) - would be helpful for customers hosting Connect on EKS to be able to easily have a service user access Vault
* Given R6 interface, I found it hard to pick up / navigate docs initially, but most flows, once established, can be wrapped: `vault_client()` -> login -> fetch secret from a given path

hadley:
Thanks @richfitz & @fh-mthomson!

Issue #928, opened 2023-07-24
signatureHeaders sees OpenSSL error when FIPS enabled


aronatkins:
When using FIPS in a Rocky Linux 9 container, calling `openssl::signature_create` can err like:

```
16.63 > rawsig <- openssl::signature_create(charToRaw(canonicalRequest), key = private_key)
16.63 Error in hash_sign(md, sk) :
16.63   OpenSSL error: 807014FC877F0000:error:03000098:digital envelope routines:evp_pkey_ctx_set_md:invalid digest:crypto/evp/pmeth_lib.c:959:
16.63 Calls: <Anonymous> -> hash_sign
16.63 Execution halted
```

This can happen when people attempt to configure a Posit Account using the RStudio IDE or using `rsconnect::connectUser`.

Folks encountering this problem can use `rsconnect::connectApiUser` as a workaround.

Here is the `Dockerfile`:

```dockerfile
FROM rstudio/r-base:4.3-rockylinux9

RUN dnf update -y && dnf install -y openssl-devel
RUN fips-mode-setup --enable

COPY debugging.R /content/

RUN R -f /content/debugging.R
```

Here is `debugging.R`:

```r
options(repos = c(CRAN = "https://cran.rstudio.com/"))

install.packages("digest")
install.packages("openssl")

# equivalent to rsconnect:::generateToken
generateToken <- function() {
  key <- openssl::rsa_keygen(2048L)
  priv.der <- openssl::write_der(key)
  pub.der <- openssl::write_der(key$pubkey)
  tokenId <- paste(c("T", openssl::rand_bytes(16)), collapse = "")

  list(
    token = tokenId,
    public_key = openssl::base64_encode(pub.der),
    private_key = openssl::base64_encode(priv.der)
  )
}

# equivalent to rsconnect:::rfc2616Date
rfc2616Date <- function(time = Sys.time()) {
  # set locale to POSIX/C to ensure ASCII date
  old <- Sys.setlocale("LC_TIME", "C")
  on.exit(Sys.setlocale("LC_TIME", old))

  strftime(Sys.time(), "%a, %d %b %Y %H:%M:%S GMT", tz = "GMT")
}

token <- generateToken()
path <- "/users/current"
method <- "GET"
date <- rfc2616Date()
md5 <- openssl::base64_encode(digest::digest("", algo = "md5", serialize = FALSE, raw = TRUE))
canonicalRequest <- paste(method, path, date, md5, sep = "\n")
[...]

PR #1054, opened 2024-03-18
sha1 token signing workaround

jonkeane:
Seeing the use of `digest` for md5 hashes made me realize that we could do the same for the sha1 hashs (and PKI for signing). 

I tested this publishing to connect using an existing token from RStudio and it worked just fine.

I also ran the tests in docker running rockylinux9 with FIPS mode enabled (using the docker file described in #928 and running `R CMD build . && R CMD check .`) There was [one test that fails](https://github.com/rstudio/rsconnect/blob/81b18c95c43d33cb395d6bef66b90d6aa5cc2075/tests/testthat/test-bundlePackage.R#L144-L145), though this looks like an encoding issue unless I'm missing something.

```
[59] "Running the tests in âtests/testthat.Râ failed."
[60] "Last 13 lines of output:"
[61] " 'test-appMetadata-quarto.R:54:3', 'test-appMetadata-quarto.R:64:3',"
[62] " 'test-appMetadata-quarto.R:74:3', 'test-appMetadata-quarto.R:90:3',"
[63] " 'test-appMetadata-quarto.R:115:3', 'test-appMetadata-quarto.R:134:3',"
[64] " 'test-appMetadata.R:14:3', 'test-deploySite.R:2:3'"
[65] " "
[66] " ââ Failed tests ââââââââââââââââââââââââââââââââââââââââââââââââââââââââââââââââ"
[67] " ââ Failure ('test-bundlePackage.R:145:3'): package_record works ââââââââââââââââ"
[68] " windows1251Record$Author (`actual`) not equal to \"Ð¡ÐµÑÐ³ÐµÐ¹ ÐÑÐ¸Ð½\" (`expected`)."
[69] " "
[70]" `actual`: \"\\xd1\\xe5\\xf0\\xe3\\xe5\\xe9 \\xc1\\xf0\\xe8\\xed\""
[71] " `expected`: \"Ð¡ÐµÑÐ³ÐµÐ¹ ÐÑÐ¸Ð½\" "
[72] " "
[73] " [ FAIL 1 | WARN 3 | SKIP 145 | PASS 631 ]"
```

We should still move away from sha1 (possibly by moving to default to API keys, which already work and avoid this issue), but this PR buys us some time to do that work (both here and on the connect side).


R/http.R
```diff
@@ -501,8 +501,14 @@ signatureHeaders <- function(authInfo, method, path, file = NULL) {
       der = TRUE
     )
 
-    # OpenSSL defaults to sha1 hash function (which is what we need)
-    rawsig <- openssl::signature_create(charToRaw(canonicalRequest), key = private_key)
+    # convert key into PKI format for signing, note this only accepts RSA, but
+    # that's what rsconnect generates already
+    pki_key <- PKI::PKI.load.key(strsplit(openssl::write_pem(private_key), "\n")[[1]], format = "PEM")
```
jonkeane, line 506: I'm sure there's a better way to do this, but I wanted to get feedback on this quicker

tests/testthat/test-http.R
```diff
@@ -22,8 +22,8 @@ test_that("authHeaders() picks correct method based on supplied fields", {
   )
 
   # Dummy key created with
-  # openssl::base64_encode(openssl::ed25519_keygen())
-  key <- "MC4CAQAwBQYDK2VwBCIEIDztfEgkp5CX7Jz0NCyrToaRW1L2tfmrWxNDgYyjO9bQ"
+  # openssl::base64_encode(openssl::rsa_keygen(2048L))
+  key <- readLines(test_path("fake-key"))
```
jonkeane, line 26: PKI doesn't support non-RSA keys. If I understand correctly this is totally fine because we only [generate RSA keys](https://github.com/rstudio/rsconnect/blob/81b18c95c43d33cb395d6bef66b90d6aa5cc2075/R/accounts.R#L150). 

Was the use of ed25519 just because it was shorter? 

I also committed these changes separately to be able to pull that and see if tests past there.

R/http.R
aronatkins: This is equivalent:

```r
pem <- openssl::write_pem(private_key)
pem_lines <- readLines(textConnection(pem))
pki_key <- PKI::PKI.load.key(pem_lines, format = "PEM")
```

tests/testthat/test-http.R
aronatkins: I'm guessing so, yes (shorter).

I would also be fine if the `fake-key` were a constant in-code rather than a file living alongside the tests. You could store it in PEM format, which has the start/end markers and (generally) shorter lines.

aronatkins (approved):
I feel like we should have a test proving that this implementation mirrors the old implementation. Right now, we're using the textual snapshot as a way of convincing ourselves, but it would be nice to have something that's enforced by code, as well.

It might mean refactoring `signatureHeaders` so we can pass the `canonicalRequest` into some `signRequest` helper, but that would let us test equivalence beyond the signature check. We could confirm that the "new" implementation returns the same signature as the old approach, which used openssl.

```r
signRequestPrivateKey_old <- function(private_key, canonicalRequest) {
  rawsig <- openssl::signature_create(charToRaw(canonicalRequest), key = private_key)
  openssl::base64_encode(rawsig)
}

signRequestPrivateKey_new <- function(private_key, canonicalRequest) {
  pem <- openssl::write_pem(private_key)
  key <- readLines(textConnection(pem))
  pki_key <- PKI::PKI.load.key(key, format = "PEM")
  digested <- digest::digest(charToRaw(canonicalRequest), "sha1", serialize = FALSE, raw = TRUE)
  rawsig <- PKI::PKI.sign(key = pki_key, digest = digested)
  openssl::base64_encode(rawsig)
}
```

aronatkins:
Forgot to mention: This needs NEWS.

jonkeane:
Thanks for the detailed review. I'm broadly onboard with the suggestions and will get to doing them shortly. But if this ends up blocking anything, (as always) I don't mind if you pushed to this branch to unblock it.

tests/testthat/test-http.R
jonkeane: Thanks, I pulled the PEM format into the test. It's not the prettiest, but the sidecar file also isn't either.

jonkeane:
This is ready for review again

aronatkins (approved):
Thanks for that additional test.

jonkeane:
> Thanks for that additional test.

Of course. Oh and I should have mentioned this FTR in the previous comment: I wanted to split that test out into it's own `test_that` block (so it's clear what it's testing, but didn't want to also have to move the key elsewhere or duplicate it, so tacked it on as the least-disruptive option.

aronatkins:
> move the key elsewhere

No worries. You could create a helper function adjacent to the `test_that` block, if that helps separate things. That helper could return something that resembles a "token", so you have both the public and private keys.

```r
generateTokenFixed <- function() {
    return list(
        token = "TDECAFBAD",
        public_key = "...",
        private_key = "...",
    )
}
test_that("thing one", {
    token <- generateTokenFixed()
    # check for stable signature (snapshot)
})
test_that("thing two", {
    token <- generateTokenFixed()
    # check for equivalent signature (openssl)
})
```

I don't think this adjustment is necessary, though. If we were to continue to expand the tests in this area, it's a natural next step.

venpopov:
Unfortunately (re)-introducing the PKI dependency makes it impossible to install rsconnect on MacOS when R is installed from homebrew rather than CRAN: https://github.com/rstudio/rsconnect/issues/1073

PR #1055, opened 2024-03-22
quarto: more flexible discovery

aronatkins:
* script.R with _quarto.yml is identified as quarto-static.
* script.R without _quarto.yml is identified as quarto-static assuming it is not identified as some other content type.
* script.R can be identified as the primary quarto-static document.
* bundling only renames the primary script.R for shiny content.

A rendered script `script.R` might look like:

````r
#' ---
#' title: cats are cats
#' ---

cat("i am a cat\n")
````

It may or may not have an accompanying `_quarto.yml` file. A simple version of the Quarto configuration might look like:

````yaml
project:
    type: default
````

* A directory containing this `script.R` is identified as quarto-static by `rsconnect::writeManifest()` and `rsconnect::deployApp()`.
* A directory containing this `script.R` and `_quarto.yml` is identified as quarto-static by `rsconnect::writeManifest()` and `rsconnect::deployApp()`.

## Other changes

This work also necessitated a change to how we bundle content. In particular, RStudio is able to run and deploy Shiny single-file applications which are contained in files _not_ named `app.R`. For example, if we create a project containing the default Geyser app and save it to `notapp.R` rather than `app.R`, RStudio can deploy that file. It is able to deploy the Shiny application because `notapp.R` is renamed to `app.R` when the content is bundled.

The action taken by RStudio is effectively:

```r
rsconnect::deployApp(appPrimaryDoc = "notapp.R")
```

This forces rsconnect to treat the content as a Shiny application and also forces the rename of `notapp.R` to `app.R`.

Prior to this change, rsconnect would rewrite _any_ `appPrimaryDoc=*.R` to `app.R` regardless of content type. Unfortunately, `appPrimaryDoc` is computed internally and used by Quarto content. Because that primary asset was an R script, bundling renamed it. To avoid this, the rename is now specific to Shiny content.

Test that RStudio can deploy a Shiny application in a `notapp.R` file. Download that deployed bundle and see that the application has been renamed to `app.R`.

Other types of content (e.g., Plumber APIs, Quarto rendered scripts) do not have their R scripts renamed when deployed.

### DIFF
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -123,7 +123,11 @@ appDependencies <- function(appDir = getwd(),
     ))
   }
 
-  bundleDir <- bundleAppDir(appDir, appFiles)
+  bundleDir <- bundleAppDir(
+    appDir = appDir,
+    appFiles = appFiles,
+    appMode = appMetadata$appMode
+  )
   defer(unlink(bundleDir, recursive = TRUE))
 
   extraPackages <- inferRPackageDependencies(appMetadata)
--- a/R/appMetadata.R
+++ b/R/appMetadata.R
@@ -29,7 +29,10 @@ appMetadata <- function(appDir,
 
   if (is.null(appMode)) {
     # Generally we want to infer appPrimaryDoc from appMode, but there's one
-    # special case
+    # special case: RStudio provides appPrimaryDoc when deploying Shiny
+    # applications. They may have name.R, not app.R or server.R.
+    #
+    # This file is later renamed to app.R when deployed by bundleAppDir().
     if (!is.null(appPrimaryDoc) &&
           tolower(tools::file_ext(appPrimaryDoc)) == "r") {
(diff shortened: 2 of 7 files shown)


toph-allen (approved):
Looks good to me.

Out of curiosity: I have only been loosely following the Connect-side changes for Quarto R scripts. I know we have kicked around the idea of an "R script" content type; is this substantially different for how we think that'd function? I'm just idly curious about the use cases for this, but that's totally unrelated to this PR looking good at a read-through.

aronatkins:
> Out of curiosity: I have only been loosely following the Connect-side changes for Quarto R scripts. I know we have kicked around the idea of an "R script" content type; is this substantially different for how we think that'd function?

This uses Quarto to run the script because that gives us very opinionated ways of capturing output, attaching resources, etc. This may not be enough for some folks, but is much less invasive than rewriting an R script into an R Markdown document, for example.

PR #1070, opened 2024-05-13
deployTFModel and writeManifest support TensorFlow

aronatkins:
Fixes https://github.com/rstudio/connect/issues/26885

### DIFF
--- a/R/appMetadata.R
+++ b/R/appMetadata.R
@@ -38,9 +38,9 @@ appMetadata <- function(appDir,
       appMode <- "shiny"
     } else {
       # Inference only uses top-level files
-      rootFiles <- appFiles[dirname(appFiles) == "."]
       appMode <- inferAppMode(
-        file.path(appDir, rootFiles),
+        appDir,
+        appFiles,
         usesQuarto = quarto,
         isShinyappsServer = isShinyappsServer
       )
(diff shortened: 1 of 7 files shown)


R/appMetadata.R
```diff
@@ -98,10 +98,18 @@ checkAppLayout <- function(appDir, appPrimaryDoc = NULL) {
...
+# anywhere in the hierarchy.
+inferAppMode <- function(
+  appDir,
+  appFiles,
+  usesQuarto = NA,
+  isShinyappsServer = FALSE) {
+
+  rootFiles <- appFiles[dirname(appFiles) == "."]
+  absoluteRootFiles <- file.path(appDir, rootFiles)
+  absoluteAppFiles <- file.path(appDir, appFiles)
```
toph-allen, line 112: Weird that TensorFlow files can be anywhere in the hierarchy! I guess for those, you may be using R in any working dir and are just pointing to a TF file? Maybe not weird, just different.

toph-allen (approved):
Looks good to me.

R/appMetadata.R
aronatkins: It's not "anywhere", but it's not at the root. For example, this is a "model" directory with an accompanying `manifest.json` at the root.

```
model
model/1
model/1/variables
model/1/variables/variables.data-00000-of-00001
model/1/variables/variables.index
model/1/saved_model.pb
model/manifest.json
```

PR #1071, opened 2024-05-16
Add access to URL on deploy

jonkeane:
Adds `/access` to the URL at the end of a successful deployment so that the person publishing has easy access to sharing what they just published.

There is one open question if doing this in the publishers versus changing the API. If we do this in the client I'll open a matching PR to `rsconnect-python` as well.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -686,7 +686,8 @@ openURL <- function(client, application, launch.browser, on.failure, deploymentS
   # Check to see if we should open config url or app url
   if (!is.null(client$configureApplication)) {
     config <- client$configureApplication(application$id)
-    url <- config$config_url
+    # Direct the publisher to the access pane to share
+    url <- paste0(config$config_url, "/access")
     if (!deploymentSucceeded && validURL(config$logs_url)) {
       # With 1.5.5+, Connect application configuration includes
       # a logs URL to be shown on unsuccessful deployment.


R/deployApp.R
```diff
@@ -686,7 +686,8 @@ openURL <- function(client, application, launch.browser, on.failure, deploymentS
   # Check to see if we should open config url or app url
   if (!is.null(client$configureApplication)) {
     config <- client$configureApplication(application$id)
-    url <- config$config_url
+    # Direct the publisher to the access pane to share
+    url <- paste0(config$config_url, "/access")
```
jonkeane, line 690: Alternatively, we could change the API that produces `config_url` here which would feed all of the publishing clients the same. But that seems more drastic and I'm not sure if anything other than publishing clients use this part of the API.

nealrichardson: I think this is the right change, this feels like a client-side concern. But if you were worried that we might decide differently in the future, you could wrap this in something like `if (!endsWith(config$config_url, "/access"))`

aronatkins: This URL is produced by Connect and has changed over time as the dashboard and other URLs have been adjusted. I'll send references externally.

aronatkins: My preference is to have the server declare where a publisher should be directed after deployment. That lets multiple Connect versions communicate different URLs to the clients. Assume, for example, that we rewrite the content settings tabs and ditch `/access` in favor of some other route. Old Connect would want `/access` while new Connect wants `/something-else`.

PR #1087, opened 2024-06-26
tasks() and taskLog() are only for shinyapps.io and posit.cloud

aronatkins:
While auditing the Connect APIs used by rsconnect, noticed that some paths are not APIs, which led to this change.

* trim listTasks() and killTask() from the Connect client - non-APIs
* trim getTask() from the Connect client - waitForTask is used

### DIFF
--- a/R/client-connect.R
+++ b/R/client-connect.R
@@ -102,23 +102,6 @@ connectClient <- function(service, authInfo) {
       GET(service, authInfo, paste0("/applications/", applicationId))
     },
 
-    ## Tasks API
-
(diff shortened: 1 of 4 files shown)


toph-allen (approved):
Looks good. Were these never in the API, or were they deprecated?

aronatkins:
> Looks good. Were these never in the API, or were they deprecated?

The "kill task" API existed as a route from ~2014 until 1.7.2, but had an empty implementation and not called. The "list tasks" API was never a route, as far as I can tell.

PR #1100, opened 2024-08-29
Feature/add deployment timestamp

GCRev:
See: https://github.com/rstudio/rstudio-pro/issues/5597

We have this provisioning on the Workbench homepage to display the timestamp (unix epoch in seconds) for when you recently published something via rsconnect. For whatever reason, there are no timestamps added to the history entries in the `history.dcf` file that Workbench reads. This fixes that missing timestamp without requiring any changes on the Workbench side.

image

### DIFF
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -132,7 +132,7 @@ deploymentRecord <- function(name,
                              bundleId = NULL,
                              url = NULL,
                              version = deploymentRecordVersion,
-                             timestamp = NULL,
+                             when = NULL,
                              metadata = list()) {
 
   check_character(envVars, allow_null = TRUE)
@@ -148,7 +148,7 @@ deploymentRecord <- function(name,
     appId = appId %||% "",
     bundleId = bundleId %||% "",
     url = url %||% "",
-    timestamp = timestamp %||% format(as.numeric(as.POSIXct(Sys.time())) * 1000, scientific = FALSE),
+    when = when %||% as.numeric(as.POSIXct(Sys.time())),
     version = version
   )
   # convert any multi-value metadata entries into comma-separated values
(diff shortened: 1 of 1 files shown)


R/deployments.R
```diff
@@ -164,6 +166,7 @@ writeDeploymentRecord <- function(record, filePath) {
 addToDeploymentHistory <- function(appPath, deploymentRecord) {
   # add the appPath to the deploymentRecord
```
aronatkins, line 167: `deploymentFields` needs updating.

R/deployments.R
```diff
@@ -147,6 +148,7 @@ deploymentRecord <- function(name,
     appId = appId %||% "",
     bundleId = bundleId %||% "",
     url = url %||% "",
+    when = when %||% as.numeric(as.POSIXct(Sys.time())),
```
aronatkins, line 151: Please add a test confirming that we read the same `when` data that was written.

aronatkins (commented):
Do you think this is worth mentioning in NEWS?

aronatkins:
The `when` field was removed here: https://github.com/rstudio/rsconnect/pull/770 (@hadley)

hadley:
Here's the news bullet for context:

* Deployment records no longer contain the time the app was deployed (`when`)
  or when it's metadata was last synched (`lastSyncTime`) as these variables
  are not very useful, and they lead to uninteresting diffs if you have 
  committed the deployment records to git (#770).

IMO it's not worth adding back just for that display in workbench; if it wants that metadata it should talk to the connect API.

GCRev:
🤔 In that case I'll just remove the display field from Workbench until we work on a larger feature to obtain the history information from the connect API (which I haven't looked into yet) rather than reading a file -- if we're getting one field from an API we might as well get all of them.

The Workbench code predates any context of mine, so I assume it was probably the easiest short-term "integration" since the `rworkspaces` process already has access to the user's local files, so it can read this `history.dcf` "for free" off the file system without needing to track and manage other credentials. I suppose we already have an integration in the IDE, but I also don't know anything about that at the moment.

I'll open up a feature request for this, since I don't think it will be as trivial as a bug fix.

PR #1139, opened 2025-02-24
isDocumentPath no longer treats directories as documents

### LINKED ISSUE #1138: `rsconnect` doesn't correctly detect if it's publishing a single file, or a directory for `recordDir` location
Apparently, `rsconnect` incorrectly determines whether your project/Shiny application is a directory vs a file based on the R package `tools::file_ext() ` here solely based on the existence of any "extension". The issue is that a naming convention of `framework.appname`, ex: `shiny.awesomeapp`. Since `rsconnect` is only checking for the existence of an extension, the `recordDir /rsconnect` is being placed in the parent directory of their application rather than alongside the application in their `workingDir`. This results in two negative outcomes, 

1. They are put in the strange position of needing to copy this file manually into their `shiny.awesomeapp` directory in order to keep the `shiny.dcf` and their application together in version control. 
2.  It results in any future collaborators who clone their repository needing to copy this directory out of `shiny.awesomeapp` one level higher in the directory structure in order to republish the application. Additionally, the Publish button in RStudio doesn't show Republish when the `rsconnect` directory is in the current working directory of `shiny.awesomeapp` .

The fix to this issue may be as simple as changing the logic here:

https://github.com/rstudio/rsconnect/blob/a3ebf70adb96a020fb8844b7194870ae04500a1d/R/config.R#L134

```
#Current Logic: 
> isDocumentPath <- function(path) {
+   tools::file_ext(path) != ""
+ }
> isDocumentPath("/home/sam.cofer/shiny.test")
[1] TRUE
> deploymentConfigDir("/home/sam.cofer/shiny.test")
[1] "/home/sam.cofer/rsconnect/documents/shiny.test"
> isDocumentPath <- function(path) {
+   !fs::is_dir(path)
+ }
> isDocumentPath("/home/sam.cofer/shiny.test")
/home/sam.cofer/shiny.test 
                     FALSE 
> deploymentConfigDir("/home/sam.cofer/shiny.test")
[1] "/home/sam.cofer/shiny.test/rsconnect"
```

With this example, we can see that this change correctly identifies this directory as a directory, rather than a file. I'm not sure if there are any other downstream repercussions, but this change seems like it would better align with how client OS's detect files.

aronatkins:
Additionally, remove the uncalled defaultAppName, which removes one use of isDocumentPath.

fixes #1138

### DIFF
--- a/R/deploymentTarget.R
+++ b/R/deploymentTarget.R
@@ -315,32 +315,6 @@ updateDeployment <- function(previous, appTitle = NULL, envVars = NULL) {
   )
 }
 
-defaultAppName <- function(recordPath, server = NULL) {
-  if (isDocumentPath(recordPath)) {
-    name <- file_path_sans_ext(basename(recordPath))
-    if (name == "index") {
-      # parent directory will give more informative name
-      name <- basename(dirname(recordPath))
-    } else {
-      # deploying a document
-    }
-  } else {
-    # deploying a directory
-    name <- basename(recordPath)
-  }
-
-  if (isShinyappsServer(server)) {
-    # Replace non-alphanumerics with underscores, trim to length 64
-    name <- tolower(gsub("[^[:alnum:]_-]+", "_", name, perl = TRUE))
-    name <- gsub("_+", "_", name)
-    if (nchar(name) > 64) {
-      name <- substr(name, 1, 64)
-    }
-  }
-
-  name
-}
-
 shouldUpdateApp <- function(application,
                             uniqueName,
                             forceUpdate = FALSE,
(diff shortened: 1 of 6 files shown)


samcofer (approved):
LGTM!

R/deploymentTarget.R
```diff
@@ -315,32 +315,6 @@ updateDeployment <- function(previous, appTitle = NULL, envVars = NULL) {
   )
 }
 
-defaultAppName <- function(recordPath, server = NULL) {
```
toph-allen, line 318: Removed because it wasn't called anywhere, right?

toph-allen (approved):
Looks good to me!

R/deploymentTarget.R
aronatkins: Yes, that's right. The PR description noted its removal.

PR #1141, opened 2025-02-28
Add `*.lic` to `.gitignore`

jonkeane:
Adding `*.lic` to `.gitignore` so that accidentally committing license files requires additional steps.


.gitignore
```diff
@@ -9,3 +9,6 @@ docs/
 .rscignore
 tags
 inst/doc
+
+# license files should not be commited to this repository
+*.lic
```
aronatkins, line 14:
```suggestion
*.lic

```

Check that your editor adds end-of-file newlines.

aronatkins (approved):

aronatkins:
No objections. The risk here is low, as no rsconnect tests install products.

PR #1144, opened 2025-03-25
Add support for _server.yml

thomasp85:
This small PR adds support for recognising the existence of `_server.yml` as a signal that the project is an API type

I believe this is all that is needed but if I have missed anything please let me know


aronatkins:
@toph-allen - I've pushed up tests and a NEWS update. Let me know if you need anything additional as you help @thomasp85.

toph-allen (approved):
Nice and concise — looks good to me! I don't think anything else is required here.

PR #1146, opened 2025-03-31
require an executable engine when quarto doc indicates a server

### LINKED ISSUE #1145: quarto: require knitr when using a server
````markdown
---
title: "not so shiny"
server: shiny
---
````

When a Quarto document states that it wants to use Shiny, `quarto inspect` must indicate that the content uses the `knitr` engine.

### PR DESCRIPTION
**Pull request by aronatkins**
@toph-allen - this error handling feels like it's in an arbitrary location; any opinions about its placement?

fixes #1145

### DIFF
--- a/R/appMetadata.R
+++ b/R/appMetadata.R
@@ -67,6 +67,14 @@ appMetadata <- function(appDir,
       appDir = appDir,
       appPrimaryDoc = appPrimaryDoc
     )
+    if (appMode == "quarto-shiny") {
+      if (!"knitr" %in% quartoInfo[["engines"]]) {
+        cli::cli_abort(c(
+          "The Quarto document requires a server but does not use an executable engine.",
+          "Consider including some executable code, specifying an engine, or removing the server configuration."
+        ))
+      }
+    }
   } else {
     quartoInfo <- NULL
   }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: R/appMetadata.R [hunk 1]
```diff
@@ -67,6 +67,14 @@ appMetadata <- function(appDir,
...
       appPrimaryDoc = appPrimaryDoc
     )
+    if (appMode == "quarto-shiny") {
+      if (!"knitr" %in% quartoInfo[["engines"]]) {
+        cli::cli_abort(c(
+          "The Quarto document requires a server but does not use an executable engine.",
+          "Consider including some executable code, specifying an engine, or removing the server configuration."
+        ))
+      }
+    }
```
toph-allen, line 77: @aronatkins Hmm. Its placement _does_ feel a little arbitrary, I see what you mean. And I don't see that many app-mode-specific errors here. But this _is_ where the requisite info is gathered, and just glancing over other `cli_abort` calls in the code, I don't see an overarching strategy here, so… I think it's probably fine.

toph-allen: Could a `quarto-shiny` file specify an engine other than `knitr`, e.g. Python? The error talks about needing to use "an executable engine" but the code checks specifically for `knitr`.

toph-allen (approved):
Generally looks good; just one question. If the answer to that question doesn't imply changes, feel free to merge.

R/appMetadata.R
aronatkins: Good question. Yes, they may use the jupyter or knitr engines.

I've expanded the check to allow the jupyter engine, but am not making any changes to the downstream logic. rsconnect generally only supports the R engine for Shiny Quarto documents.

PR #1150, opened 2025-04-08
spcs support

edavidaja:
- **curl 43??**
- **keep whitespace in snowflake token**
- **store connection name, refresh on demand**

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,10 +45,11 @@ Suggests:
     reticulate,
     rmarkdown (>= 1.1),
     shiny,
+    snowflakeauth,
     testthat (>= 3.1.9),
     webfakes,
     withr
-VignetteBuilder: 
+VignetteBuilder:
     knitr, rmarkdown
 Config/Needs/website: tidyverse/tidytemplate
 Config/testthat/edition: 3
@@ -56,3 +57,4 @@ Config/testthat/parallel: true
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.1
+Remotes: atheriel/snowflakeauth
--- a/R/accounts.R
+++ b/R/accounts.R
@@ -76,27 +76,11 @@ connectSPCSUser <- function(account = NULL, server = NULL, snowflake_connection_
   server <- findServer(server)
   user <- getSPCSAuthedUser(server, snowflake_connection_name)
 
-  ingressURL <- serverInfo(server)$url
-  parsedIngress <- parseHttpUrl(ingressURL)
-
-  if (!is.null(snowflake_connection_name)) {
-    token <- snowflakeauth::snowflake_credentials(
-      snowflakeauth::snowflake_connection(snowflake_connection_name),
-      spcs_endpoint = parsedIngress$host
-    )
-  }
-
   registerAccount(
     serverName = server,
     accountName = account %||% user$username,
-    snowflakeToken = token$Authorization,
-    accountId = user$id
-  )
-
-  account <- list(
-    server = server,
-    snowflake_connection_name = snowflake_connection_name,
-    snowflakeToken = token$Authorization
+    accountId = user$id,
+    snowflakeConnectionName = snowflake_connection_name
   )
 
   if (!quiet) {
@@ -109,19 +93,10 @@ connectSPCSUser <- function(account = NULL, server = NULL, snowflake_connection_
 
 getSPCSAuthedUser <- function(server, snowflake_connection_name) {
 
-  ingressURL <- serverInfo(server)$url
-  parsedIngress <- parseHttpUrl(ingressURL)
-
-  if (!is.null(snowflake_connection_name)) {
-    token <- snowflakeauth::snowflake_credentials(
-      snowflakeauth::snowflake_connection(snowflake_connection_name),
-      spcs_endpoint = parsedIngress$host
-    )
-  }
-
+  serverAddress <- serverInfo(server)
   account <- list(
     server = server,
-    snowflakeToken =  token$Authorization
+    snowflakeToken = getSnowflakeAuthToken(serverAddress$url, snowflake_connection_name)
   )
 
   client <- clientForAccount(account)
@@ -392,7 +367,7 @@ registerAccount <- function(serverName,
                             secret = NULL,
                             private_key = NULL,
                             apiKey = NULL,
-                            snowflakeToken = NULL)
+                            snowflakeConnectionName = NULL)
                             {
 
   check_string(serverName)
@@ -409,12 +384,12 @@ registerAccount <- function(serverName,
     secret = secret,
     private_key = private_key,
     apiKey = apiKey,
-    snowflakeToken = snowflakeToken
+    snowflakeConnectionName = snowflakeConnectionName
   )
 
   path <- accountConfigFile(accountName, serverName)
   dir.create(dirname(path), recursive = TRUE, showWarnings = FALSE)
-  write.dcf(compact(fields), path, width = 100, keep.white = c("snowflakeToken"))
+  write.dcf(compact(fields), path, width = 100)
 
   # set restrictive permissions on it if possible
   if (identical(.Platform$OS.type, "unix"))
--- a/R/accounts.R
+++ b/R/accounts.R
@@ -77,10 +77,12 @@ connectSPCSUser <- function(account = NULL, server = NULL, snowflake_connection_
   user <- getSPCSAuthedUser(server, snowflake_connection_name)
 
   ingressURL <- serverInfo(server)$url
+  parsedIngress <- parseHttpUrl(ingressURL)
+
   if (!is.null(snowflake_connection_name)) {
     token <- snowflakeauth::snowflake_credentials(
       snowflakeauth::snowflake_connection(snowflake_connection_name),
-      spcs_endpoint = httr2::url_parse(ingressURL)$hostname
+      spcs_endpoint = parsedIngress$host
     )
   }
 
@@ -106,12 +108,14 @@ connectSPCSUser <- function(account = NULL, server = NULL, snowflake_connection_
 }
 
 getSPCSAuthedUser <- function(server, snowflake_connection_name) {
+
   ingressURL <- serverInfo(server)$url
+  parsedIngress <- parseHttpUrl(ingressURL)
 
   if (!is.null(snowflake_connection_name)) {
     token <- snowflakeauth::snowflake_credentials(
       snowflakeauth::snowflake_connection(snowflake_connection_name),
-      spcs_endpoint = httr2::url_parse(ingressURL)$hostname
+      spcs_endpoint = parsedIngress$host
     )
   }
 
@@ -410,7 +414,7 @@ registerAccount <- function(serverName,
 
   path <- accountConfigFile(accountName, serverName)
   dir.create(dirname(path), recursive = TRUE, showWarnings = FALSE)
-  write.dcf(compact(fields), path, width = 100)
+  write.dcf(compact(fields), path, width = 100, keep.white = c("snowflakeToken"))
 
   # set restrictive permissions on it if possible
   if (identical(.Platform$OS.type, "unix"))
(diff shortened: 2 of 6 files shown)


R/http.R
```diff
@@ -142,7 +142,12 @@ handleResponse <- function(response, error_call = caller_env()) {
   if (isContentType(response$contentType, "application/json")) {
     # parse json responses
     if (nzchar(response$content)) {
-      json <- jsonlite::fromJSON(response$content, simplifyVector = FALSE)
+      json <- tryCatch({
+        jsonlite::fromJSON(response$content, simplifyVector = FALSE)
+      }, error = function(e) {
+
+        return(response$content)
+      } )
```
edavidaja, line 150:
```suggestion
      })
```

R/import-standalone-config.R
```diff
@@ -0,0 +1,241 @@
...
+#' }
+#' @examples
+#' # Pass connection parameters manually, which is useful if there is no
+#' # connections.toml file. For example, to use key-pair authentication:
+#' conn <- snowflake_connection(
+#'   account = "myaccount",
+#'   user = "me",
+#'   private_key = "rsa_key.p8"
+#' )
+#' @export
```
aronatkins, line 73: We do not want these names to become exported by rsconnect.

We have similar files from rlang, but those don't end up as exported.
https://github.com/rstudio/rsconnect/blob/main/R/import-standalone-obj-type.R

R/accounts.R
```diff
@@ -72,6 +76,40 @@ connectApiUser <- function(account = NULL, server = NULL, apiKey, quiet = FALSE)
   invisible()
 }
 
+
+#' @rdname connectApiUser
+#' @export
```
aronatkins, line 81: You can use `@inheritParams` so you can document `snowflakeConnectionName` here, while still avoid duplicating the other arguments. See `deployDoc` for an example.

R/accounts.R
```diff
@@ -72,6 +76,40 @@ connectApiUser <- function(account = NULL, server = NULL, apiKey, quiet = FALSE)
...
+connectSPCSUser <- function(account = NULL, server = NULL, snowflakeConnectionName, quiet = FALSE) {
```
aronatkins, line 82: It is unexpected to have arguments-with-defaults before a required argument. Should the connection name come first?

R/http.R
```diff
@@ -142,7 +142,12 @@ handleResponse <- function(response, error_call = caller_env()) {
   if (isContentType(response$contentType, "application/json")) {
     # parse json responses
     if (nzchar(response$content)) {
-      json <- jsonlite::fromJSON(response$content, simplifyVector = FALSE)
+      json <- tryCatch({
+        jsonlite::fromJSON(response$content, simplifyVector = FALSE)
+      }, error = function(e) {
+
+        return(response$content)
```
aronatkins, line 149: When is a response that claims to be `application/json` not JSON? Why is that no longer an error?

R/http.R
```diff
@@ -460,7 +465,10 @@ authHeaders <- function(authInfo, method, path, file = NULL) {
     signatureHeaders(authInfo, method, path, file)
   } else if (!is.null(authInfo$apiKey)) {
     list(`Authorization` = paste("Key", authInfo$apiKey))
-  } else {
+  } else if (!is.null(authInfo$snowflakeToken)) {
+    # snowflakeauth returns a list
```
aronatkins, line 469:
```suggestion
    # snowflakeauth returns a list of named header values
```

R/keypair.R
```diff
@@ -0,0 +1,92 @@
+# Support for Snowflake key-pair authentication for SPCS and the REST API.
+keypair_credentials <- function(
+  account,
```
aronatkins, line 3: Consider using `snowflake_` prefixes for some of these argument names to remind folks that this is not the rsconnect account name used elsewhere.

R/keypair.R
```diff
@@ -0,0 +1,92 @@
+# Support for Snowflake key-pair authentication for SPCS and the REST API.
+keypair_credentials <- function(
```
aronatkins, line 2: Consider using `snowflake_` prefixes for any function that only applies to snowflake.

R/keypair.R
```diff
@@ -0,0 +1,92 @@
...
+  # algorithm than Snowflake does.
+  fp <- openssl::base64_encode(openssl::sha256(openssl::write_der(key$pubkey)))
+  sub <- toupper(paste0(account, ".", user))
+  # Note: Snowflake employs a malformed issuer claim, so we have to inject it
+  # manually after jose's validation phase.
+  claim <- jose::jwt_claim(
+    iss = "dummy",
+    sub = sub,
+    iat = iat,
+    # Expire in 5 minutes.
```
aronatkins, line 53: What happens when we hit expiration? Would a "long" deploy hit this?

R/keypair.R
```diff
@@ -0,0 +1,92 @@
...
+      # distinguish them from OAuth access tokens.
+      `X-Snowflake-Authorization-Token-Type` = "KEYPAIR_JWT"
+    ))
+  }
+  account_url <- sprintf("https://%s.snowflakecomputing.com", account)
+  token <- exchange_jwt_for_token(account_url, jwt, spcs_endpoint, role)
+  # Yes, this is actually the format of the Authorization header that SPCS
+  # requires.
+  return(list(
+    Authorization = sprintf('Snowflake Token="%s"', token)
```
aronatkins, line 26: These are my favorite two quotes of all time. So inspirational! 😬

aronatkins (commented):
Let's avoid the additional exported functions. Everything else looks good.

aronatkins:
Oh. Let's also make sure that we update NEWS.

R/accounts.R
edavidaja: It certainly could; I was following the pattern from:

https://github.com/rstudio/rsconnect/blob/11f0434975858f1b6a81d85e6b5bd8d47611b4fd/R/accounts.R#L57

R/http.R
edavidaja: Maybe it would be preferable to condition this behavior on responses from snowflake hosts only?

R/accounts.R
aronatkins: How about that. OK, then!

R/http.R
aronatkins: Could you explain the situation where this occurs?

edavidaja: The access token returned by the exchange is just the response body, [rather than the JSON message the spec would lead you to expect.](https://github.com/posit-dev/snowflakeauth/blob/fa679d898f151f766c20f97e0ff5188ed4b99ef3/R/keypair.R#L90-L94)

DESCRIPTION
```diff
@@ -32,7 +32,9 @@ Imports:
     rlang (>= 1.0.0),
     rstudioapi (>= 0.5),
     tools,
-    yaml (>= 2.1.5)
+    yaml (>= 2.1.5),
+    RcppTOML,
```
nealrichardson, line 36: We don't seem to be importing from these new dependencies, should we? 

Alternatively, could they instead go in `Suggests` to make them optional?

R/import-standalone-config.R
```diff
@@ -0,0 +1,241 @@
...
+snowflake_connection <- function(name = NULL, ..., .config_dir = NULL) {
+  params <- list(name = name)
+  config_dir <- .config_dir %||% default_config_dir()
+  cfile <- file.path(config_dir, "connections.toml")
+
+  # TODO: How should we distinguish "no ambient Snowflake credentials" here?
+
+  if (file.exists(cfile)) {
+    connections <- RcppTOML::parseTOML(cfile, fromFile = TRUE)
```
nealrichardson, line 82: If you were to move `RcppTOML` to Suggests, you'd want to `requireNamespace` or similar here to check that it exists (or `rlang` has a nicer helper for it I think?)

R/keypair.R
```diff
@@ -0,0 +1,92 @@
...
+  ))
+}
+
+# Generate a JWT that can be used for Snowflake "key-pair" authentication.
+generate_jwt <- function(account, user, private_key, iat = NULL, jti = NULL) {
+  check_installed("jose", "for key-pair authentication")
```
nealrichardson, line 32: Ah, you've already set up `jose` as an optional dependency. So it should be in `Suggests` not `Imports`, and if that makes sense here, it probably makes sense for `RcppTOML` too.

R/servers.R
```diff
@@ -154,6 +159,8 @@ findServer <- function(server = NULL,
 #'   character vector containing the certificate's contents.
 #' @param validate Validate that `url` actually points to a Posit Connect
 #'   server?
+#' @param snowflakeConnectionName Name for the Snowflake connection parameters
```
nealrichardson, line 162: Do we want to have a company's name baked into a function argument like this? (Would we expect to do this for AWS, Azure, etc.?) I don't have a ready alternative to suggest, but I wonder if there's a different way to propagate the information.

nealrichardson (commented):
Just checking in to see how this was going and had a couple of thoughts

DESCRIPTION
aronatkins: @nealrichardson - We had discussed this and decided to take these as `Imports` in an attempt to avoid additional friction when folks publish.

R/http.R
aronatkins: Oh, that's really unfortunate. As this is only for a single call path (to obtain the access token), is it worth having a separate HTTP helper for that?

Another idea: add an argument to the `POST` call indicating that a "raw" response should be returned. That would be passed down to `handleResponse` and this code becomes:

```r
if (rawResponse) {
    if (response$status %in% 200:399) {
      out <- response$content
    } else {
      reportError(response$content)
    }
} else if (isContentType(response$contentType, "application/json)) {
  # existing code
} # ...
```

This basically duplicates the final clause, but for a different reason.

R/accounts.R
mconflitti-pbc: 😮 how does that even work?

if you called connectApiUser("some-key"), does it just make a delineation between named and unnamed params?

R/servers.R
aronatkins: The snowflake connection name indicates the named connection in an individual user's configuration file, which is not the same as the URL.

```toml
[connections.production]
account = "deadbeef-bob-the-builder-com"
```

Here, the connection name is "production". The account value has some bearing on a component of the URL, but I believe that multiple connections could reference the same account (other portions of the configuration would vary). The target URL might be `https://decafbad-deadbeef-bob-the-builder-com.snowflakecomputing.app/` for customer "bob the builder dot com".

R/accounts.R
aronatkins: It's awkward, for sure. You need to use named arguments when the earlier positional arguments aren't provided.

```r
doit <- function(a=42, b) { paste("a",a,"b",b) }
doit(13)
#> Error in doit(13) : argument "b" is missing, with no default
doit(b=13)
#> [1] "a 42 b 13"
doit(a=101,b=17)
#> [1] "a 101 b 17"
```

R/accounts.R
```diff
@@ -72,6 +72,52 @@ connectApiUser <- function(account = NULL, server = NULL, apiKey, quiet = FALSE)
...
+#' `connectSPCSUser()` connects your Posit Connect account to the rsconnect
+#' package so it can deploy and manage applications on your behalf.
+#' Configure a
+#' [`connections.toml` file](https://docs.snowflake.com/en/developer-guide/snowflake-cli/connecting/configure-cli#location-of-the-toml-configuration-fil)
+#' in the appropriate location.
+#'
+#'
+#' @inheritParams connectApiUser
+#' @param snowflakeConnectionName Name for the Snowflake connection parameters
+#'   stored in `config.toml` or `connections.toml`.
```
christierney, line 88: I might be misreading but I think this is potentially confusing. We will check `config.toml` for the default connection name but not for any other connection info, right? i.e. if you don't have a `connections.toml`, only a `config.toml` with all the connection info, that won't work?

christierney (approved):

R/http.R
aronatkins: The introduction of `rawResponse = TRUE` from `GET_server_settings` appears to have caused https://github.com/rstudio/rsconnect/issues/1166.

PR #1154, opened 2025-04-30
air: formatting

aronatkins:
Ran the commands:

```bash
air format R
air format tests
```

Ran the call:

```r
usethis::use_github_action(url = "https://github.com/posit-dev/setup-air/blob/main/examples/format-suggest.yaml")
```

Reviewed all of the edits and then manually converted a number of `if (expr) foo` to `if (expr) {foo}` so air would flow statement across multiple lines. The original reformatting was especially poor when the statement had if+else or if+if-else+else.

### DIFF
--- a/R/account-find.R
+++ b/R/account-find.R
@@ -26,11 +26,8 @@ findAccount <- function(
   }
 
   if (!is.null(accountName) && !is.null(server)) {
-    theseAccounts <- accounts[
-      accounts$server == server & accounts$name == accountName,
-      ,
-      drop = FALSE
-    ]
+    selected <- accounts$server == server & accounts$name == accountName
+    theseAccounts <- accounts[selected, , drop = FALSE]
     if (nrow(theseAccounts) == 0) {
       cli::cli_abort(
         c(
@@ -41,7 +38,8 @@ findAccount <- function(
       )
     }
   } else if (is.null(accountName) && !is.null(server)) {
-    theseAccounts <- accounts[accounts$server == server, , drop = FALSE]
+    selected <- accounts$server == server
+    theseAccounts <- accounts[selected, , drop = FALSE]
     if (nrow(theseAccounts) == 0) {
       cli::cli_abort(
         c(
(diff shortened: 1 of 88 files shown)


R/account-find.R
```diff
@@ -17,7 +26,11 @@ findAccount <- function(accountName = NULL, server = NULL, error_call = caller_e
   }
 
   if (!is.null(accountName) && !is.null(server)) {
-    theseAccounts <- accounts[accounts$server == server & accounts$name == accountName, , drop = FALSE]
+    theseAccounts <- accounts[
+      accounts$server == server & accounts$name == accountName,
+      ,
+      drop = FALSE
+    ]
```
toph-allen, line 33: This one looks weird, even though it's technically correct.

R/deployments-find.R
```diff
@@ -37,9 +44,14 @@ disambiguateDeployments <- function(appDeployments, error_call = caller_env()) {
...
-    "(", accountLabel(appDeployments$account, appDeployments$server), "): ",
-    "{.url ", appDeployments$url, "}"
+    appDeployments$name,
+    " ",
+    "(",
+    accountLabel(appDeployments$account, appDeployments$server),
+    "): ",
+    "{.url ",
+    appDeployments$url,
+    "}"
```
toph-allen, line 54: This one seems worse. I'd put the `#fmt: skip` comment in myself, but it should probably go above line 46.

toph-allen (approved):
Looks good! I left a comment on two places where we might want to skip formatting, and there might have been a few I skipped, but — such an improvement!

R/account-find.R
aronatkins: I've rewritten to avoid the need to use a multiline expression.

R/deployments-find.R
aronatkins: This one, I've rewritten to use `sprintf`. It's more readable than the original or reformatted form.

aronatkins:
Note: Test failures occurred because some test data was reformatted, which meant that some line number expectations needed to be updated.

PR #1159, opened 2025-05-14
sanitize json created by system2

edavidaja:
`system2()` splits the json produced by `quarto inspect` if chunks are too long, producing errors on windows. Before parsing json we now trim whitespace and collapse the character vector produced by `system2()`.


aronatkins (commented):
Could you add a NEWS entry?

tdstein (approved):

PR #1162, opened 2025-05-15
head: declare utils import and use of utils::head

aronatkins:


toph-allen (approved):
Small change, no real comments, looks good!

PR #1163, opened 2025-05-15
tests: generate more test projects

aronatkins:
Generate test projects from code rather than persisting files on-disk. 

Use existing helpers to dynamically create projects. This avoids including "hidden" files (.python-version) in the repository and lets tests modify the target project directory without altering the repository contents (quarto!)


toph-allen (approved):
Looks good! Thanks for going through these.

tests/testthat/helper-content.R
```diff
@@ -0,0 +1,194 @@
+# A basic Quarto website that uses Python.
+quarto_website_py_files <- list(
```
marcosnav, line 2: These helpers are 💯 ! 

Thank you for improving these 🙇

PR #1168, opened 2025-05-20
set `rawResponse=TRUE` for snowflake connections only

### LINKED ISSUE #1155: Use v1 APIs in client-connect.R
We should upgrade https://github.com/rstudio/rsconnect/blob/main/R/client-connect.R to use public APIs.

edavidaja:
condition whether rawResponse is set to true on whether request is made to snowflake

closes #1155

### DIFF
--- a/R/servers.R
+++ b/R/servers.R
@@ -228,13 +228,21 @@ validateConnectUrl <- function(
         url,
         snowflakeConnectionName
       )
+      return(
+        GET(
+          parseHttpUrl(url),
+          auth_info,
+          "/server_settings",
+          timeout = timeout,
+          rawResponse = TRUE
+        )
+      )
     }
     GET(
       parseHttpUrl(url),
       auth_info,
       "/server_settings",
-      timeout = timeout,
-      rawResponse = TRUE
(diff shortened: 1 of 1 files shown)


aronatkins (approved):

ChaitaC:
Verified the fix. 
Verified the fix, but found an edge case:
Steps:
1. In the IDE, go to Publish → Manage Accounts...
2. Select an existing account and click Reconnect
3. When the authentication window appears, cancel out instead of completing the auth flow (screen recording attached below)
4. Try to go through the auth flow again via Publish → Manage Accounts...

Result: No pre-existing accounts are visible.

```
Error in rle(toplevel) : 'x' must be a vector of an atomic type
⠸ Waiting for authentication...
```

https://github.com/user-attachments/assets/9c13bba4-8d93-45de-9d34-349043ad0e64

aronatkins:
The observed error is likely an already-fixed RStudio issue (https://github.com/rstudio/rstudio/issues/10672).

ChaitaC:
https://github.com/rstudio/rstudio/issues/16054

PR #1172, opened 2025-06-03
Detect plumber vs. plumber2 in `appDependencies()`

### LINKED ISSUE #1151: Automatically detect `plumber2` APIs, infer dependencies
`rsconnect` infers R package dependencies [here](https://github.com/rstudio/rsconnect/blob/main/R/appDependencies.R#L150-L163) (called from [here](https://github.com/rstudio/rsconnect/blob/main/R/bundle.R#L136-L147)).

Forthcoming support for `plumber2` in Connect will mean that `rsconnect` needs to detect content that uses `plumber2`.

Right now, this code just keys off the app mode.

As far as I understand it, for `plumber2`, it will need to check for the presence of a `_server.yml` file and check that file's `engine` property.

### LINKED ISSUE #1170: Check if plumber2, not plumber, is installed when deploying plumber2 APIs
We currently require plumber to be installed when deploying APIs. Going forward we will want to require plumber for plumber APIs, but plumber2 for plumber2 APIs.

https://github.com/rstudio/rsconnect/blob/78700b8e97ad902b9a37797bb257a346cbe0ecc1/R/deployAPI.R#L18-L22

karawoo:
Updates `appDependencies()` to determine whether plumber or plumber2 (or another package provided as the engine in a plumber2 `_server.yml` file) is needed. 

Also removes `deployAPI()`'s check for plumber. I believe in our discussions we settled on feeling ok about removing this check altogether, however we could alternatively rewrite it to:

```
if (apiIsPlumber2(api)) { 
  # check for plumber 2
} else { 
  # check for plumber
}
```

Closes #1151, closes #1170.

### DIFF
--- a/R/appMetadata.R
+++ b/R/appMetadata.R
@@ -82,13 +82,24 @@ appMetadata <- function(
     quartoInfo <- NULL
   }
 
+  plumberInfo <- NULL
+  if (appMode == "api") {
+    if (apiIsPlumber2(appDir)) {
+      server_yaml <- yaml::read_yaml(file.path(appDir, "_server.yml"))
+      plumberInfo <- server_yaml$engine
+    } else {
+      plumberInfo <- "plumber"
+    }
+  }
+
   list(
     appMode = appMode,
     appPrimaryDoc = appPrimaryDoc,
     hasParameters = hasParameters,
     contentCategory = contentCategory,
     documentsHavePython = documentsHavePython,
-    quartoInfo = quartoInfo
+    quartoInfo = quartoInfo,
+    plumberInfo = plumberInfo
   )
 }
 
@@ -316,7 +327,6 @@ appIsQuartoDocument <- function(appMode) {
     )
 }
 
-
 appHasParameters <- function(
   appDir,
   appPrimaryDoc,
@@ -373,4 +383,9 @@ documentHasPythonChunk <- function(filename) {
   lines <- readLines(filename, warn = FALSE, encoding = "UTF-8")
   matches <- grep("`{python", lines, fixed = TRUE)
   return(length(matches) > 0)
+}
+
+apiIsPlumber2 <- function(appDir) {
+  files <- list.files(appDir)
+  any(grepl("_server\\.yml", files))
 }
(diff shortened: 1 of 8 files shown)


R/appMetadata.R
```diff
@@ -374,3 +384,8 @@ documentHasPythonChunk <- function(filename) {
   matches <- grep("`{python", lines, fixed = TRUE)
   return(length(matches) > 0)
 }
+
+apiIsPlumber2 <- function(appDir) {
+  files <- list.files(appDir)
+  any(grepl("^_server\\.yml$", files))
+}
```
jonkeane: FYI, looks like plumber2 did end up implementing support for yaml too.
```suggestion
apiIsPlumber2 <- function(appDir) {
  files <- list.files(appDir)
  any(grepl("^_server\\.ya?ml$", files))
}
```

https://github.com/posit-dev/plumber2/issues/37#issuecomment-2939319617

R/appMetadata.R
```diff
@@ -82,13 +82,24 @@ appMetadata <- function(
     quartoInfo <- NULL
   }
 
+  plumberInfo <- NULL
+  if (appMode == "api") {
+    if (apiIsPlumber2(appDir)) {
+      server_yaml <- yaml::read_yaml(file.path(appDir, "_server.yml"))
```
aronatkins: Similar to the suggestion by @jonkeane, we need to handle both `_server.yml` and `server.yaml`.

Perhaps `apiIsPlumber2()` could be `findServerYAML()` and return the appropriate filename or NA when it does not exist.

This can happen in a follow-up PR, if you'd prefer.

R/appMetadata.R
aronatkins: This modified pattern also needs to be changed in `inferAppMode()`. Feel free to land the `yml` vs `yaml` change separately, however.

R/appMetadata.R
aronatkins: If you'd like, you could follow the pattern of Quarto and have a `inferPlumberInfo()` function that contains all of this logic.

One difference between Quarto and Plumber: The Quarto info is a list of `version` and `engines` (itself a list). We do not need to make the `plumberInfo` a more complicated object today, but I like that `appMetadata()` is returning a `plumberInfo` field, letting us extract additional information from the YAML file  in the future, if needed.

PR #1173, opened 2025-06-04
depend on snowflakeauth for SPCS functions

edavidaja:
- removes vendored code from snowflakeauth package
- removes keypair auth dependencies


NAMESPACE
```diff
@@ -62,6 +60,8 @@ export(updateAccountEnvVars)
 export(writeManifest)
 import(rlang)
 importFrom(lifecycle,deprecated)
+importFrom(snowflakeauth,snowflake_connection)
```
aronatkins: Prefer `package::function()` rather than importing from that package into our namespace. 
https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-imports-r-code

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # rsconnect (development version)
 
+* Functions for interacting with Posit Connect deployments in
+  Snowpark Container Services are now provided by snowflakeauth.
```
aronatkins:
```suggestion
  Snowpark Container Services are now provided by the snowflakeauth package.
```

PR #1177, opened 2025-06-13
certs: use openssl to manipulate certificates, remove duplicates

### LINKED ISSUE #1175: curl_fetch_memory error on curl 6.3.0
When running `rsconnect::setAccountInfo`, I receive the following error:

```
> rsconnect::setAccountInfo(name='asdfasdf', token='blah', secret='blah')
Error in curl::curl_fetch_memory(url, handle = handle) : 
  SSL connect error [api.shinyapps.io]: TLS connect error: error:0BFFF065:x509 certificate routines:CRYPTO_internal:cert already in hash table
```

This looks to be an incompatibility with the latest version of the curl package, `6.3.0` - this works fine after installing `6.2.3`. Its [NEWS](https://github.com/jeroen/curl/blob/master/NEWS) makes reference to a libcurl upgrade to 8.14.1 and a change to mac-native LibreSSL:

```
 - MacOS/Windows: update to libcurl 8.14.1 
 - MacOS: our static build of libcurl now links to the MacOS native LibreSSL
   and nghttp in order to use the same keychain certificates as native curl.
```

I'm able to reproduce this on MacOS - not sure if this is also an issue on Windows.

aronatkins:
removed the existing (fake) localhost.crt and store.crt, replacing them with files created by mkcert.

mkcert example.com
mkcert localhost

fixes #1175


edavidaja (approved):
👍🏾

Issue #1118, opened 2024-11-27
.rscignore file doesn't ignore files in subfolders


tomcardoso:
Hello! I have a project with a folder, `data/`, in the root, and then two sets of data in subfolders, for a structure like this:
```
my-shiny-app/
├─ data/
│  ├─ filtered/
│  │  ├─ dataset_1.rds
│  │  ├─ dataset_2.rds
│  ├─ unfiltered/
│  │  ├─ dataset_1.rds
│  │  ├─ dataset_2.rds
```

If I try to exclude either the `unfiltered` folder or its specific files, neither works when running `rsconnect::listDeploymentFiles()`.

Here's the `.rscignore` I'm using:
```
data/unfiltered/dataset_1.rds
data/unfiltered/dataset_2.rds
```

I did some testing and it looks like anything with a subfolder path doesn't work. I could remove a level (making the `filtered` and `unfiltered` folders both root-level), but would be nice for organization's sake to have the `.rscignore` recognize paths like this.

tomcardoso:
Ah, interesting. Thanks for this tip!

Issue #774, opened 2023-03-16
bug: `. rcsignore` doesn't work with folder or files in subfolders


ColinFay:
## Issue

The . rcsignore file doesn't work with : 
  1. folders that are listed as `folder/` instead of `folder`
  2. files that are in subdirectories

## Diagnosis

1. For `dev/` : the `setdiff` here is not smart enough : https://github.com/rstudio/rsconnect/blob/cf0972d0d69357cf2bc81ab140f940085aac92cb/R/bundleFiles.R#L200

`list.files()` (that created the `contents` object) will do 

```
[1] "DESCRIPTION" "dev"         "inst"        "man"         "NAMESPACE"  
[6] "R" 
```

When a directory can usually be written `folder/` in an ignore file. 

2. The list file that creates the contents object doesn't list subfolders: recursiveBundleFiles doesn't do a recursive in the list file https://github.com/rstudio/rsconnect/blob/cf0972d0d69357cf2bc81ab140f940085aac92cb/R/bundleFiles.R#L137

## How to reproduce 

1. Go to /tmp & create a golem app:

```
; cd /tmp && Rscript -e 'golem::create_golem("reprexrsconnect")' && cd reprexrsconnect
```

4. Add 'DESCRIPTION' to `.rscignore` => works as expected 

```
; echo 'DESCRIPTION' >> .rscignore  
; ls
DESCRIPTION     NAMESPACE       R               dev             inst            man
; Rscript -e 'rsconnect::listDeploymentFiles(".")'
 [1] ".here"                    ".Rbuildignore"           
 [3] "dev/01_start.R"           "dev/02_dev.R"            
 [5] "dev/03_deploy.R"          "dev/run_dev.R"           
 [7] "inst/app/www/favicon.ico" "inst/golem-config.yml"   
 [9] "man/run_app.Rd"           "NAMESPACE"               
[11] "R/app_config.R"           "R/app_server.R"          
[13] "R/app_ui.R"               "R/run_app.R"
```

5. Trying to ignore `dev/` => doesn't work

```
; echo 'dev/' >> .rscignore
; Rscript -e 'rsconnect::listDeploymentFiles(".")'
 [1] ".here"                    ".Rbuildignore"           
 [3] "dev/01_start.R"           "dev/02_dev.R"            
 [5] "dev/03_deploy.R"          "dev/run_dev.R"           
 [7] "inst/app/www/favicon.ico" "inst/golem-config.yml"   
[...]

PR #1178, opened 2025-06-15
Add generateFileManifest function for enhanced pattern support

BrianLang:
Rather than directly suggesting an update to current implementation of .rscignore files, I'd like to propose the addition of a new exported function that does have more flexible matching than the current .rscignore file approach. 

Happy for feedback and a discussion on how we can facilitate this. 

Without an appetite for inclusion I'd be willing to just make a small package instead. 

- Introduced `generateFileManifest()` to process enhanced ignore patterns from `.rscignore-patterns` files, allowing for advanced file filtering during deployment.
- Added support for exact filenames, wildcard patterns, and directory exclusions.
- Implemented a helper function `parseRscignoreFile()` to handle pattern file parsing.
- Created comprehensive tests for various pattern matching scenarios.
- Updated documentation and added a vignette for usage examples and integration with `deployApp()`.
- Exported the new function in NAMESPACE.

#1118 
#774 @colinfay could this be a path forward? (yes, a very old issue of yours)


BrianLang:
Hi @aronatkins, thanks for the response. I am happy to see this move forward in either case, whether it is an enhancement of the .rscignore processing or if someone makes available a nice helper for creating a file manifest with a similar level of flexibility.

My main consideration was mostly avoiding digging too deep into your codebase. I see that there is a recursive calculation of bundle size, which is facilitated by the fact that you are only ever considering the root-level files and directories.

karawoo:
I think we'd like to bring `.rscignore` in line with `.renvignore`/`.gitignore` behavior without adding a separate file. This would also involve continuing to support `.rscignore` files placed in subdirectories within the file hierarchy. Additionally, we'd have to ensure that the correct precedence of `.rscignore` files is respected. Right now this is not an issue because `.rscignore` files can only reference files/folders within their immediate directory. If we expand the rules, then an `.rscignore` in the root directory could contradict entries in an `.rscignore` in a child directory; in these cases the child directory's `.rscignore` needs to take precedence.

It is a fairly big effort so I definitely see why you opted for a separate file first. However I think the best user experience will be for `.rscignore` to behave just like `.renvignore`/`.gitignore`.

Would you be interested in taking a pass at updating this PR? Totally understand if it's more than you want to take on. If that's the case then we'll start work on an implementation when we're able (but no guarantees as to when that will be).

karawoo:
> Is this why the current implementation depends on recursion?

Yes, currently the recursive approach allows us to ignore files in subdirectories if there's a .rscignore file present in the subdirectory.

I'd be ok with broadening our rscignore syntax incrementally and not supporting negation yet, but I do think we'd want to implement it eventually, as it's a useful feature that users may expect if they are already familiar with .gitignore. 

Another option is to [vendor renv](https://rstudio.github.io/renv/reference/vendor.html?q=vendor) and reuse its (unexported) parsing of .renvignore files. That is a little confusing since rsconnect already depends on renv too. But it might be the quickest way to get support for the full syntax. 

If we're not implementing the full .renvignore syntax from scratch right at this moment then I think I lean toward the latter option (vendoring renv). But I'd be willing to review a PR that takes either approach as long as it moves us in the direction we want to go with these files, without removing functionality that currently exists.

BrianLang:
closing as this will be addressed on a different branch and PR #1185

PR #1183, opened 2025-07-25
chore: remove unused functions from Connect client

toph-allen:
Remove three functions from the Connect client which are neither used nor supported. As far as I can tell, the other functions should remain.

There is a test in `test-connect-client.R` which references the removed functions. It's described as "Users API" but it tests all the functions in `connectClient` in such a way that later parts of it reference the early parts. It's skipped everywhere. I don't know what the best thing to do with it is, and would welcome opinions. (Going with the first option listed below unless there are differences of opinion!)

- Delete it? It serves no purpose now. Create an issue to add tests for `connectClient`. Maybe paste the old function body in there. Leaning towards this, so I'm going to do this now.
- Keep it? It at least serves to document things that _need_ to be tested, and maybe it could be made to work? (I don't think it'd be trivial to get it to work.)
- Keep it, but delete the parts that reference the user API? I don't think so, that'd leave something that's vestigial doesn't even make internal sense.


tests/testthat/test-client-connect.R
```diff
@@ -91,123 +91,3 @@ findConnect <- function() {
     stop("Couldn't find an appropriate 'connect' binary")
   }
 }
-
-test_that("Users API", {
```
aronatkins, line 95: These were the original tests written against Connect. I have never run them myself, which tells you something about their age. They made very specific assumptions about a local server configuration.

They should have been removed years ago. Thanks for removing.

PR #1198, opened 2025-08-29
fix: emit the underlying server validation error

### LINKED ISSUE #1197: provide underlying error when addServer() fails
```
> rsconnect::addServer("https://this-is-not-a-valid-server")
Error in `rsconnect::addServer()`:
! `url` does not appear to be a Posit Connect server.
```

We should provide the underlying error; it is available, but dropped.

aronatkins:
fixes #1197


aronatkins:
Closing; this is useful for some types of errors, but not for many others.

PR #1209, opened 2025-09-22
feat: getLogs for shinyapps.io applications

### LINKED ISSUE #1086: Feature request: showLogs() output to table
It would be nice if there was a way to convert the output of `showLogs()` to a table, or an argument to define the format of what is returned.  Columns could perhaps be `date_time`, `app_id`, `message`.

(Currently it isn't even possible to capture the output in order to manually parse it into a table):

```r
x <- showLogs()
x
#> NULL
```

aronatkins:
fixes #1086

```
> logs <- rsconnect::getLogs(appName="shiny", server="shinyapps.io")
> head(logs)
            timestamp account_id application_id                                                                message
1 2025-09-19 11:24:22      XXXXX        YYYYYYY                         Container event from container-12416658: start
2 2025-09-19 11:24:24      XXXXX        YYYYYYY                                          Running on host: 53eeecb13a99
3 2025-09-19 11:24:24      XXXXX        YYYYYYY                                                         Process ID: 31
4 2025-09-19 11:24:24      XXXXX        YYYYYYY Running as user: uid=10001(shiny) gid=10001(shiny) groups=10001(shiny)
5 2025-09-19 11:24:24      XXXXX        YYYYYYY                                             Connect version: 2025.02.0
6 2025-09-19 11:24:24      XXXXX        YYYYYYY                                                          LANG: C.UTF-8

```

### DIFF
--- a/R/applications.R
+++ b/R/applications.R
@@ -344,16 +344,11 @@ getLogs <- function(
   application <- getAppByName(client, accountDetails, deployment$name)
 
   payload <- client$getLogs(application$id, entries, format = "json")
-  payload
 
-  # entries in logs$results have unstable field order.
-  ordered <- lapply(payload$results, function(row) row[order(names(row))])
-  combined <- do.call(rbind, ordered)
-
-  # JSON loves wrapped values.
-  unwrapped <- lapply(colnames(combined), function(nm) unlist(combined[, nm]))
-  df <- as.data.frame(unwrapped)
-  names(df) <- colnames(combined)
+  # Convert to a dataframe before combining because the JSON payload has inconsistent field order
+  # containing nested single-element lists.
+  converted <- lapply(payload$results, as.data.frame)
+  df <- do.call(rbind, converted)
 
   # shinyapps.io returns ns timestamps.
   df$timestamp <- as.POSIXct(df$timestamp / (1000 * 1000))


R/applications.R
```diff
@@ -316,6 +322,52 @@ showLogs <- function(
...
+    server = server,
+    account = account
+  )
+  accountDetails <- accountInfo(deployment$account, deployment$server)
+  client <- clientForAccount(accountDetails)
+  application <- getAppByName(client, accountDetails, deployment$name)
+
+  payload <- client$getLogs(application$id, entries, format = "json")
+  payload
+
```
aronatkins, line 347: For reviewers: The following code transforms the (already parsed) JSON payload into something that we can use. Suggestions for improvement?

R/applications.R
```diff
@@ -316,6 +322,52 @@ showLogs <- function(
...
+    appName = appName,
+    server = server,
+    account = account
+  )
+  accountDetails <- accountInfo(deployment$account, deployment$server)
+  client <- clientForAccount(accountDetails)
+  application <- getAppByName(client, accountDetails, deployment$name)
+
+  payload <- client$getLogs(application$id, entries, format = "json")
+  payload
```
karawoo:
```suggestion
```

R/applications.R
```diff
@@ -316,6 +322,52 @@ showLogs <- function(
...
+
+  # entries in logs$results have unstable field order.
+  ordered <- lapply(payload$results, function(row) row[order(names(row))])
+  combined <- do.call(rbind, ordered)
+
+  # JSON loves wrapped values.
+  unwrapped <- lapply(colnames(combined), function(nm) unlist(combined[, nm]))
+  df <- as.data.frame(unwrapped)
+  names(df) <- colnames(combined)
```
karawoo:
```suggestion
  ordered <- lapply(payload$results, function(row) as.data.frame(row[order(names(row))]))
  df <- do.call(rbind, ordered)
```

I think if you convert to data frame earlier then some of the subsequent parsing becomes unnecessary

R/applications.R
aronatkins: Done. Thanks. Remnant from earlier version where I was returning `payload` and playing with the object.

R/applications.R
aronatkins: Thanks for the nudge. I've done something similar. It looks like we don't need to reorder the columns if we convert to a data frame ahead of `rbind`.

karawoo (approved):

R/applications.R
```diff
@@ -258,8 +262,10 @@ streamApplicationLogs <- function(authInfo, applicationId, entries, skip) {
 #'   function does not return; instead, log entries are written to the console
 #'   as they are made, until R is interrupted. Defaults to `FALSE`.
 #'
-#' @note This function only uses the \code{libcurl} transport, and works only for
-#'   ShinyApps servers.
+#' @note These functions only use the \code{libcurl} transport, and only work
+#'   for applications deployed to ShinyApps.io.
+#'
+#' @return `getLogs()` returns a data frame containing the logged lines.
```
toph-allen, line 268: Should this also document what `getLogs()` returns? It says "displays the logs" earlier in the docs, but does this just print them and return NULL or does it return them as text?

toph-allen (approved):
This looks good to me! Took me a second to find where `jsonlite` actually gets called. :)

R/applications.R
aronatkins: Did you mean `showLogs()`? I didn't add documentation about its return value because it either never returns (when it streams logs) or returns the result from `cat()`.

What's your preference here?

toph-allen: I'm fine leaving it as-is, or changing it — this is a super soft recommendation.

If I was gonna write about `showLogs()` return value here:

```suggestion
#' @return `getLogs()` returns a data frame containing the logged lines. `showLogs()` returns a character vector or streams logs without returning, depending on the value of the `streaming` argument.
```

It's definitely duplicative of the arg docs, but it might be helpful. I lean more towards more text usually, but that's not always good, so I'll leave the final decision up to you.

karawoo: I don't believe this is correct, `cat()` returns `NULL`.

toph-allen: Oh yeah you're right

toph-allen: Thanks — let's just ignore this then. :)

karawoo: Though I do think it's sort of unintuitive behavior, so we could consider invisibly returning the log lines in addition to `cat`-ing them.

aronatkins: Given `getLogs()`, it feels like `showLogs()` doesn't need to return a value.

aronatkins: I'm leaving `showLogs()` as-is for now.

PR #1213, opened 2025-09-23
fix: ignore expires= values which cannot be parsed

### LINKED ISSUE #1212: Cookie expiration check failing
Reported by a customer.

With `rsconnect 1.5.1`, they are getting the following error when deploying content:
```
Error in if (any(cookies$expires < as.integer(Sys.time()))) { : 
  missing value where TRUE/FALSE needed
Calls: <Anonymous> ... POST -> httpRequestWithBody -> http -> appendCookieHeaders
Execution halted
```

Downgrading to `1.5.0` works for this customer.

This customer is deploying from Workbench.

aronatkins:
Both strptime and curl::parse_date return NA when the incoming date cannot be parsed. Ignore NA results and treat those cookies as session cookies.

Shift to curl::parse_date because it recognizes more date formats.

Add tests showing treatment of expires= values which are not valid dates.

Fixes #1212

### DIFF
--- a/R/cookies.R
+++ b/R/cookies.R
@@ -138,8 +138,8 @@ parseCookie <- function(cookieHeader, requestPath = NULL) {
 
   # Cookies without max-age= or expires= are session cookies.
   #
-  # Cookies never survive beyond the current R session, so mark session
-  # cookies as never expiring.
+  # Mark session cookies as never expiring (far in the future). Cookies
+  # are managed in-memory and never survive beyond the current R session.
   expires <- Sys.time() + 10^10
 
   # Max-Age takes precedence over Expires.
@@ -166,12 +166,10 @@ parseCookie <- function(cookieHeader, requestPath = NULL) {
     )[[1]]
     if (length(expiration_value) > 0) {
       # Time format is Wdy, DD-Mon-YYYY HH:MM:SS GMT
-      # could also use curl::parse_date()
-      expires <- strptime(
-        expiration_value[2],
-        "%a, %d %b %Y %H:%M:%S GMT",
-        "GMT"
-      )
+      parsed <- curl::parse_date(expiration_value[2])
+      if (!is.na(parsed)) {
+        expires <- parsed
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


R/cookies.R
```diff
@@ -194,6 +194,11 @@ appendCookieHeaders <- function(requestURL, headers) {
 
   cookies <- get(host, envir = .cookieStore)
 
+  if (nrow(cookies) == 0) {
+    # no initial cookies.
+    return(headers)
+  }
+
   # If any cookies are expired, remove them from the cookie store
   if (any(cookies$expires < as.integer(Sys.time()))) {
```
karawoo: if `cookies$expires` is `NA` then I'm not sure this will work, because it's not a zero-row df so won't return before it reaches the `if`. maybe `isTRUE` instead?

```suggestion
  # If any cookies are expired, remove them from the cookie store
  if (isTRUE(any(cookies$expires < as.integer(Sys.time())))) {
```

aronatkins: I've adopted this suggestion and also found the source of the NA.

R/cookies.R
```diff
@@ -195,14 +193,14 @@ appendCookieHeaders <- function(requestURL, headers) {
   cookies <- get(host, envir = .cookieStore)
 
   # If any cookies are expired, remove them from the cookie store
-  if (any(cookies$expires < as.integer(Sys.time()))) {
+  if (isTRUE(any(cookies$expires < as.integer(Sys.time())))) {
```
aronatkins, line 196: @karawoo - Took this suggestion from you even though we are now avoiding the NA values.

R/cookies.R
```diff
@@ -166,12 +166,10 @@ parseCookie <- function(cookieHeader, requestPath = NULL) {
     )[[1]]
     if (length(expiration_value) > 0) {
       # Time format is Wdy, DD-Mon-YYYY HH:MM:SS GMT
-      # could also use curl::parse_date()
-      expires <- strptime(
-        expiration_value[2],
-        "%a, %d %b %Y %H:%M:%S GMT",
-        "GMT"
-      )
+      parsed <- curl::parse_date(expiration_value[2])
```
aronatkins, line 169: `curl::parse_date` accepts more date formats.

PR #1215, opened 2025-09-24
remove posit cloud functionality

mslynch:


R/deployApp.R
```diff
@@ -241,8 +237,7 @@ deployApp <- function(
   image = NULL,
   envManagement = NULL,
   envManagementR = NULL,
-  envManagementPy = NULL,
-  space = NULL
```
aronatkins, line 245: It is a breaking change to remove an argument. Given that it is supremely unlikely that folks are relying on positional `deployApp()` arguments, we can probably get away with it, but we also could leave it and document it as an obsolete, ignored argument.

CC @karawoo - what are your thoughts about this?

R/servers.R
```diff
@@ -94,7 +74,7 @@ isRPubs <- function(server) {
 }
 
 isConnectServer <- function(server) {
-  !isCloudServer(server) && !isRPubs(server)
+  !isShinyappsServer(server) && !isRPubs(server)
 }
 
 cloudServerInfo <- function(name, url) {
```
aronatkins: nit: `cloudServerInfo()` continues to use the "cloud" name. Probably not enough to rewrite, but it only applies to shinyapps.io, now.

tests/testthat/test-servers.R
```diff
@@ -142,55 +141,23 @@ test_that("certificates can't be attached to plain http servers", {
...
 })
 
 test_that("only shinyapps.io is identified as shinyapps.io", {
   checkShinyappsServer("shinyapps.io")
-  expect_error(checkShinyappsServer("rstudio.cloud"))
-  expect_error(checkShinyappsServer("posit.cloud"))
   expect_error(checkShinyappsServer("connect.internal"))
 })
 
 test_that("predefined servers includes cloud and shinyapps", {
```
aronatkins: nit: adjust test name now that posit.cloud is absent.

R/deployApp.R
karawoo: If `space` were earlier in the argument list I'd say maybe let's deprecate with a warning and remove in a later release, but since it's the last argument I think it's ok to just remove it. The only case where removing it should be an issue is if they've provided ALL arguments positionally, _including_ `space`, which means that they are probably trying to deploy to Posit Cloud, which is going to break anyway.

aronatkins: That's how I was feeling, too, @karawoo. Thanks.

@mslynch - Let's make sure that this is included in NEWS.

NEWS.md
```diff
@@ -11,6 +11,10 @@
 
 * Improved documentation and advice for `deployApp(envVars...)`.
 
+* Removed support for publishing to Posit Cloud. (#1215)
+
+* Removed the Cloud-exclusive `space` argument from `deployApp()`. (#1215)
```
aronatkins:
```suggestion
* Removed the Posit Cloud-exclusive `space` argument from `deployApp()`. (#1215)
```

R/deployApp.R
```diff
@@ -377,16 +372,6 @@ deployApp <- function(
     )
   }
 
-  # Run checks prior to first saveDeployment() to avoid errors that will always
-  # prevent a successful upload from generating a partial deployment
-  if (!isCloudServer(accountDetails$server) && identical(upload, FALSE)) {
```
aronatkins, line 382: Why did this block disappear? It ran against non-cloud servers.

mslynch: Oops, I think I missed the `!` here. I'll put this back.

NEWS.md
```diff
@@ -15,6 +15,11 @@
 
 * Improve documentation and advice for `deployApp(envVars...)`.
 
+* Removed support for publishing to Posit Cloud. (#1215)
+  * Existing Posit Cloud account records may be removed by using `rsconnect::removeAccount("yourname", "posit.cloud")`.
```
aronatkins:
```suggestion

  Existing Posit Cloud account records may be removed by using `rsconnect::removeAccount("yourname", "posit.cloud")`.
```

Let's use a second paragraph for this item rather than a sublist. We have previous examples of this in the release notes for 0.8.28.

aronatkins (approved):

PR #1217, opened 2025-09-25
add support for deploying to Posit Connect Cloud

mslynch:
- **start adding support for connect cloud**
- **remove posit cloud functionality**
- **remove space variable reference**
- **remove space from deployApp.Rd**
- **remove more posit cloud references**
- **add back old bundle upload for shinyapps.io**


aronatkins:
soft suggestion: It might be simpler for reviewers to see the addition of connect cloud support before posit cloud is removed.

R/deployApp.R
```diff
@@ -642,13 +729,26 @@ applicationDeleted <- function(client, deployment, recordPath, appMetadata) {
   unlink(path)
 
   accountDetails <- accountInfo(deployment$account, deployment$server)
-  client$createApplication(
-    deployment$name,
-    deployment$title,
-    "shiny",
```
mslynch, line 648: This seems strange... should it be appMetadata$appMode?

aronatkins: The "shiny" value is used as the application `template` for shinyapps.io; unused by Connect and Posit Cloud (on main).

What is the purpose of the `template`?

mslynch: Ah, got it. We use that in Cloud to differentiate between shinyapps.io apps, Posit Cloud IDE projects, and Jupyter projects.

R/accounts.R
```diff
@@ -153,6 +153,126 @@ connectUser <- function(
...
+
+connectCloudUser <- function() {
+  authClient <- cloudAuthClient()
+  deviceAuth <- authClient$createDeviceAuth()
+
+  # Alert user and open browser for verification
+  cli::cli_alert_info(
+    "Opening login page. When requested, please enter the code {deviceAuth$user_code}."
+  )
+  utils::browseURL(deviceAuth$verification_uri_complete)
```
aronatkins: Look at how `connectUser` and `deployApp` implement their `launch.browser` option. Consider mirroring that here and in each place where `utils::browseURL` is called.

R/accounts.R
```diff
@@ -153,6 +153,126 @@ connectUser <- function(
...
+          identical(permission, "content:create")
+        },
+        logical(1)
+      ))
+    },
+    accounts
+  )
+}
+
+connectCloudUser <- function() {
```
aronatkins: Documentation, please.

R/accounts.R
```diff
@@ -153,6 +153,126 @@ connectUser <- function(
...
+  deviceAuth <- authClient$createDeviceAuth()
+
+  # Alert user and open browser for verification
+  cli::cli_alert_info(
+    "Opening login page. When requested, please enter the code {deviceAuth$user_code}."
+  )
+  utils::browseURL(deviceAuth$verification_uri_complete)
+
+  pollingInterval <- deviceAuth$interval
+  while (TRUE) {
```
aronatkins: Consider using a helper function to separate some of this looping logic from the provisioning workflow. See `waitForAuthedUser` for one example (along with its progress updates).

R/accounts.R
```diff
@@ -153,6 +153,126 @@ connectUser <- function(
...
+
+  cloudUiUrl <- connectCloudUrls()$ui
+  if (length(accountsWhereUserCanPublish) == 0) {
+    if (length(accounts) == 0) {
+      cli::cli_alert_info(
+        "To deploy, you must finish creating an account. Opening account creation page..."
+      )
+      utils::browseURL(cloudUiUrl)
+      while (TRUE) {
+        Sys.sleep(2)
```
aronatkins, line 277: How long do folks normally take to create an account? Is it better to exit after opening the error or loop until success?

R/accounts.R
```diff
@@ -153,6 +153,126 @@ connectUser <- function(
...
+
+  # prompt the user to select an account if there's more than one they can publish to
+  if (length(accountsWhereUserCanPublish) > 1) {
+    cli::cli_alert_info("You have permission to publish to multiple accounts.")
+    accountNames <- vapply(
+      accountsWhereUserCanPublish,
+      function(account) account$name,
+      character(1)
+    )
+    selected <- utils::menu(accountNames, title = "Select an account to use:")
```
aronatkins: Other "choices" workflows use `cli_menu()`, which uses a number of `cli` functions under the hood. Related to https://github.com/r-lib/cli/issues/228, as a utils::menu replacement isn't part of cli, yet.

R/appMetadata.R
```diff
@@ -161,12 +165,19 @@ inferAppMode <- function(
   hasShinyQmd <- any(sapply(qmdFiles, isShinyRmd))
 
   if (hasShinyQmd) {
-    return("quarto-shiny")
+    return(list(
+      appMode = "quarto-shiny",
+      primaryFile = basename(qmdFiles[sapply(qmdFiles, isShinyRmd)][1])
+    ))
   } else if (hasShinyRmd) {
+    shinyRmdFile <- rmdFiles[sapply(rmdFiles, isShinyRmd)][1]
```
aronatkins: Is there a way to avoid repeatedly running `isShinyRmd` across the same set of files?

Something like (untested):

```r
shinyRmdFiles <- rmdFiles[sapply(rmdFiles, isShinyRmd)]

if (length(shinyRmdFiles) > 0) { ... }
```

R/appMetadata.R
```diff
@@ -208,11 +230,14 @@ inferAppMode <- function(
     "^(saved_model.pb|saved_model.pbtxt)$"
   )
   if (length(modelFiles) > 0) {
-    return("tensorflow-saved-model")
+    return(list(
+      appMode = "tensorflow-saved-model",
+      primaryFile = basename(modelFiles[1])
```
aronatkins: Using `basename` is appropriate for every type of content other than TensorFlow because every file is a direct descendant of `appDir`. TensorFlow model files are not at the top level.

R/client-cloudAuth.R
```diff
@@ -0,0 +1,67 @@
+# Map rsconnect appMode to Connect Cloud contentType
+
+getClientId <- function() {
```
aronatkins, line 2: Could you add comments describing the purpose for the various helper functions?

R/client-cloudAuth.R
```diff
@@ -0,0 +1,67 @@
+# Map rsconnect appMode to Connect Cloud contentType
```
aronatkins: Is this comment misplaced?

R/deployApp.R
```diff
@@ -374,7 +374,11 @@ deployApp <- function(
 
   # Run checks prior to first saveDeployment() to avoid errors that will always
   # prevent a successful upload from generating a partial deployment
-  if (!isShinyappsServer(accountDetails$server) && identical(upload, FALSE)) {
+  if (
+    !isShinyappsServer(accountDetails$server) &&
```
aronatkins: Could we use `isConnectServer` here?

R/deployApp.R
```diff
@@ -496,7 +554,14 @@ deployApp <- function(
 
     # create, and upload the bundle
     taskStart(quiet, "Uploading bundle...")
-    if (isShinyappsServer(accountDetails$server)) {
+    if (isPositConnectCloudServer(accountDetails$server)) {
+      uploadUrl <- updateResponse$next_revision$source_bundle_upload_url
+      success <- client$uploadBundle(bundlePath, uploadUrl)
+      if (!success) {
+        stop("Could not upload bundle.")
```
aronatkins: Consider `cli::cli_abort` instead of `stop()`.

R/accounts.R
mslynch: I think you might have been looking at an old version? This function has documentation now.

aronatkins: Thanks. They must have arrived while I was reviewing.

R/client-cloudAuth.R
mslynch: Yeah, that was a copy-paste error.

R/accounts.R
karawoo: could you please run `devtools::document()` and commit the resulting .Rd file?

R/appMetadata.R
mslynch: What would be an appropriate primaryFile in this case? Since only Connect supports this right now, maybe we could just use NULL.

aronatkins: TF requires a very particular directory structure. https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/saved_model/README.md#the-savedmodel-format

`NULL` is fine. When Connect runs a TF model, it uses the unpacked bundle directory as the argument to `--model_base_path`.

R/accounts.R
mslynch: I would expect this to usually take a minute or two at most. Are you asking if this should loop with the only exit condition being account creation?

R/appMetadata.R
```diff
@@ -157,16 +161,34 @@ inferAppMode <- function(
...
+  shinyRmdFiles <- if (length(rmdFiles) > 0) {
+    rmdFiles[sapply(rmdFiles, isShinyRmd)]
+  } else {
+    NULL
+  }
+  shinyQmdFiles <- if (length(qmdFiles) > 0) {
+    qmdFiles[sapply(qmdFiles, isShinyRmd)]
+  } else {
+    NULL
+  }
```
karawoo:
```suggestion
  shinyRmdFiles <- NULL
  if (length(rmdFiles) > 0) {
    shinyRmdFiles <- rmdFiles[sapply(rmdFiles, isShinyRmd)]
  } 
  shinyQmdFiles <- NULL
  if (length(qmdFiles) > 0) {
    shinyQmdFiles <- qmdFiles[sapply(qmdFiles, isShinyRmd)]
  } 
```

this is a nit you can take or leave, but I rarely see assignments from a whole `if` block in R; I think this way is a little more readable.

R/client-connectCloud.R
```diff
@@ -0,0 +1,206 @@
...
+    "python-shiny" = "shiny",
+    "shiny" = "shiny",
+    "python-streamlit" = "streamlit",
+    "quarto" = "quarto",
+    "quarto-static" = "quarto",
+    "quarto-shiny" = "quarto",
+    "rmd-static" = "rmarkdown",
+    "rmd-shiny" = "rmarkdown",
+    "static" = "static",
+    stop("appMode '", appMode, "' is not supported by Connect Cloud")
```
karawoo:
```suggestion
    stop("appMode '", appMode, "' is not supported by Connect Cloud", call. = FALSE)
```

this `stop()` is nested down in enough layers of unexported functions that I think it makes sense to omit the call from the error message.

R/deployApp.R
```diff
@@ -515,17 +584,27 @@ deployApp <- function(
...
-  task <- client$deployApplication(application, bundle$id)
-  taskId <- if (is.null(task$task_id)) task$id else task$task_id
-  # wait for the deployment to complete (will raise an error if it can't)
-  response <- client$waitForTask(taskId, quiet)
+  if (isPositConnectCloudServer(accountDetails$server)) {
+    client$publish(application$id)
+    revisionId <- updateResponse$next_revision$id
+    cloudDeployResult <- client$awaitCompletion(revisionId)
+    deploymentSucceeded <- cloudDeployResult$success
+    application$url <- cloudDeployResult$url
```
karawoo:
```suggestion
    response <- client$awaitCompletion(revisionId)
    deploymentSucceeded <- response$success
    application$url <- response$url
```

I might be missing something but I think if we make this `response` then we can simplify the branching at 625-631

R/deployApp.R
```diff
@@ -534,14 +613,22 @@ deployApp <- function(
...
       )
     } else {
-      cli::cli_alert_danger("Deployment failed with error: {response$error}")
+      if (isPositConnectCloudServer(accountDetails$server)) {
+        cli::cli_alert_danger(
+          "Deployment failed with error: {cloudDeployResult$error}"
+        )
+      } else {
+        cli::cli_alert_danger("Deployment failed with error: {response$error}")
+      }
```
karawoo:
```suggestion
      cli::cli_alert_danger("Deployment failed with error: {response$error}")
```

see above comment

R/http.R
```diff
@@ -172,7 +173,11 @@ handleResponse <- function(
     if (response$status %in% 200:399) {
       out <- json
     } else if (!is.null(json$error)) {
-      reportError(unlist(json$error))
+      if (!is.null(json$error_type)) {
+        reportError(unlist(json$error), errorType = unlist(json$error_type))
+      } else {
+        reportError(unlist(json$error))
+      }
```
karawoo:
```suggestion
      reportError(unlist(json$error), errorType = unlist(json$error_type))
```

I think you can simplify to this because if `json$error_type` is NULL then `unlist(json$error_type)` will also be NULL.

tests/testthat/test-client-connectCloud.R
```diff
@@ -0,0 +1,168 @@
...
+      )
+      stop(err)
+    } else {
+      # Second call succeeds
+      list(success = TRUE, data = "success after refresh")
+    }
+  }
+
+  # Mock cloudAuthClient
+  mockery::stub(connectCloudClient, "cloudAuthClient", function() {
```
karawoo: Is it possible to rewrite these mocks to use `testthat::local_mocked_bindings()`? The mockery package has been superseded so I'd avoid it if possible.

R/accounts.R
aronatkins: I was considering which option would be a better fit: "try forever" or "try again after creating your account". Fine trying what you've got.

mslynch: Thanks, I updated to make the messaging clearer with "Try again after creating a new account."

DESCRIPTION
```diff
@@ -41,6 +41,7 @@ Suggests:
     foreign,
     knitr,
     MASS,
+    mockery,
```
aronatkins: Did you remove use of mockery?

PR #1219, opened 2025-10-09
get logs from posit connect cloud

mslynch:
For Connect Cloud,
- Print messages about publishing status changes (building, rendering, deploying, etc)
- Print logs logs on failure, like we do for shinyapps.io

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,6 @@ Suggests:
     foreign,
     knitr,
     MASS,
-    mockery,
     plumber (>= 0.3.2),
     quarto,
     RCurl,


R/client-connectCloud.R
```diff
@@ -176,16 +176,114 @@ connectCloudClient <- function(service, authInfo) {
...
         path <- paste0("/revisions/", revisionId)
-        response <- withTokenRefreshRetry(GET, path)
+        revision <- withTokenRefreshRetry(GET, path)
+
+        newStatus <- revision$status
+        lastStatusIndex <- if (is.null(lastStatus)) {
+          -1
+        } else {
+          match(lastStatus, possibleStates)
+        }
```
toph-allen: Soft suggestion to move assignment inside the `if` statement.

```suggestion
        if (is.null(lastStatus)) {
          lastStatusIndex <- -1
        } else {
          lastStatusIndex <- match(lastStatus, possibleStates)
        }
```

R/client-connectCloud.R
```diff
@@ -176,16 +176,114 @@ connectCloudClient <- function(service, authInfo) {
...
+        newStatusIndex <- match(newStatus, possibleStates)
+
+        if (newStatusIndex > lastStatusIndex) {
+          # Note: since we poll every second, it's possible to skip states in
+          # the output here
+          cli::cli_alert_info(stateMessages[[newStatus]])
+          lastStatus <- newStatus
+        }
```
toph-allen: For my own comprehension: Is there a reason we message based on comparing the indices rather than comparing the statuses themselves, i.e. `if (newStatus != lastStatus)`?

R/client-connectCloudLogs.R
```diff
@@ -0,0 +1,26 @@
+# Creates a client for interacting with the Connect Cloud Logs API.
+connectCloudLogsClient <- function() {
+  list(
+    getLogs = function(logChannel, authToken) {
+      # Parse the logs URL to get service components
+      logsUrl <- connectCloudUrls()$logs
```
toph-allen, line 6: My on-prem-Connect-conditioned impulse was definitely "Wait, how can we know the server URL if we don't pass it in??" 😂

R/client-connectCloud.R
```diff
@@ -176,16 +176,114 @@ connectCloudClient <- function(service, authInfo) {
...
+
+        newStatus <- revision$status
+        lastStatusIndex <- if (is.null(lastStatus)) {
+          -1
+        } else {
+          match(lastStatus, possibleStates)
+        }
+        newStatusIndex <- match(newStatus, possibleStates)
+
+        if (newStatusIndex > lastStatusIndex) {
```
karawoo: Agree with @toph-allen, I think this could perhaps be simplified to something like


```suggestion
      stateMessages <- list(
        publish_deferred = "Content is currently publishing; your request will start soon.",
        publish_requested = "Publish requested; waiting to start...",
        publish_started = "Publish started.",
        fetching = "Retrieving code...",
        building = "Installing dependencies...",
        rendering = "Rendering...",
        publishing = "Publishing content...",
        published = "Done."
      )

      lastStatus <- NULL
      repeat {
        path <- paste0("/revisions/", revisionId)
        revision <- withTokenRefreshRetry(GET, path)
        newStatus <- revision$status

        if (!isTRUE(newStatus == lastStatus)) {
```

(needs the `!isTRUE()` because `lastStatus` starts as `NULL`)

mslynch: Yeah, I don't think there's any reason to keep track of the index. I think when I wrote this I was thinking about receiving out-of-order updates, but we take care of all that on the backend.

I applied the suggestion here in my recent commit (Github wouldn't let me do it through the UI).

R/client-connectCloud.R
mslynch: see https://github.com/rstudio/rsconnect/pull/1219#discussion_r2429340514

R/servers.R
```diff
@@ -110,17 +110,20 @@ connectCloudUrls <- function() {
...
       ui = "https://staging.connect.posit.cloud",
-      auth = "https://login.staging.posit.cloud"
+      auth = "https://login.staging.posit.cloud",
+      logs = "https://logs.staging.connect.posit.cloud"
     ),
     development = list(
       api = "https://api.dev.connect.posit.cloud/v1",
       ui = "https://dev.connect.posit.cloud",
-      auth = "https://login.staging.posit.cloud"
+      auth = "https://login.staging.posit.cloud",
```
aronatkins, line 125: Could you confirm that "dev" using the staging login host?

aronatkins (approved):

R/servers.R
mslynch: Yes, that's correct. We only have a dev environment set up for our newer Connect Cloud infrastructure.

PR #1220, opened 2025-10-14
UTM query param

mslynch:
Add utm query param when opening connect cloud URLs. This helps us determine how many people are using rsconnect to publish.

Also tweaks some wording when we suggest the user to register their account.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,6 @@ Suggests:
     foreign,
     knitr,
     MASS,
-    mockery,
     plumber (>= 0.3.2),
     quarto,
     RCurl,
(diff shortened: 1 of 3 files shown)


R/utm.R
```diff
@@ -0,0 +1,17 @@
+getUtmValue <- function() {
+  if (Sys.getenv("RSTUDIO") == 1) {
```
karawoo:
```suggestion
  if (Sys.getenv("RSTUDIO") == "1") {
```

R/accounts.R
```diff
@@ -188,20 +188,24 @@ connectCloudUser <- function(launch.browser = TRUE) {
   authClient <- cloudAuthClient()
   deviceAuth <- authClient$createDeviceAuth()
 
+  verififcationUriComplete <- addUtmParameters(
```
aronatkins: `verificationUriComplete`?

Issue #1222, opened 2025-10-15
[BUG] Setting 'appVisibility' for Connect Cloud deployments does not modify the privacy settings on the published content


tyatposit:
We have support for this in connect cloud and it can be set after the fact in the UI.  Seems like this optional param should work when passed with the deployment.

`rsconnect::deployApp(appDir='/Repos/publisher-examples/vivid-storage/shiny-gallery', appVisibility='private')`

note:  I tried with and without quotes and it didn't change the result.

Related: #1216

PR #1224, opened 2025-10-15
better server incompatibility handling

mslynch:
Explicit changes to denote compatibility:
- https://github.com/rstudio/rsconnect/issues/1222
- `listAccountEnvVars` and `updateAccountEnvVars` error properly (fixed through the update to `isConnectServer`)
- syncAppMetadata skips connect cloud deployments
- applications() errors explicitly when called for connect cloud
- added new `checkConnectServer()` call in  `connectApiUser`, `connectUser`, `connectSPCSUser`, `removeServer`, `addServerCertificate`

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -501,12 +501,6 @@ deployApp <- function(
 
   # Change _visibility_ & set env vars before uploading contents
   if (isPositConnectCloudServer(accountDetails$server)) {
-    if (!is.null(appVisibility)) {
-      cli::cli_abort(
-        "Setting {.arg appVisibility} is not supported when deploying to Posit Connect Cloud."
-      )
-    }
-
     taskStart(quiet, "Updating content...")
     # Use appPrimaryDoc if available, otherwise fall back to inferredPrimaryFile
     primaryFile <- appMetadata$appPrimaryDoc %||%
@@ -693,6 +687,9 @@ needsVisibilityChange <- function(server, application, appVisibility = NULL) {
 
   if (isConnectServer(server)) {
     # Defaults to private visibility
+    return(FALSE)
+  }
+  if (isPositConnectCloudServer(server)) {
     return(FALSE)
   }
 


R/applications.R
```diff
@@ -39,6 +39,12 @@ applications <- function(account = NULL, server = NULL) {
   serverDetails <- serverInfo(accountDetails$server)
   client <- clientForAccount(accountDetails)
 
+  if (isPositConnectCloudServer(accountDetails$server)) {
```
aronatkins, line 42: Should we note the lack of support in the documentation?

R/applications.R
```diff
@@ -378,8 +384,10 @@ syncAppMetadata <- function(appPath = ".") {
   for (i in seq_len(nrow(deploys))) {
     curDeploy <- deploys[i, ]
 
-    # don't sync if published to RPubs
-    if (isRPubs(curDeploy$server)) {
+    # don't sync if published to RPubs or Connect Cloud
```
aronatkins, line 387: Alternatively, use two if clauses and avoid the complicated clause.

```r
if (isRPubs(curDeploy$server)) {
    next
} else if (isPositConnectCLoudServer(curDeploy$server)) {
    next
}
```

R/deployApp.R
```diff
@@ -501,6 +501,12 @@ deployApp <- function(
 
   # Change _visibility_ & set env vars before uploading contents
   if (isPositConnectCloudServer(accountDetails$server)) {
+    if (!is.null(appVisibility)) {
```
aronatkins: Note: This is different than how we handle Connect, which is simply ignored by `needsVisibilityChange()`.

The documentation states the following, which implies that ignoring is acceptable.

> Currently has an effect only on deployments to shinyapps.io.

R/applications.R
mslynch: Yeah. Ingrid is prepping a separate PR with more documentation updates.

Issue #1227, opened 2025-10-18
Update documentation to explicitly call out supported servers


rodriin:
Update documentation to explicitly call out supported servers at the very top for all exposed functions.

PR #1228, opened 2025-10-18
Updating the documentation with supported server info

rodriin:
Resolves: #1227

This PR adds the supported server info to the documentation for all exposed functions at the very top of the description area. In the cases when the function specifically supports only one type of server it will say the server type. For functions that are supported for all servers it will say: "All servers".

### DIFF
--- a/R/accounts.R
+++ b/R/accounts.R
@@ -38,8 +38,6 @@ accounts <- function(server = NULL) {
 #' Register account on Posit Connect
 #
 #' @description
-#' Supported servers: Posit Connect servers
-#'
 #' `connectUser()` and `connectApiUser()` connect your Posit Connect account to
 #' the rsconnect package so that it can deploy and manage applications on
 #' your behalf.
@@ -48,6 +46,8 @@ accounts <- function(server = NULL) {
 #' authenticate in-browser to your Posit Connect server. `connectApiUser()` is
 #' appropriate for non-interactive settings; you'll need to copy-and-paste the
 #' API key from your account settings.
+#'
+#' Supported servers: Posit Connect servers
 #'
 #' @param account A name for the account to connect.
 #' @param server The server to connect to.
--- a/man/connectApiUser.Rd
+++ b/man/connectApiUser.Rd
@@ -30,8 +30,6 @@ interactive sessions only. If a function is passed, it will be called
 after the app is started, with the app URL as a parameter.}
 }
 \description{
-Supported servers: Posit Connect servers
-
 \code{connectUser()} and \code{connectApiUser()} connect your Posit Connect account to
 the rsconnect package so that it can deploy and manage applications on
 your behalf.
@@ -40,6 +38,8 @@ your behalf.
 authenticate in-browser to your Posit Connect server. \code{connectApiUser()} is
 appropriate for non-interactive settings; you'll need to copy-and-paste the
 API key from your account settings.
+
+Supported servers: Posit Connect servers
 }
 \seealso{
 Other Account functions: 


R/deployTFModel.R
```diff
@@ -1,5 +1,8 @@
 #' Deploy a TensorFlow saved model
 #'
+#' @description
+#' Supported servers: Posit Connect, ShinyApps, and RPubs servers
```
aronatkins: RPubs only supports R Markdown documents; it definitely does not support TensotFlow.

R/accounts.R
```diff
@@ -1,8 +1,12 @@
 #' Account Management Functions
 #'
+#' @description
+#' Supported servers: All servers
```
aronatkins: According to https://roxygen2.r-lib.org/articles/rd.html#the-introduction, the description needs to briefly describe what the function does. Information about server support feels like it belongs in the details, and perhaps at the end of the function details.

@karawoo - Do you have a different suggestion?

rodriin: @aronatkins I purposely placed it on top of the description so that it would be visible right away to anyone trying to use the function. I'm willing to bring it down to the details if that is preferable by all.

R/deployTFModel.R
rodriin: Thank you for catching this, I will fix it in multiple places.

R/accounts.R
aronatkins: @rodriin - I understand the motivation.

I'm trying to keep us roughly following the R documentation guidance, where the description is supposed to describe what the function does using only one paragraph / a few lines.

Reading this guidance makes me feel like server information may want to describe server compatibility in the description or possibly as a note.

https://cran.r-project.org/doc/manuals/R-exts.html#Documenting-functions

karawoo: I would definitely expect the first paragraph of the description section to describe what the function does. I'd be ok with keeping the info on server compatibility in that section, but as a second paragraph after the description of the function.

R/auth.R
```diff
@@ -29,6 +29,11 @@ cleanupPasswordFile <- function(appDir) {
   invisible(TRUE)
 }
 
+#' Add authorized user to application
+#'
+#' @description
+#' Supported servers: ShinyApps servers
+#'
 #' Add authorized user to application
```
karawoo, line 35: Here and elsewhere, can we update so the title and one paragraph of the description don't exactly duplicate one another?

R/accounts.R
rodriin: Ok, I'll go with adding the server compatibility as the 2nd or last thing on the description section.

R/auth.R
rodriin: @karawoo The only reason I had to duplicate the description from the title in these cases is because the existing documentation did not contain anything for the description hence the generated docs came out with those things duplicated. I'm not sure I have all the knowledge to populate the specific missing descriptions. Maybe that can be done on a separate PR by someone more familiar with the existing functionality?

rodriin: Also if I didn't do the duplication for all the cases missing the description, then only the supported server info would be in the description area which didn't feel right since I was also trying to have the generated docs match the existing docs and only have the extra info I was adding for the supported servers.

R/deployTFModel.R
rodriin: Addressed on commit: https://github.com/rstudio/rsconnect/pull/1228/commits/23a3a02242b6b772210caa9548e4d21a2ce98314

R/accounts.R
rodriin: Addressed on commit: https://github.com/rstudio/rsconnect/pull/1228/commits/72bf9b75345f69b0534f72563e322012c197524e

R/accounts.R
```diff
@@ -34,6 +38,8 @@ accounts <- function(server = NULL) {
 #' Register account on Posit Connect
 #
 #' @description
+#' Supported servers: Posit Connect servers
```
aronatkins: This is still the leading text for "Description".

image


The supported servers should move to the end of the description section.

Screenshot taken from locally installed documentation for `?rsconnect::connectUser`

R/auth.R
```diff
@@ -185,6 +200,11 @@ showUsers <- function(
 
 #' List invited users for an application
 #'
+#' @description
```
aronatkins, line 203: I'm OK with this duplication. It mirrors the implied duplication we had previously. It does point out where we lack sufficient documentation, but you're not addressing that, here.

image

(screenshot from https://rstudio.github.io/rsconnect/reference/showInvited.html)

aronatkins (approved):
One misplaced update, but otherwise looks good. Thanks for this exhaustive update!

R/accounts.R
rodriin: Thank you for catching this one, I missed it. I'll fix it now.

rodriin: Addressed in commit: https://github.com/rstudio/rsconnect/pull/1228/commits/14af74bb7cf41985814f6a5ee366b90ff36d78b8

PR #1229, opened 2025-10-20
pass origin to as.POSIXct

mslynch:
I don't really understand what's happening here, but on oldrel-3 and oldrel-4 it errors because ` as.POSIXct` wants to be passed `origin`.

### DIFF
--- a/tests/testthat/test-client-connectCloud.R
+++ b/tests/testthat/test-client-connectCloud.R
@@ -198,17 +198,12 @@ test_that("awaitCompletion handles failure with logs", {
   stderr_text <- paste(stderr_output, collapse = "\n")
   info <- paste0("stderr_text was:\n", stderr_text)
 
-  expect_grepl <- function(pattern, text) {
-    info <- paste0("Did not find pattern '", pattern, "' in text:\n", info)
-    expect_true(grepl(pattern, text), info = info)
-  }
-
-  expect_grepl("Begin Publishing Log", stderr_text)
-  expect_grepl("End Publishing Log", stderr_text)
-  expect_grepl("Starting deployment...", stderr_text)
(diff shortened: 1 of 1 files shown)


R/client-connectCloud.R
```diff
@@ -251,7 +251,8 @@ connectCloudClient <- function(service, authInfo) {
                     for (log_entry in logs$data) {
                       local_timestamp <- as.POSIXct(
                         # Convert to seconds
-                        log_entry$timestamp / 1e6
+                        log_entry$timestamp / 1e6,
+                        origin = "1970-01-01",
```
aronatkins, line 255: From the R 4.5.1 documentation for `as.POSIXct`:

> Optional since R 4.3.0, where the equivalent of "1970-01-01" is used.

tests/testthat/test-client-connectCloud.R
```diff
@@ -196,13 +196,19 @@ test_that("awaitCompletion handles failure with logs", {
...
   stderr_text <- paste(stderr_output, collapse = "\n")
+  info <- paste0("stderr_text was:\n", stderr_text)
 
-  expect_true(grepl("Begin Publishing Log", stderr_text))
-  expect_true(grepl("End Publishing Log", stderr_text))
-  expect_true(grepl("Starting deployment...", stderr_text))
-  expect_true(grepl("Your app is busted!!", stderr_text))
-  expect_true(grepl("INFO:", stderr_text))
-  expect_true(grepl("ERROR:", stderr_text))
+  expect_grepl <- function(pattern, text) {
```
aronatkins: Is `testthat::expect_match` appropriate here?
https://testthat.r-lib.org/reference/expect_match.html

aronatkins (approved):

Issue #1236, opened 2025-10-22
Give the user a choice for content marked for deletion in PCC


rodriin:
### Steps & Current Behavior

1. Deploy content to PCC using `rsconnect`
2. Go to PCC and delete the content in PCC
3. Try to redeploy the same content to PCC using `rsconnect`
4. Deployment fails with a 409 error saying:
> Content 019a0d09-8801-4c0b-e700-4d83728cb35a is marked for deletion.

### Expected Behavior

Handle the error in `rsconnect` to give the user the choice on how to proceed similarly to how it is handled when deploying an app in a ShinyApps.io server:

![Image](https://github.com/user-attachments/assets/10448233-54df-4dc7-a13c-8a5ed8a77792)

mslynch:
`error_type` in this case is `resource_pending_deletion`

rodriin:
The 409 error will get converted into a 404 error in `rsconnect` so the `deployApp` code will then handle the error as intended by default, which means the prompt in the screenshot will be displayed next.

tyatposit:
verified for `deployApp()` function ONLY

- User is now prompted with the option as expected on redeploy after deletion in CC of previous content.
- Both selection options work as intended.
- Confirmed I'm able to deploy again successfully without being blocked after deleting and republishing.
- Other servers' scenarios still function as before.

NOTE:  The same issue still exists when deploying other content using the other deployment functions: `deployDoc()`, `deploySite()`, etc.  Those still run into the 409.

PR #1238, opened 2025-10-22
handle 409 for pending deletion the same way as a 404

mslynch:
When we delete a piece of content, there's a ~5 minute window where the content hasn't been fully deleted. When that's the case, we're not able to redeploy to it - we want to present the user the same prompt as when their app has been fully deleted.

Resolves https://github.com/rstudio/rsconnect/issues/1236.

### DIFF
--- a/R/client-connectCloud.R
+++ b/R/client-connectCloud.R
@@ -141,7 +141,17 @@ connectCloudClient <- function(service, authInfo) {
 
     getContent = function(contentId) {
       path <- paste0("/contents/", contentId)
-      withTokenRefreshRetry(GET, path)
+      content <- withTokenRefreshRetry(GET, path)
+      if (content$state == "deleted") {
+        cli::cli_abort(
+          "Content is pending deletion.",
+          class = c(
+            "rsconnect_http_404",
+            "rsconnect_http"
+          )
+        )
+      }
+      content
     },
 
     updateContent = function(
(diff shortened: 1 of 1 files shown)


rodriin (approved):
Looks pretty straight forward, treating it as a 404 error.

R/client-connectCloud.R
```diff
@@ -173,7 +173,23 @@ connectCloudClient <- function(service, authInfo) {
         )
       )
 
-      content <- withTokenRefreshRetry(PATCH_JSON, path, json)
+      content <- tryCatch(
+        withTokenRefreshRetry(PATCH_JSON, path, json),
+        rsconnect_http_409 = function(e) {
```
aronatkins: Should this be `err` not `e`? Additionally, is it worth adding `parent=err` to the `cli_abort()`?

mslynch: Something about this wasn't working right... But today I realized I can make the check sooner in getContent instead. Seems to be working fine with that change.

R/client-connectCloud.R
```diff
@@ -141,7 +141,17 @@ connectCloudClient <- function(service, authInfo) {
 
     getContent = function(contentId) {
       path <- paste0("/contents/", contentId)
-      withTokenRefreshRetry(GET, path)
+      content <- withTokenRefreshRetry(GET, path)
+      if (content$state == "deleted") {
+        cli::cli_abort(
+          "Content is pending deletion.",
```
karawoo, line 147: At this point couldn't the deletion also be complete (not pending)? Not that it matters _too_ much since this message will most likely be `tryCatch`-ed

mslynch: No - once deletion finishes, the content is fully deleted from the database and the API will return 404.

karawoo: ah got it, makes sense, thanks

Issue #1235, opened 2025-10-22
Fix the showLogs function to fail gracefully for Connect Cloud


rodriin:
### Current Behavior

Try using `showLogs` for a PCC content:
```
> rsconnect::showLogs("rsconnect1", account="samperman")
Error in `rsconnect::showLogs()`:
! No application found
ℹ Specify the application directory, name, and/or associated account.
Run `rlang::last_trace()` to see where the error occurred.
```

### Expected Behavior

`showLogs` should say display the "`server` must be shinyapps.io" error message instead.

tyatposit:
verified.

- confirmed we have a proper error now when attempting against a pcc deployment:


```
Error in `rsconnect::showLogs()`:
! `server` must be shinyapps.io
Show Traceback
> rsconnect::showLogs(appPath = getwd())
Error in `rsconnect::showLogs()`:
! `server` must be shinyapps.io
Show Traceback
```

PR #1239, opened 2025-10-23
graceful showLogs failure for connect cloud

mslynch:
Resolves https://github.com/rstudio/rsconnect/issues/1235.

### DIFF
--- a/R/applications.R
+++ b/R/applications.R
@@ -297,11 +297,7 @@ showLogs <- function(
     account = account
   )
 
-  if (isPositConnectCloudServer(deployment$server)) {
-    cli::cli_abort(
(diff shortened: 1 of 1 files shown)


R/applications.R
```diff
@@ -296,6 +296,13 @@ showLogs <- function(
     server = server,
     account = account
   )
+
+  if (isPositConnectCloudServer(deployment$server)) {
```
aronatkins:
```suggestion
  if (!isShinyappsServer(deployment$server)) {
```

What about producing an error for other server types? You could apply the same approach to other shinyapps.io-only functions.

You could also use `checkShinyappsServer(server)`.

mslynch: Yeah, `checkShinyappsServer(server)` is much better.

PR #1252, opened 2025-10-29
fix: registerUserToken defaults for new arguments

### LINKED ISSUE #1250: RStudio account registration broken in 1.6.0
`registerUserToken` gained arguments without defaults

```
Error in rsconnect:::registerUserToken(serverName = serverName, accountName = accountName,  : 
  argument "accessToken" is missing, with no default
```

Called by RStudio.
https://github.com/rstudio/rstudio/blob/main/src/cpp/session/modules/SessionRSConnect.R

Neither `accessToken` nor `refreshToken` have default values and RStudio / Workbench do not know about this change (already-released RStudio only ever knows about the old signature).

https://github.com/rstudio/rsconnect/blob/f0938ee844cc335fbce7ff9a73edfa8911da9fef/R/ide.R#L49-L67

Arrived in https://github.com/rstudio/rsconnect/pull/1217/files#diff-ae71c8761e556da984fb942359a2ca6a5849c4c3f83f58098cbe52357d8b61be

aronatkins:
fixes #1250

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # rsconnect (development version)
 
+* Fix account registration from RStudio. (#1250)
+
 * SPCS/Snowflake authentication supports Connect API keys for user
   identification. The `connectSPCSUser()` function now requires an `apiKey`
   parameter, and the API key is included in the `X-RSC-Authorization` header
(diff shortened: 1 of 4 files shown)


mslynch (approved):

aronatkins:
Confirmed that Connect account configuration in the RStudio UI succeeds with this change.

PR #1264, opened 2025-11-26
Allow deploying from manifest.json files

### LINKED ISSUE #1259: Using a manifest.json file with the deploy*() functions. Docs unclear.
Initially I expected that the manifest file created using `writeManifest()` would be the thing that gets used when deploying an app.

My intended workflow was to have a script generating a manifest.json file, essentially constructing a customized list of files to be included in the bundle before calling `writeManifest(appFiles=files)`.

After trying to deploy a shiny app with `deployApp()` to a Posit connect server and getting some weird errors, I downloaded the bundle and quickly noticed that the manifest wasn't the one I generated. The unintended inclusion of files was the cause of these errors. 
The docs are really unclear on this. What is the purpose of generating a custom `manifest.json` and checking it into my repo if `deployApp()` regenerates it from scratch anyway?

Is there anyway to actually deploy based on an existing manifest file?
The python package `rsconnect-python` seems to enable this in a straightforward way, but I cannot  figure it out here.

jonkeane:
Resolves #1259

I initially tried to make this a separate function rather than an argument to `deployApp()` (yet another additional!). But it turned out the logic for bundle creation, etc. is too intertwined. It would be good to disentangle that such that a separate function is possible, but separately from this more tactical PR.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -363,9 +363,8 @@ deployApp <- function(
       ))
     }
 
-    manifestJson <- readLines(manifestPath, warn = FALSE, encoding = "UTF-8")
     manifest <- tryCatch(
-      jsonlite::fromJSON(manifestJson, simplifyVector = FALSE),
+      jsonlite::read_json(manifestPath, simplifyVector = FALSE),
       error = function(e) {
         cli::cli_abort(c(
           "Failed to parse manifest file.",


R/deployApp.R
```diff
@@ -333,7 +348,81 @@ deployApp <- function(
...
+      cli::cli_abort(c(
+        "Manifest file not found.",
+        "x" = "Could not find {.file {manifestPath}}.",
+        "i" = "Create a manifest with {.fn writeManifest} first."
+      ))
+    }
+
+    manifestJson <- readLines(manifestPath, warn = FALSE, encoding = "UTF-8")
+    manifest <- tryCatch(
+      jsonlite::fromJSON(manifestJson, simplifyVector = FALSE),
```
karawoo: could we use `jsonlite::read_json` here and avoid having to read the file and parse the json as two separate steps?

R/deployApp.R
```diff
@@ -816,21 +907,35 @@ bundleApp <- function(
...
+  # generate or use existing manifest
+  if (!is.null(existingManifest)) {
+    logger("Using existing manifest: ", existingManifest)
+    manifest <- existingManifest
+
+    # Apply envManagement overrides if provided
+    if (!is.null(envManagement)) {
+      envManagementR <- envManagement
+      envManagementPy <- envManagement
+    }
```
karawoo: could you say a little more about this? I'm not sure I follow this change and whether it's necessary to deploy from a manifest.

jonkeane: Ah no this is cruft from the LLM, sorry I should have removed it.

In the original proposal some of the arguments were allowed to override what's in the manifest. But I don't actually think that's a good idea — if the values in the manifest are wrong, just change them in the manifest!

R/deployApp.R
jonkeane: Yes, good catch.

karawoo (approved):
I don't think the `jsonlite::read_json` change got pushed, but otherwise looks good (and that's just a nit anyway). I tried this locally and was able to deploy successfully with a pre-existing manifest ✅

PR #1272, opened 2025-12-11
Add manifest-file override for package repository resolution strategy

jonkeane:
This adds support for Connect's new feature of manifest-file overrides for package resolution strategies. Most useful for sending apps with `packageRepositoryResolutionR = "lockfile"` which will use all and only the repositories in the manifest / renv.lock file.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 * Resolved a bug where `renv.lock` files that had multiple repositories were not
   being translated faithfully when creating the manifest file. (#1268)
 
+* Added support for overriding R package repository resolution behavior. 
+
 # rsconnect 1.7.0
 
 * Added support for deploying from `manifest.json` files created by


R/bundle.R
```diff
@@ -185,6 +186,20 @@ createAppManifest <- function(
     }
   }
 
+  # Validate packageRepositoryResolutionR
+  if (!is.null(packageRepositoryResolutionR)) {
+    validValues <- c("lax", "strict", "legacy", "lockfile")
```
toph-allen: I love this addition! I missed that this was something new in Connect.

toph-allen (approved):
This looks good! Thank you.

R/bundle.R
```diff
@@ -185,6 +186,20 @@ createAppManifest <- function(
...
+    if (!packageRepositoryResolutionR %in% validValues) {
+      stop(
+        "Invalid packageRepositoryResolutionR value: '",
+        packageRepositoryResolutionR,
+        "'. ",
+        "Must be one of: ",
+        paste(validValues, collapse = ", ")
+      )
+    }
```
karawoo: nit that you can take or leave: I might simplify this to 
```suggestion
    match.arg(packageRepositoryResolutionR, c("lax", "strict", "legacy", "lockfile"))
```

karawoo (approved):

R/bundle.R
jonkeane: Thanks! I'll deal with the test failures in a bit

karawoo (approved):
lgtm!

nealrichardson:
Have you tested this with a live Connect server? #1276 makes me think it won't work until the APIs are switched to v1. (I have a branch locally for that, which is based off of my other 2 PRs, so I'll push that up once those merge.)

R/bundle.R
```diff
@@ -185,6 +186,14 @@ createAppManifest <- function(
     }
   }
 
+  # Validate packageRepositoryResolutionR
+  if (!is.null(packageRepositoryResolutionR)) {
+    match.arg(
+      packageRepositoryResolutionR,
+      c("lax", "strict", "legacy", "lockfile")
```
nealrichardson, line 193: Is there's a reason the default values aren't specified in the signature, the "usual way" when using `match.arg()`? (I would believe it if you told me there was a reason, it's just surprising.)

jonkeane:
> Have you tested this with a live Connect server? https://github.com/rstudio/rsconnect/issues/1276 makes me think it won't work until the APIs are switched to v1. (I have a branch locally for that, which is based off of my other 2 PRs, so I'll push that up once those merge.)

I did when developing it, but I will try again and confirm before merging.

R/bundle.R
jonkeane: Ah no they should be. Let me fix that

karawoo: If you put them in the signature then the arg won't default to NULL

jonkeane: Ah yeah, you're right Kara. I should add a test that confirms that though.

karawoo:
It did work for me locally as well in the sense that I could deploy, but if integrations are included in the manifest they don't get set on deploy.

PR #1274, opened 2025-12-12
ci: Add basic integration test with many Connect versions

### LINKED ISSUE #1253: Add integration tests to CI
`rsconnect` has no integration tests. This makes it harder to do things like migrate to v1 APIs (#1155) with confidence.

We could use the [`with-connect` CI helper](https://github.com/posit-dev/with-connect) to add some basic integration tests.

nealrichardson:
Closes #1253

Adds a new test suite that does not run in regular checks, which has a simple test that a Shiny app can be deployed successfully, then updated. New GitHub Actions workflow using [with-connect](https://github.com/posit-dev/with-connect) to test across a range of Connect server versions. Setup helper registers account and server and tries to clean up everything when the suite is done. 

While a simple test, it covers nearly all of the APIs in Connect we use. Locally, I ran a variation of the command in `with-connect` using covr:

```
Rscript -e 'saveRDS(covr::package_coverage(type="none", code="testthat::test_dir(\"tests/integration\", package=\"rsconnect\", load_package=\"installed\")"), file="connect-integration-coverage.rds")'
```

and analyzed the data with `covr::report()`. These tests alone hit 31% of `rsconnect`'s code, but 91% of client-connect.R (when including the deletions in #1275). The only lines not covered are two unused functions (`service()` and `serverSettings()`) and two auth functions we can't reasonably test (`addToken()` and `getSnowflakeAuthToken()`). 

This should give some assurance that we haven't broken something terribly when refactoring the API client layer, or when APIs in Connect change. We could add more if we want, but we don't need to test all combinations of features--there are tests in Connect that make sure we can deploy all kinds of content, and unit tests in the package that (presumably) confirm all of the detection and bundle-making on this side.

### DIFF
--- a/.github/workflows/connect-integration.yaml
+++ b/.github/workflows/connect-integration.yaml
@@ -34,19 +34,18 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - uses: r-lib/actions/setup-pandoc@v2
-
       - uses: r-lib/actions/setup-r@v2
         with:
-          r-version: '4.4'
           use-public-rspm: true
-
-      - uses: quarto-dev/quarto-actions/setup@v2
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: local::., Biobase=?ignore, BiocManager=?ignore
-          needs: check
+          # Only install what we need for this test, which is not all of Suggests
+          dependencies: '"hard"'
+          extra-packages: |
+            local::.
+            any::testthat
+            any::shiny
 
       - name: Run integration tests
         uses: posit-dev/with-connect@main


.github/workflows/connect-integration.yaml
```diff
@@ -0,0 +1,57 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - "preview"   # Nightly builds of Connect
+          - "release"   # special value that always points to the latest Connect release
+          - "2025.09.0" # jammy
+          - "2025.03.0" # jammy (every 6 months, just bc every release would be overkill)
+          - "2024.09.0" # jammy
+          # These versions use R 4.2 and seem to take forever to install R dependencies
```
nealrichardson, line 22: FTR these did eventually pass, but probably not worth 20 minutes of CI time for things that don't change. If things pass on 2023.03.0, they should pass on these versions too.

https://github.com/rstudio/rsconnect/actions/runs/20179566092/job/57936324677?pr=1274

nealrichardson:
The Windows failure seems like a flake, it passed on previous runs, and there is nothing in this PR that affects the regular `R CMD check` flow.

.github/workflows/connect-integration.yaml
```diff
@@ -0,0 +1,57 @@
...
+      R_KEEP_PKG_SOURCE: yes
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: r-lib/actions/setup-pandoc@v2
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: '4.4'
```
nealrichardson: Was aiming to run locally on an R version that would be present on the majority of the docker images in the matrix, but it appears to fall back to a version that is present on Connect (hence 4.2 in many cases), so maybe it doesn't matter.

.github/workflows/connect-integration.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: local::., Biobase=?ignore, BiocManager=?ignore
+          needs: check
+
+      - name: Run integration tests
+        uses: posit-dev/with-connect@main
+        with:
+          version: ${{ matrix.version }}
+          license: ${{ secrets.CONNECT_LICENSE_FILE }}
```
nealrichardson, line 54: I added this secret; the license expires 2026-12-12 and will need to be replaced then.

.github/workflows/connect-integration.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: actions/checkout@v6
+
+      - uses: r-lib/actions/setup-pandoc@v2
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: '4.4'
+          use-public-rspm: true
+
+      - uses: quarto-dev/quarto-actions/setup@v2
```
nealrichardson: This is probably not needed here because the test app doesn't use Quarto, just copied it from the main workflow.

.github/workflows/connect-integration.yaml
```diff
@@ -0,0 +1,57 @@
...
+    name: Connect ${{ matrix.version }}
+
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      R_KEEP_PKG_SOURCE: yes
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: r-lib/actions/setup-pandoc@v2
```
nealrichardson: Also probably not needed.

.github/workflows/connect-integration.yaml
```diff
@@ -0,0 +1,56 @@
...
+          use-public-rspm: true
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          # Only install what we need for this test, which is not all of Suggests
+          dependencies: '"hard"'
+          extra-packages: |
+            local::.
+            any::testthat
+            any::shiny
```
jonkeane, line 48: Here's an example where having a dev dependency description would be lovely. 🫠 But yeah this is right for installing rsconnect, testthat, and shiny (though I did need to go look up that `local::.` and `dependencies: '"hard"'` mean that)

tests/integration/test-deploy.R
```diff
@@ -0,0 +1,13 @@
+test_that("deploy does not error", {
+  expect_true(deployApp("example-shiny", appTitle = "Test", account = account))
+})
+
+test_that("re-deploy does not error", {
+  # Let's exercise the env vars as well
+  expect_true(deployApp("example-shiny", envVars = "TEST", account = account))
```
jonkeane, line 7: The `envVars` exercise isn't require to redeploy, just added on to exercise a bit more code, yeah?

tests/integration/setup.R
```diff
@@ -0,0 +1,51 @@
...
+)
+withr::defer(
+  {
+    removeAccount(account)
+    removeServer(account)
+    # also clean up any artifacts left behind
+    # do it twice, first to remove files, second to remove empty dirs
+    file.remove(
+      grep(
+        "rsconnect/testing20",
```
jonkeane, line 36: Any reason to limit this to 2099? I doubt this particular code will be running then, but do we expect any other `testing` folders there? But also: could we use the actual value of `account` here? Would the `defer` resolve that? I would imagine it would, but the environment search with defer might not?

tests/integration/test-deploy.R
nealrichardson: Correct

tests/integration/setup.R
nealrichardson: Laziness, plus iteratively trying to clean up from runs where the cleanup didn't work (before I got it working to the extent that it does now). 

We could probably use the value of `account`, but I kinda like how this should generally clean up any past test runs you've done, and I don't think it's harmful.

jonkeane: This merged already, we can deal with this in 75 years if this code is still being called, but would `testing` or `testing2` be even broader (but not too broad)?

nealrichardson: 🤷 feel free to change it, either way seems fine

jonkeane: If one of us is still here in 2100 we can deal with it.

PR #1275, opened 2025-12-12
chore: delete unused UserRecord functions

nealrichardson:
Noticed these were not referenced anywhere, they aren't exported, and I could not find any references to them outside the package.


jonkeane (approved):
Thanks for this cleanup. If tests pass I'm good with merging, if we need to bring it back we have the git history

Issue #1278, opened 2025-12-15
refactor(connect): use v1 API for currentUser


nealrichardson:
This is probably safe because I _think_ all we ever actually use in the Connect code paths is the `username` field.

dotNomad:
Currently `GET /users/current` is called, Connect rewrites it to `/me/ide` which returns the user record plus the integer `id`. The v1 endpoint returns `guid` and no `id`.

There are two spots we need to change:

### `rsconnect`

The `account_id` app listing filter. All of the Connect registration paths write `accountId = user$id` to the account `.dcf`. That is then used in the `account_id` filter to list content they own. We can use the `account_guid` filter for this replacement.

### RStudio IDE

`getUserFromRawToken()` returns the current user to RStudio which expects and validates on the `id` being part of the user.

---

We can move `rsconnect` over to the new v1 API for CLI-only registration paths (the paths would need to split where they call `currentUser`) with backwards compatibility fallback for existing `.dcf` files that currently use `accountId`.

Then we can update RStudio to use `guid`, with `rsconnect` accepting both the old `userId` and the GUID, and RStudio would need to accept both payloads. That way we can ship both independently.

PR #1280, opened 2025-12-15
refactor(connect): use v1 APIs for create/upload/deploy

### LINKED ISSUE #1277: refactor(connect): use v1 APIs for create/upload/deploy
These are pretty straightforward to swap and are covered with basic integration tests in #1274

nealrichardson:
Closes #1277

Connect integration tests cover all lines modified, so we can have assurance that the APIs are being called and not erroring. Having made this change recently in `rsconnect-python`, I believe this is safe, and that the v1 APIs (I was going to say "new APIs" but they've actually been there for years) work close enough to the previous ones, at least in the ways they're being used here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,9 @@
 
 * Added support for overriding R package repository resolution behavior.
 
+* Upgraded to use `v1` APIs for deploying to Connect servers, which enables
+  new features for specifying settings in the manifest file. (#1280)
+
 * Removed several functions, including `addConnectServer()` and
   `discoverServer()`, as well as HTTP backends other than libcurl,
   which were deprecated in rsconnect 1.0.0. (#1282)


R/client-connect.R
```diff
@@ -15,19 +15,26 @@ connectClient <- function(service, authInfo) {
...
     serverSettings = function() {
-      GET(service, authInfo, file.path("/server_settings"))
+      GET(service, authInfo, unversioned_url("server_settings"))
     },
 
     ## User API
 
     currentUser = function() {
-      GET(service, authInfo, "/users/current")
+      # All callers only need $id and $username,
```
nealrichardson, line 24: This comment is just some research for #1278

R/client-connect.R
```diff
@@ -15,19 +15,26 @@ connectClient <- function(service, authInfo) {
...
+      # username is used to identify the "account", so this should be safe
+      # to upgrade to v1.
+      GET(service, authInfo, unversioned_url("users", "current"))
     },
 
     ## Tokens API
 
     addToken = function(token) {
-      POST_JSON(service, authInfo, "/tokens", token)
+      POST_JSON(service, authInfo, unversioned_url("tokens"), token)
```
nealrichardson, line 37: API not changed, just using the `unversioned_url()` wrapper, similar to what we do in `connectapi`. Not only does it centralize and standardize the URL construction logic, which had a couple of different flavors in this file alone, it makes it easier to grep the codebase to see which calls are using v1 APIs and which aren't.

R/client-connect.R
```diff
@@ -61,18 +68,20 @@ connectClient <- function(service, authInfo) {
...
+      result <- POST_JSON(service, authInfo, v1_url("content"), details)
       list(
-        id = application$id,
-        guid = application$guid,
-        url = application$url
+        id = result$id,
+        guid = result$guid,
+        url = result$content_url,
+        # Include dashboard_url so we can open it or logs path after deploy
+        dashboard_url = result$dashboard_url
```
nealrichardson, line 79: By including `dashboard_url` here, we can construct the URLs that we were previously fetching from a "config" endpoint. Presence of "dashboard_url" in the response here marks it as being from Connect.

R/client-connect.R
```diff
@@ -61,18 +68,20 @@ connectClient <- function(service, authInfo) {
...
       )
     },
 
-    uploadApplication = function(appId, bundlePath) {
-      path <- file.path("/applications", appId, "upload")
+    uploadBundle = function(contentGuid, bundlePath) {
```
nealrichardson, line 83: Function is renamed and swapped to taking guid. Other similar functions for shinyapps and Connect Cloud already refer to bundles.

R/client-connect.R
```diff
@@ -83,42 +92,29 @@ connectClient <- function(service, authInfo) {
...
-        path <- paste0(
-          "/v1/tasks/",
-          taskId,
-          "?first=",
-          first,
-          "&wait=",
-          wait
-        )
-        response <- GET(service, authInfo, path)
+        # ick, manual url construction
```
nealrichardson, line 113: I started refactoring this because I thought we could just let `httr::GET()` handle the query string assembly, but as it turns out, `GET()` here is not from `httr`.

R/client.R
```diff
@@ -123,7 +123,7 @@ isContentType <- function(x, contentType) {
   grepl(contentType, x, fixed = TRUE)
 }
 
-
+# These functions should move to client-shinyapps.R
```
nealrichardson, line 126: Just an observation, everything else is stashed in a `client` object, except this one and the function below it. Not touching the shinyapps code in this PR though.

R/configureApp.R
```diff
@@ -62,13 +62,7 @@ configureApp <- function(
     propertyValue <- properties[[i]]
 
     # dispatch to the appropriate client implementation
-    if (is.function(client$configureApplication)) {
```
nealrichardson, line 65: This is a shinyapps function, but the only `configureApplication()` method was in the Connect client, and it didn't take these arguments (and didn't actually configure anything). Deleted.

R/deployApp.R
```diff
@@ -965,13 +965,12 @@ openURL <- function(
...
-    if (!deploymentSucceeded && validURL(config$logs_url)) {
-      # With 1.5.5+, Connect application configuration includes
-      # a logs URL to be shown on unsuccessful deployment.
-      url <- config$logs_url
+  if (!is.null(application$dashboard_url)) {
+    # Posit Connect should land here because it inserts a dashboard_url
+    if (deploymentSucceeded) {
+      url <- paste(application$dashboard_url, "access", sep = "/")
+    } else {
+      url <- paste(application$dashboard_url, "logs", sep = "/")
```
nealrichardson, line 963: This is literally what the old `config` endpoint does, returns these two URLs.

R/client-connect.R
jonkeane: This might be a separate PR regardless, but IIUC one of the reasons we rolled our own `GET` was to support configurations that are no longer a real issue (IIRC, though my searching just now didn't turn up anything elucidating, there was a period where we needed to support less-than-standardly-secure http connections). I did find https://github.com/rstudio/rsconnect/pull/762 which started reducing the backends...)

nealrichardson: Interesting read. I made #1281 as a next step there.

PR #828, opened 2023-05-01
Revise quarto argument

hadley:
I'm assuming that the previous interface of supplying a path is mostly an accident of history (not a desired UI) and hasn't been used to select between different versions of quarto.

Fixes #658

PR #1282, opened 2025-12-16
refactor: remove alternate http backends and other deprecated functions

### LINKED ISSUE #1281: Remove non-libcurl backends
These were deprecated in #762 (March 2023), and the news bullet said "The old backends will remain available for at least 2 years." We're coming up on 3 years now. There are no open issues that suggest problems with libcurl, so they should be safe to remove. 

Maybe once we axe those, we can migrate to `httr2` and offload even more responsibility to well tested and maintained libraries.

nealrichardson:
Closes #1281 and then some

I don't see any of these functions referenced in https://github.com/search?q=repo%3Arstudio%2Frstudio%20%22rsconnect%3A%3A%22&type=code so I presume this is safe, but maybe that's not enough assurance that nothing here would affect usage in RStudio.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,11 @@
 * Resolved a bug where `renv.lock` files that had multiple repositories were not
   being translated faithfully when creating the manifest file. (#1268)
 
-* Added support for overriding R package repository resolution behavior. 
+* Added support for overriding R package repository resolution behavior.
+
+* Removed several functions, including `addConnectServer()` and
+  `discoverServer()`, as well as HTTP backends other than libcurl,
+  which were deprecated in rsconnect 1.0.0. (#1282)
 
 # rsconnect 1.7.0
 
@@ -11,9 +15,9 @@
   `writeManifest()`: use the `manifestPath` argument of `deployApp()` and related
   functions to specify the path to an existing manifest file. (#1259)
 
-* `urlEncode()` now uses `curl::curl_escape()` instead of `RCurl::curlEscape()`, 
+* `urlEncode()` now uses `curl::curl_escape()` instead of `RCurl::curlEscape()`,
   as RCurl is a Suggests dependency. (#1265)
-  
+
 * The `User-Agent` header in requests made from rsconnect will now be of the
   format `RSConnect/x.y.z` instead of `rsconnect/x.y.z` in order to satisfy web
   application firewalls that enforce Pascal case.
@@ -61,7 +65,7 @@
 
   Existing Posit Cloud account records may be removed by using
   `removeAccount("yourname", "posit.cloud")`.
-  
+
   Existing Posit Cloud deployment records may be removed by using
   `forgetDeployment(name="deployment", account="yourname", server="posit.cloud")`.
 


R/applications.R
```diff
@@ -274,7 +271,7 @@ streamApplicationLogs <- function(authInfo, applicationId, entries, skip) {
 #'   function does not return; instead, log entries are written to the console
 #'   as they are made, until R is interrupted. Defaults to `FALSE`.
 #'
-#' @note These functions only use the \code{libcurl} transport, and only work
+#' @note These functions only work
 #'   for applications deployed to ShinyApps.io.
```
jonkeane, line 275: Very minor but this can all be on one line now, yeah?

R/appMetadata.R
```diff
@@ -8,16 +8,7 @@ appMetadata <- function(
...
-    lifecycle::deprecate_warn(
-      when = "1.0.0",
-      what = "deployApp(quarto = 'can no longer be a path')",
-      with = I("quarto = `TRUE` instead")
-    )
-    quarto <- TRUE
-  } else {
-    check_bool(quarto, allow_na = TRUE)
-  }
+  check_bool(quarto, allow_na = TRUE)
```
jonkeane, line 11: Thanks for this cleanup, FTR https://github.com/rstudio/rsconnect/pull/828 is when this deprecation was added

R/auth.R
```diff
@@ -303,14 +303,8 @@ resendInvitation <- function(
...
 }
 
-#' (Deprecated) List authorized users for an application
-#'
-#' @param appDir Directory containing application. Defaults to current working
-#'  directory.
-#' @export
+# Previously exported, but deprecated since 2015
 authorizedUsers <- function(appDir = getwd()) {
-  .Deprecated("showUsers")
```
jonkeane, line 312: But we still use this internally so we're keeping it around? Is it not possible to shift everything to `showUsers`?

And if yes: did the `.Deprecated("showUsers")` get exposed to users in that internal use? Do we want to maintain that?

R/deployApp.R
```diff
@@ -266,25 +266,6 @@ deployApp <- function(
...
-      appTitle = appTitle,
-      account = account,
-      server = server,
-      upload = upload,
-      recordDir = recordDir,
-      launch.browser = launch.browser,
-      logLevel = logLevel,
-      lint = lint
-    ))
-  }
```
jonkeane, line 287: FTR https://github.com/rstudio/rsconnect/pull/698 introduced this.

What's the error look like when someone does send a document to `appDir`? Not that we need to keep that as easy as possible for folks to get on to the right path, but I want to make sure it's not *very hard* to decipher.

R/http.R
```diff
@@ -436,39 +434,6 @@ httpHeaders <- function() {
...
-  httpType <- getOption("rsconnect.http", "libcurl")
-
-  if (is_string(httpType) && httpType != "libcurl") {
-    lifecycle::deprecate_warn(
-      "1.0.0",
-      I("The `rsconnect.http` option"),
-      details = c(
-        "It should no longer be necessary to set this option",
-        "If the default http handler doesn't work for you, please file an issue at <https://github.com/rstudio/rsconnect/issues>"
-      )
```
jonkeane, line 449: FTR: https://github.com/rstudio/rsconnect/pull/762

R/servers-deprec.R
```diff
@@ -1,63 +0,0 @@
-#' Discover servers automatically
```
jonkeane, line 1: FTR https://github.com/rstudio/rsconnect/pull/750

R/auth.R
nealrichardson: Remarkably, yes, it is used internally in one place, and wrapped in `suppressWarnings()` presumably to squelch the deprecation warning. `showUsers` looked sufficiently different that I didn't want to swap it in, so instead of deleting this, I just made it internal.

jonkeane: > and wrapped in suppressWarnings() presumably to squelch the deprecation warning.

😱 ooph.

ok ok I'm fine with this here, but maybe we should have an issue to clean this up in the future?

nealrichardson: Maybe, though it's not clear that this function is problematic per se, maybe it's useful in its internal use?

PR #1283, opened 2025-12-18
feat: make snowflakeConnectionName optional for some cases

atheriel:
This commit allows `addServer()` and `connectSPCSUser()` to provide a default for the `snowflakeConnectionName` parameter using the existing conventions for a default Snowflake connection. This should simplify programmatic deployment from RStudio, particularly in the Native App, where we can assume a default connection is available.

The default connection is only used when it appears to match the account used for the Connect server URL.

Unit tests are included.

### DIFF
--- a/R/client-connect.R
+++ b/R/client-connect.R
@@ -173,6 +173,59 @@ getSnowflakeAuthToken <- function(url, snowflakeConnectionName) {
   token
 }
 
+# Gets the default Snowflake connection name if (1) it exists; and (2) it seems
+# to match the server URL.
+getDefaultSnowflakeConnectionName <- function(url) {
+  connection <- tryCatch(
+    snowflakeauth::snowflake_connection(),
+    error = function(e) {
+      cli::cli_abort(
+        c(
+          "No default {.arg snowflakeConnectionName}.",
+          i = "Provide {.arg snowflakeConnectionName} explicitly."
+        ),
+        parent = e
+      )
+    }
+  )
+
+  # Validate that the default connection seems to match the account hosting the
+  # Connect server.
+  parsedURL <- parseHttpUrl(url)
+  serverAccount <- extractSnowflakeAccount(parsedURL$host)
+  normalizedAccount <- gsub("_", "-", connection$account, fixed = TRUE)
+  if (!identical(normalizedAccount, serverAccount)) {
+    cli::cli_abort(c(
+      "The default Snowflake connection account {.str {connection$account}} does
+       not appear to match the Connect server.",
+      i = "Pass {.arg snowflakeConnectionName} to use a different connection."
+    ))
+  }
+
+  connectionName <- connection$name
+  if (is.null(connectionName) || !nzchar(connectionName)) {
+    # This should never happen.
+    cli::cli_abort(c(
+      "The Snowflake connection has an empty or missing name field.",
+      i = "Provide {.arg snowflakeConnectionName} explicitly."
+    ))
+  }
+
+  connectionName
+}
+
+# Extract account name from an SPCS hostname.
+extractSnowflakeAccount <- function(hostname) {
+  # For SPCS (including privatelink) URLs, there is some alphanumeric prefix
+  # followed by the hyphenated form of the account, followed by the Snowflake or
+  # Snowflake Computing domain, e.g. "bf2oiajb-testorg-testaccount.snowflakecomputing.app".
+  gsub(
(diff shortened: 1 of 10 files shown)


R/client-connect.R
```diff
@@ -173,6 +173,59 @@ getSnowflakeAuthToken <- function(url, snowflakeConnectionName) {
   token
 }
 
+# Gets the default Snowflake connection name if (1) it exists; and (2) it seems
+# to match the server URL.
+getDefaultSnowflakeConnectionName <- function(url) {
```
aronatkins, line 178: I have successfully tested that a default connection (resolved using `default_connection_name`) is used. The name of that connection is not indicated; should it be included?

```
✔ Registered account for server: SERVER-NAME / username: MY-USER-NAME
```

Here is how I configure my default connection:

```toml
# config.toml
default_connection_name = "my-default-connection-name"
```

In this way, I can use meaningful names for all of my connections and still target a default connection.

Note: When testing, I removed my default connection configuration and received the error:

```
Error in `getDefaultSnowflakeConnectionName()`:
! No default `snowflakeConnectionName`.
ℹ Provide `snowflakeConnectionName` explicitly.
Caused by error in `snowflakeauth::snowflake_connection()`:
! No default connection defined in /Users/aron/.snowflake/connections.toml.
ℹ Define a [default] section in /Users/aron/.snowflake/connections.toml, pass another connection by `name`, or pass connection
  parameters to `snowflake_connection()` directly.
```

This error does not suggest `default_connection_name`; should it? This suggests an improvement to snowflakeauth, not rsconnect.

R/client-connect.R
```diff
@@ -173,6 +173,59 @@ getSnowflakeAuthToken <- function(url, snowflakeConnectionName) {
...
+  connectionName
+}
+
+# Extract account name from an SPCS hostname.
+extractSnowflakeAccount <- function(hostname) {
+  # For SPCS (including privatelink) URLs, there is some alphanumeric prefix
+  # followed by the hyphenated form of the account, followed by the Snowflake or
+  # Snowflake Computing domain, e.g. "bf2oiajb-testorg-testaccount.snowflakecomputing.app".
+  gsub(
+    "([^-]+)-([^\\.]+)(|\\.privatelink)\\.(snowflakecomputing|snowflake)\\.app$",
```
aronatkins, line 223: Note: We see org-account.snowflakecomputing.com hosts, but service-org-account.snowflakecomputing.app for services.

R/servers.R
```diff
@@ -229,6 +230,11 @@ addServer <- function(
   check_name(name, allow_null = TRUE)
 
   if (validate) {
+    # Check for SPCS-like URLs.
+    if (grepl(pattern = "(snowflakecomputing|snowflake)\\.app", x = url)) {
```
aronatkins: Someday, custom domains. Someday.

tests/testthat/_snaps/client-connect.md
```diff
@@ -44,3 +44,25 @@
...
+# getDefaultSnowflakeConnectionName errors when no default connection exists
+
+    Code
+      getDefaultSnowflakeConnectionName(
+        "https://prefix-org-account.snowflakecomputing.app/__api__")
+    Condition
+      Error in `getDefaultSnowflakeConnectionName()`:
+      ! No default `snowflakeConnectionName`.
+      i Provide `snowflakeConnectionName` explicitly.
+      Caused by error in `snowflakeauth::snowflake_connection()`:
```
aronatkins, line 66: nit: This captures the snowflakeauth error, which means that our tests break if snowflakeauth ever changes.

R/client-connect.R
atheriel: On whether `snowflakeauth` should suggest this: the exact logic for determining the default connection name is a bit arcane. I think it makes more sense for users to pass another connection name in most cases than try to change their defaults (which might be injected by the environment).

tests/testthat/_snaps/client-connect.md
atheriel: These errors are actually from the upcoming `snowflakeauth` release, not the current one, and yet it doesn't seem to break CI. Do you want me to scrub it anyway?

R/client-connect.R
aronatkins: I'm fine with this position; does that mean that snowflakeauth is removing the `[default]` suggestion? The guidance from Snowflake does change..

tests/testthat/_snaps/client-connect.md
aronatkins: Oh, sorry. I totally missed that the error is canned via `local_mocked_bindings`. Ignore me.

PR #1284, opened 2025-12-29
feat: add httr2 as HTTP client

nealrichardson:
There are some new features I'd like to support in `rsconnect`, including an endpoint that takes a multipart form upload, and that would involve writing more bespoke code on top of `curl` to handle that. So, I thought I'd take a stab at switching the package to use `httr2`, which would make that a lot more straightforward to do, and hopefully would allow us to delete some custom code here and rely on that library for handling HTTP business.

This PR adds an alternative to the `libcurl` HTTP client that uses `httr2`. It is currently feature flagged under `options(rsconnect.httr2)`, default to `TRUE`. My intent would be to release with httr2 as default, and have the ability to fall back to libcurl if there's an issue, but to try to move towards only using httr2 as soon as is reasonable.

I tried to make this minimally invasive in this PR, but I suspect there are some utilities that can be removed in favor of functions in httr2, even when using the libcurl backend (I'm thinking of the URL parsing and query string building, but possibly others too). That can be handled in followups.

I don't know that the separate cookie store is still needed, I expect httr2 (and libcurl, for that matter) handles cookies fine on its own, but there were some tests asserting behavior of the cookie store so I made sure to keep everything passing.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # rsconnect (development version)
 
+* `rsconnect` now uses [`httr2`](https://httr2.r-lib.org/) as its HTTP client.
+  There should be no user-visible changes as a result, but if something does
+  not work as expected, please file an issue, and you can set
+  `options(rsconnect.httr2 = FALSE)` as a temporary workaround. (#1284)
+
 * Resolved a bug where `renv.lock` files that had multiple repositories were not
   being translated faithfully when creating the manifest file. (#1268)
 


R/http.R
```diff
@@ -23,30 +23,44 @@ httpRequest <- function(
...
-    method = method,
-    path = path,
-    headers = headers,
-    timeout = timeout,
-    certificate = certificate
-  )
-
-  while (isRedirect(httpResponse$status)) {
-    service <- redirectService(service, httpResponse$location)
+  if (isTRUE(getOption("rsconnect.httr2", TRUE))) {
```
aronatkins, line 26: Any reason to use a new option name rather than `rsconnect.http`, which existed until https://github.com/rstudio/rsconnect/pull/1282?

```r
options(rsconnect.http = "httr2")
```

An argument for a new option is that `rsconnect.http` was deprecated (and thus removed).

That said, there are a number of `rsconnect.http.*` options which remain. I don't have a strong opinion, here.

R/http.R
```diff
@@ -23,30 +23,44 @@ httpRequest <- function(
...
-      path = service$path,
+      path = path,
       headers = headers,
       timeout = timeout,
       certificate = certificate
     )
+
+    while (isRedirect(httpResponse$status)) {
+      service <- redirectService(service, httpResponse$location)
+      httpResponse <- httpLibCurl(
```
aronatkins, line 53: Could you check to see how httr handles redirects? Here, this is a full resubmission of the original request using the targeted location. This is different than having the client follow the redirect.

R/http-httr2.R
```diff
@@ -0,0 +1,100 @@
...
+  # SSL certificate handling
+  if (!isTRUE(getOption("rsconnect.check.certificate", TRUE))) {
+    req <- httr2::req_options(req, ssl_verifypeer = FALSE)
+  } else if (!is.null(certificate)) {
+    req <- httr2::req_options(req, cainfo = certificate)
+  }
+  # User-specified libcurl options
+  user_opts <- getOption("rsconnect.libcurl.options")
+  if (is.list(user_opts)) {
+    req <- httr2::req_options(req, !!!user_opts)
```
aronatkins, line 41: `!!!` is new to me. almost too cute!!!

R/http-httr2.R
```diff
@@ -0,0 +1,100 @@
...
+
+  # Verbose output
+  if (httpVerbose()) {
+    req <- httr2::req_verbose(req)
+  }
+
+  # Add file body with progress for large files
+  if (!is.null(file)) {
+    file_size <- file.info(file)$size
+    req <- httr2::req_body_file(req, file, type = contentType)
```
aronatkins, line 57: How much do you want to mirror `httpLibCurl`? For example, it is an error if `contentType` is null when `contentFile` is non-null.

R/http-httr2.R
```diff
@@ -0,0 +1,100 @@
...
+
+  # Don't error on HTTP failures - we handle that in handleResponse
+  req <- httr2::req_error(req, is_error = function(resp) FALSE)
+
+  start <- proc.time()
+  resp <- httr2::req_perform(req)
+  httpTrace(method, path, proc.time() - start)
+
+  # Store cookies from response (same as libcurl backend)
+  cookieHeaders <- httr2::resp_headers(resp, "set-cookie")
```
aronatkins, line 70: You're probably right that we can inherit more cookie handing from httr2, but that be done as follow-up.
For reference, I recently needed to adjust cookie parsing (https://github.com/rstudio/rsconnect/pull/1191, https://github.com/rstudio/rsconnect/pull/1213).

R/http.R
```diff
@@ -88,34 +102,49 @@ httpRequestWithBody <- function(
...
+      path = path,
       headers = authed_headers,
+      contentType = contentType,
+      contentFile = file,
       certificate = certificate
     )
-    httpResponse
+    while (isRedirect(httpResponse$status)) {
+      # This is a simplification of the spec, since we should preserve
+      # the method for 307 and 308, but that's unlikely to arise for our apps
```
aronatkins, line 133: I think this came up before, as we were trying to improve redirect handling. Could you look through the Git history of this flow? We may have seen cases where folks (both Posit-internal and elsewhere) wanted to use a proxy to shuffle all traffic from an old location to a new. I don't recall the specifics.

R/http.R
aronatkins, line 26: Should this start as disabled by default, merge, and then enable after additional testing?

nealrichardson: My thinking was that it wasn't exactly the same shape as the old alternate HTTP backends, this is trying to come in at a higher level. But I don't think it really matters what we call the option.

nealrichardson: What additional testing do you have in mind? Or maybe to ask differently, what might break if the unreleased version has the default to TRUE, and is that a problem? I'm all for flipping it to opt-in if we're about to cut a release, but if not, maybe it helps with our confidence that it is safe if it is on by default in the development version?

aronatkins: I was thinking that we'd merge it disabled, you'd flip the Connect tests to enable the option, and maybe get some additional folks to burn it in before a subsequent PR enables it. Mostly, I'm suggesting merging disabled because it's (slightly) simpler for folks to test once it's on `main`.

It's not a big deal either way. I don't have a pending release, and we can disable if we need to release and don't feel it's ready.

R/http.R
nealrichardson: AFAICT httr2 just does what curl does and follows redirects: https://github.com/r-lib/httr2/blob/c7d4a865a6e94a9f482a11044dd066c7fd7ac14d/man/req_perform.Rd#L66

I believe the default is to reset the method to `GET` on 301/302, but you can set [curl opts](https://curl.se/docs/manpage.html#--post301) to not do that. But reading the code below, it looks like `GET` is expected. So I don't think any customization is required. If that's wrong, then we can set additional `req_options()` in the httr2 method.

R/http.R
aronatkins: Related: The RStudio folks have some test automation for publishing workflows, but I think that runs against the CRAN release, not the GitHub version.

R/http.R
nealrichardson: IIUC curl will preserve the method for 307 and 308: https://curl.se/docs/manpage.html#--follow

I would guess that this custom redirect handling is no longer needed using the libcurl backend either, curl will get this right for us, but I don't think it's worth refactoring to clean that up, IMO we should just move towards httr2.

R/http-httr2.R
nealrichardson: I would trust httr2 to handle cookies correctly, and if not, we should upstream those fixes there.

R/http.R
aronatkins: This PR has much of the discussion around response codes and whether or not we should re-post the body.
https://github.com/rstudio/rsconnect/pull/754

R/http-httr2.R
nealrichardson: Looks like httr2 will also error when `contentType = NULL` and file is not (such that we call this function): https://github.com/r-lib/httr2/blob/c7d4a865a6e94a9f482a11044dd066c7fd7ac14d/R/req-body.R#L81

R/http-httr2.R
nealrichardson: yes, it is clever: https://rlang.r-lib.org/reference/splice-operator.html

R/http.R
aronatkins: I'm OK removing custom redirect handling if it's safe to do so. The linked PR (and likely others) discuss some of our prior struggles with using the built-in handling.

R/http.R
nealrichardson: Yeah that's what I was thinking, I'm happy to submit a PR to flip it off if it causes problems, but merging it TRUE might help us discover more quickly what's wrong (I'm sure there's something!).

I can kick off a CI run of Connect using rsconnect from this branch, see how that goes.

R/http.R
nealrichardson: I will also add that there is a test that redirects are followed, and I had to update the function that munges the `httr2_response` object so that it pulled the final URL, not the original request, in order for the test to pass: https://github.com/rstudio/rsconnect/pull/1284/files#diff-43470aadefcc62364659d4756131b8581f601da141a9d4c9c85f8916a024f2f7R79

So that confirms that redirects are being followed by httr2 (as documented). The question I would have is about POST redirects, which IIUC you have to opt into with curl, but reading the custom redirect handling here, it is converting to `GET` (and does not resubmit the body) on redirect, so the default curl/httr2 behavior matches that.

R/http.R
nealrichardson: It passed btw (I can send you the link to the run if you're curious)

R/http-httr2.R
```diff
@@ -0,0 +1,100 @@
...
+  # No pipe, for compatibility with older R versions
+  req <- httr2::request(url)
+  req <- httr2::req_method(req, method)
+  req <- httr2::req_headers(req, !!!headers)
+  req <- httr2::req_user_agent(req, userAgent())
+
+  # Apply more options
+
+  # SSL certificate handling
+  if (!isTRUE(getOption("rsconnect.check.certificate", TRUE))) {
```
hadley:
```suggestion
  if (isFALSE(getOption("rsconnect.check.certificate", TRUE))) {
```

?

R/http-httr2.R
```diff
@@ -0,0 +1,100 @@
...
+      req <- httr2::req_progress(req)
+    }
+  }
+
+  # Don't error on HTTP failures - we handle that in handleResponse
+  req <- httr2::req_error(req, is_error = function(resp) FALSE)
+
+  start <- proc.time()
+  resp <- httr2::req_perform(req)
+  httpTrace(method, path, proc.time() - start)
```
hadley:
```suggestion
  resp <- httr2::req_perform(req)
  httpTrace(method, path, resp_timing(resp)$total)
```

R/http-httr2.R
```diff
@@ -0,0 +1,99 @@
...
+    if (is_interactive() && file_size >= 10 * 1024^2) {
+      req <- httr2::req_progress(req)
+    }
+  }
+
+  # Don't error on HTTP failures - we handle that in handleResponse
+  req <- httr2::req_error(req, is_error = function(resp) FALSE)
+
+  resp <- httr2::req_perform(req)
+  httpTrace(method, path, resp_timing(resp)$total)
```
nealrichardson:
```suggestion
  httpTrace(method, path, httr2::resp_timing(resp)$total)
```

R/http-httr2.R
nealrichardson: #1285

Issue #1286, opened 2026-01-06
refactor: use httr2 utilities to parse and construct URLs and query params


nealrichardson:
The "URL manipulation" and "Helpers" sections of http.R can mostly be offloaded to httr2, it appears, even when using the libcurl backend. That should also allow some refactoring and simplification of some requests, like those that manually construct query strings.

PR #1287, opened 2026-01-14
feat: Add identity federation support

atheriel:
This commit teaches rsconnect to automatically exchange a Posit Workbench user's identity token (if there is one) for an ephemeral Connect API key using the server's token exchange endpoint.

This enables Workbench users to authenticate without the need to configure API keys or go through the token auth flow, provided that identity federation has been configured on the Connect server side.

All of this is opportunistic, falling back to existing paths if there is no token or if the Connect server won't accept it.

The majority of the complexity in the implementation is due to the fact that we want to be as backward compatible with RStudio as possible, and RStudio very much expects to run the user through the token auth flow.

Finally, note that we're using the brand new `rstudioapi::getIdentityToken()` here, so this needs a version bump for the `rstudioapi` dependency.

Unit tests are included.

Part of https://github.com/posit-dev/connect/issues/28149.

### DIFF
--- a/R/accounts.R
+++ b/R/accounts.R
@@ -409,11 +409,38 @@ getAuthTokenAndUser <- function(server, launch.browser = TRUE) {
 
 # Used by the IDE
 getAuthToken <- function(server, userId = 0) {
+  account <- list(server = server)
+  client <- clientForAccount(account)
+
+  # Check if we already have an API key for this server, in which case we can
+  # bypass the token auth flow.
+  serverUrl <- serverInfo(server)$url
+  cachedApiKey <- getCachedApiKey(serverUrl)
+  if (!is.null(cachedApiKey)) {
+    # Verify that the API key actually works.
+    user <- tryCatch(client$currentUser(), error = function(e) NULL)
+    if (!is.null(user)) {
+      # Return the API key as the "token" with a zero-length private key.
+      # waitForAuthedUser will use this to detect federated authentication later
+      # on.
+      #
+      # This in fairly awkward in-band signalling, but reflects the fact that we
+      # can't change what RStudio expects to happen here.
+      return(list(
+        token = cachedApiKey,
+        private_key = secret(""),
+        # Open a special "you're already authenticated" page as the "claim URL".
+        claim_url = sub("/__api__$", "/connect/#/auth-success", serverUrl),
+        # Signal to future RStudio versions that auth is already complete and
+        # there is no need to open a browser window.
+        authenticated = TRUE
+      ))
+    }
+  }
+
   token <- generateToken()
 
   # Send public key to server, and generate URL where the token can be claimed
-  account <- list(server = server)
-  client <- clientForAccount(account)
   response <- client$addToken(list(
     token = token$token,
     public_key = token$public_key,
@@ -449,6 +476,12 @@ waitForAuthedUser <- function(
   private_key = NULL,
   apiKey = NULL
 ) {
+  # Detect when the "token" is actually an API key by looking for an empty
+  # secret.
+  if (!is.null(token) && !nzchar(private_key)) {
+    return(getAuthedUser(server, apiKey = token))
+  }
+
   # keep trying to authenticate until we're successful; server returns
   # 500 "Token is unclaimed error" (Connect before 2024.05.0)
   # 401 "Unauthorized" occurs before the token has been claimed.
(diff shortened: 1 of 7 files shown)


R/accounts.R
```diff
@@ -409,11 +409,38 @@ getAuthTokenAndUser <- function(server, launch.browser = TRUE) {
...
+      # waitForAuthedUser will use this to detect federated authentication later
+      # on.
+      #
+      # This in fairly awkward in-band signalling, but reflects the fact that we
+      # can't change what RStudio expects to happen here.
+      return(list(
+        token = cachedApiKey,
+        private_key = secret(""),
+        # Open a special "you're already authenticated" page as the "claim URL".
+        claim_url = sub("/__api__$", "/connect/#/auth-success", serverUrl),
```
nealrichardson, line 433: Just confirming we know that `serverUrl` does not have a trailing slash?

nealrichardson: Also I thought the convention was that the success page was served by localhost, no?

R/accounts.R
```diff
@@ -608,6 +641,18 @@ findAccountInfo <- function(
     info$private_key <- gsub("[[:space:]]", "", info$private_key)
   }
 
+  # For standard Connect servers where there are no persisted credentials, try
+  # identity federation.
```
nealrichardson: I'd note a minimum Connect version here in the comments, just for reference.

R/client-identityFederation.R
```diff
@@ -0,0 +1,74 @@
+# Attempt exchange an identity token sourced from Posit Workbench for an
+# ephemeral Connect API key. Returns NULL if this exchange fails or an API key
+# otherwise.
+attemptIdentityFederation <- function(serverUrl) {
```
nealrichardson, line 4: I can trace through the R code to see where the argument comes from, but where exactly does `serverUrl` originate? Is it pulling from a workbench setting in this case? Or does the user have to know and specify the Connect server URL?

R/client-identityFederation.R
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  token <- tryCatch(rstudioapi::getIdentityToken(), error = function(e) NULL)
+  if (is.null(token)) {
+    return(NULL)
+  }
+
+  # Call Connect's exchange endpoint.
+  service <- parseHttpUrl(serverUrl)
+  body <- paste0(
```
nealrichardson, line 25: Since httr2 is now a dependency, you should be able to use [`url_query_build`](https://httr2.r-lib.org/reference/url_query_parse.html) rather than paste, if you'd prefer (see also #1286 for refactoring to do this everywhere)

R/accounts.R
atheriel: I'm pretty sure that that URL is normalized beforehand, yes.

And as for a locally-served page: we could do that, but it's dramatically more complicated and would require additional dependencies. I added this static success page to the Connect server this week instead.

R/accounts.R
atheriel: It hasn't been released yet. Should I add it anyway?

R/client-identityFederation.R
atheriel: They have to know it, though it does get validated in various places so internally at this point we can be sure it's an actual Connect URL.

R/accounts.R
nealrichardson: Yeah I'd say 2026.01.0, for our future selves

R/client-identityFederation.R
atheriel: I looked into this and it doesn't actually do exactly what we want, so I left it as-is.

R/accounts.R
```diff
@@ -608,6 +641,18 @@ findAccountInfo <- function(
     info$private_key <- gsub("[[:space:]]", "", info$private_key)
   }
 
+  # For standard Connect servers where there are no persisted credentials, try
+  # identity federation (added in v2026.01.0).
```
zackverham, line 645: Would there be any value here in trying to determine the version of the connect server before attempting identity federation? I guess the end result is the same (no successful auth) - and maybe its more complicated to try to determine the server version vs just relying on sensible error handling.

atheriel: I kind of felt like that would complicate things, especially because we also don't know if identity federation is actually enabled on this server or whether it has an integration configured for Workbench. Not a lot to be gained by being clever about just the version.

R/accounts.R
```diff
@@ -449,6 +476,12 @@ waitForAuthedUser <- function(
   private_key = NULL,
   apiKey = NULL
 ) {
+  # Detect when the "token" is actually an API key by looking for an empty
```
zackverham, line 479: I know you called out why this is required above - but it remains unfortunate :(

PR #1288, opened 2026-01-20
fix: Make "token" authentication work with SPCS

atheriel:
This commit ensures we attach the signature headers to requests against Connect servers hosted on SPCS. This has the notable effect of enabling push-button publishing from desktop RStudio (which uses token authentication) to these servers.

The required changes here are much simpler than expected on the heels of the recent change to support a default Snowflake connection name.

I've verified that this unlocks adding a Connect server hosted on Snowflake via the desktop UI.

### DIFF
--- a/R/http.R
+++ b/R/http.R
@@ -543,6 +543,12 @@ authHeaders <- function(authInfo, method, path, file = NULL) {
     if (!is.null(authInfo$apiKey)) {
       headers$`X-RSC-Authorization` <- paste("Key", authInfo$apiKey)
     }
+    # When using the "token" flow rather than an API key, we also need the
+    # signature headers.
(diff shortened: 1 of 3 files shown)


R/http.R
```diff
@@ -543,6 +543,12 @@ authHeaders <- function(authInfo, method, path, file = NULL) {
     if (!is.null(authInfo$apiKey)) {
       headers$`X-RSC-Authorization` <- paste("Key", authInfo$apiKey)
     }
+    # When using the "token" flow rather than an API key, we also need the
```
aronatkins, line 546: Could you update `test-spcs.R` to include tests for the "token" flow? We have something similar for API key.

aronatkins (approved):
Feel free to merge once tests are added.

R/http.R
atheriel: Done!

PR #1290, opened 2026-01-20
fix: Allow browser authentication to Snowflake from the Deploy pane

### LINKED ISSUE #1289: Interactive Snowflake authentication doesn't work when deploying from RStudio
### Description

Now that we've fixed some of the Snowflake-related issues for RStudio, one major one remains: the Deploy pane doesn't permit "interactive" flows like those we use for `externalbrowser` auth:

Image

### Your environment

- rsconnect dev

atheriel:
The `snowflakeauth` package only permits external browser authentication from interactive contexts, which makes sense in most cases.

However, push-button deployments from the RStudio IDE actually run in a separate Deploy pane, not from the interactive console -- which isn't considered "interactive" by default.

So this commit tells `rlang` to temporarily treat this as an interactive context, which allows the browser flow to work.

This actually requires a change to the `snowflakeauth` package as well: https://github.com/posit-dev/snowflakeauth/commit/dbd4830f1db8cf4c73b59608d6289a60a280142f. We'll have to get another release of that package out to enable this fix to work properly.

Fixes #1289.

### DIFF
--- a/R/client-connect.R
+++ b/R/client-connect.R
@@ -165,6 +165,13 @@ getSnowflakeAuthToken <- function(url, snowflakeConnectionName) {
   parsedURL <- parseHttpUrl(url)
   ingressURL <- parsedURL$host
 
+  # Detect when we're running in the Deploy pane of RStudio and enable
+  # "interactive" temporarily so that external browser authentication is
+  # permitted.
+  if (rstudioapi::isBackgroundJob()) {
+    rlang::local_options(rlang_interactive = TRUE)
+  }
+
   token <- snowflakeauth::snowflake_credentials(
     snowflakeauth::snowflake_connection(snowflakeConnectionName),
     spcs_endpoint = ingressURL
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,7 +23,8 @@
   Snowflake.
 
 * Push-button publishing from desktop RStudio is now compatible with Connect
-  servers hosted on Snowflake.
+  servers hosted on Snowflake. This includes support for browser-based
+  authentication during deployment. (#1289)
 
 * Added support for using identity federation to authenticate against Connect
   when running in Posit Workbench, when available. This allows deploying to


R/client-connect.R
```diff
@@ -165,6 +165,15 @@ getSnowflakeAuthToken <- function(url, snowflakeConnectionName) {
   parsedURL <- parseHttpUrl(url)
   ingressURL <- parsedURL$host
 
+  # Detect when we're running in the Deploy pane of RStudio and enable
+  # "interactive" temporarily so that external browser authentication is
+  # permitted.
+  if (rstudioapi::isBackgroundJob()) {
+    prev <- getOption("rlang_interactive")
+    options("rlang_interactive" = TRUE)
```
aronatkins: Could this use `rlang::local_options()`? https://rlang.r-lib.org/reference/local_options.html

R/client-connect.R
```diff
@@ -165,6 +165,15 @@ getSnowflakeAuthToken <- function(url, snowflakeConnectionName) {
   parsedURL <- parseHttpUrl(url)
   ingressURL <- parsedURL$host
 
+  # Detect when we're running in the Deploy pane of RStudio and enable
+  # "interactive" temporarily so that external browser authentication is
+  # permitted.
+  if (rstudioapi::isBackgroundJob()) {
```
aronatkins, line 171: There is a mix of calls to `base::interactive()` and `rlang::is_interactive()`. Should they all shift to `rlang::is_interactive()`? Alternatively, should we have our own internal `is_interactive()` implementation which consults `rstudioapi::isBackgroundJob()`?

aronatkins (approved):

R/client-connect.R
atheriel: I feel like this targeted fix is what I'm after for now. Changing how we think about `interactive()` everywhere might have other side effects.

PR #1293, opened 2026-02-12
Disable shinyapps.io log streaming

### LINKED ISSUE #1292: Disable shinyapps.io log streaming
The shinyapps.io backend does not truly support log streaming and has not for many years.
We should update the code and documentation to accurately reflect this; it may have semi-worked but only via continual polling of the same last log line. This will also address backend load issues.

mbaynton:
The backend has not properly supported this for a number of years. The removed code in this change was constantly polling the api instead, which generated unreasonable load.

Connect Cloud has a user logs endpoint with a near realtime streaming capability, we may reintroduce it there in future.

Fixes #1292

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,6 +29,12 @@
   when running in Posit Workbench, when available. This allows deploying to
   Connect servers without the need to store long-lived credentials.
 
+* Removed support for log streaming from shinyapps.io due to loss of support
+  for this feature on the shinyapps.io platform (`showLogs(streaming = TRUE)`).
+  If this feature is important to your workflow, please file an issue and we
+  will consider reintroduction of log streaming via rsconnect in Connect Cloud.
+  (#1292)
+
 # rsconnect 1.7.0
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -29,6 +29,11 @@
   when running in Posit Workbench, when available. This allows deploying to
   Connect servers without the need to store long-lived credentials.
 
+* Removed support for log streaming from shinyapps.io due to loss of support
+  for this feature on the shinyapps.io platform (showLogs(streaming = TRUE)).
```
```suggestion
  for this feature on the shinyapps.io platform (`showLogs(streaming = TRUE)`).
```

NEWS.md
```diff
@@ -29,6 +29,11 @@
...
+  If this feature is important to your workflow, please file an issue and we
+  will consider reintroduction of log streaming via rsconnect in Connect Cloud.
```
karawoo, line 35:
```suggestion
  will consider reintroduction of log streaming via rsconnect in Connect Cloud.
  (#1292)
```

PR #1296, opened 2026-03-04
Support renv profiles

### LINKED ISSUE #1122: Deployment fails with lock file in custom location
I am running a `shiny` app in an R project let's call it the parent app. The project has a `renv.lock` file but it is not located in the project root. The environment variable `RENV_PATHS_LOCKFILE` is set in the `.Rprofile` of the project, before `renv/activate.R` is run. So far, so good.

The parent app generates code for another `shiny` app, let's call this one the daughter app. The daughter app file is placed in a separate directory in within the project. From there the daughter app is deployed.

This process fails because the lock file is not found. It does work when the lock file is copied from the location specified by the env var to the project root. Note that there is no attempt at creating a lock file - it is not found where expected and an error is raised.

Can this be fixed so that custom location of the lock file is taken into account?

jonkeane:
Support renv profiles and custom renv file locations specified with the `RENV_PATHS_LOCKFILE` environment variable

Fixes #1122

### DIFF
--- a/R/bundlePackageRenv.R
+++ b/R/bundlePackageRenv.R
@@ -174,10 +174,6 @@ biocRepos <- function(bundleDir) {
   repos[setdiff(names(repos), "CRAN")]
 }
 
-renvLockFile <- function(bundleDir) {
-  file.path(bundleDir, "renv.lock")
-}
-
 # Find the renv lockfile, checking both the renv-resolved path and the
 # standard location. Returns the path if found, NULL otherwise.
 resolveRenvLockFile <- function(bundleDir) {
@@ -186,7 +182,7 @@ resolveRenvLockFile <- function(bundleDir) {
     return(resolved)
   }
 
-  standard <- renvLockFile(bundleDir)
+  standard <- file.path(bundleDir, "renv.lock")
   if (file.exists(standard)) {
     return(standard)
   }
@@ -196,7 +192,7 @@ resolveRenvLockFile <- function(bundleDir) {
 
 removeRenv <- function(path, lockfile = TRUE) {
   if (lockfile) {
-    unlink(renvLockFile(path))
+    unlink(resolveRenvLockFile(path))
   }
   unlink(file.path(path, "renv"), recursive = TRUE)
 }


R/bundlePackageRenv.R
```diff
@@ -172,6 +176,41 @@ renvLockFile <- function(bundleDir) {
...
+# Ensure the renv lockfile is at the standard bundle location.
+# If found at a custom location (via RENV_PATHS_LOCKFILE or renv profiles),
+# copies it into bundleDir/renv.lock, overwriting any stale lockfile.
+# Returns TRUE if the lockfile is available, FALSE otherwise.
+ensureRenvLockFile <- function(bundleDir) {
+  standard <- renvLockFile(bundleDir)
+  resolved <- renv::paths$lockfile(project = bundleDir)
+
+  # If the resolved file exists, and is different from the standard location,
+  # copy it to the standard location.
```
nealrichardson: Why do we need to copy it at all? Why can't we use it where it lives? It's a little odd to me that this function would alter the contents of the current working directory.

R/bundlePackageRenv.R
```diff
@@ -172,6 +176,41 @@ renvLockFile <- function(bundleDir) {
...
+  if (file.exists(resolved)) {
+    # Normalize to avoid self-copy (e.g., /var vs /private/var on macOS)
+    if (
+      normalizePath(resolved, mustWork = FALSE) !=
+        normalizePath(standard, mustWork = FALSE)
+    ) {
+      if (file.exists(standard)) {
+        cli::cli_warn(c(
+          "Using lockfile at {.path {resolved}} instead of {.path {standard}}.",
+          i = "The lockfile in the project root may be outdated.",
```
nealrichardson: If I run this function twice, I'm going to get this warning, even though the files themselves are identical. Should we be checking the contents of the files instead?

R/bundlePackageRenv.R
jonkeane: One thing that's confusing about this is that it's actually not copying it into the working directory, but actually into the temporary directory the rsconnect creates and copies files over to a temporary directory

https://github.com/rstudio/rsconnect/blob/c70493180cdf30b41a29bcb59ac53fa5b45b956e/R/appDependencies.R#L130-L135 which calls https://github.com/rstudio/rsconnect/blob/c70493180cdf30b41a29bcb59ac53fa5b45b956e/R/bundle.R#L1-L12

But thinking about this more, we might need to make sure that we can handle relative paths in `RENV_PATHS_LOCKFILE` since when this is run inside of the temp bundle dir, those might not work (e.g. if `RENV_PATHS_LOCKFILE="../renv.lock"` and the app is at `/home/jonkeane/Documents/myapp/subdir/`, my lock file is at `/home/jonkeane/Documents/myapp/renv.lock` we don't want the target of this copy to be `/tmp/.../renv.lock`

R/bundlePackageRenv.R
jonkeane: I can add checking the contents, but running user-exported commands in rsconnect, this path should not alter someone's app directory (so this shouldn't happen in practice unless someone calls `rsconnect:::ensureRenvLockFile` (or the other internal functions that always operate on the temp bundle dir

R/bundlePackageRenv.R
```diff
@@ -30,13 +30,16 @@ snapshotRenvDependencies <- function(
     progress = FALSE
   )
   renv::snapshot(bundleDir, packages = deps$Package, prompt = FALSE)
+  # renv::snapshot() respects RENV_PATHS_LOCKFILE and renv profiles, so the
+  # lockfile may have been written to a non-standard location.
+  lockfile <- resolveRenvLockFile(bundleDir)
```
karawoo, line 35: might be good to have some error handling here in case something goes wrong with resolving the lockfile, otherwise we're likely to get cryptic errors from `parseRenvDependencies`
```suggestion
  lockfile <- resolveRenvLockFile(bundleDir)
  if (is.null(lockfile)) {
    cli::cli_abort("renv::snapshot() did not produce a lockfile")
  }
```

karawoo (commented):
I think we also need to make `removeRenv` aware of other potential lockfile locations

https://github.com/rstudio/rsconnect/blob/591e6527751f08f6d75ab0109ef4ca1406341b5f/R/bundlePackageRenv.R#L177

PR #1301, opened 2026-03-13
fix renv profiles

### LINKED ISSUE #1300: Deploying with renv profiles uses wrong lockfile
### Description

When deploying a project that uses an renv profile, `deployApp()` fails with "Library and lockfile are out of sync" even though `renv::restore()` confirms the library is synchronized. The issue is that `bundleAppDir()` copies project files into a temp staging directory, and the renv/ directory is excluded from bundling (as expected). But when using an renv profile like 'dev', the profile-specific lockfile lives at renv/profiles/dev/renv.lock, so it gets excluded too. Only the root renv.lock (from the default profile) ends up in the bundle.

Later, `parseRenvDependencies()` reads this default-profile lockfile and compares versions against the library, which has packages from the active profile. The versions don't match, producing the error.

### Your environment

_No response_

karawoo:
fixes #1300 by ensuring that we copy the profile renv.lock file into the temp directory when creating the bundle

### DIFF
--- a/R/bundle.R
+++ b/R/bundle.R
@@ -45,6 +45,21 @@ bundleAppDir <- function(
       tweakRProfile(to)
     }
   }
+
+  # When an renv profile is active, the profile-specific lockfile lives under
+  # renv/profiles/<name>/renv.lock, which is excluded from the bundle (the
+  # entire renv/ directory is ignored). Resolve the correct lockfile from the
+  # original appDir and copy it into the bundle so that
+  # parseRenvDependencies() compares against the right lockfile.
+  profileLockfile <- tryCatch(
+    resolveRenvLockFile(appDir),
+    error = function(e) NULL
+  )
+  if (!is.null(profileLockfile) && file.exists(profileLockfile)) {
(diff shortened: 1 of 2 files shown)


R/bundle.R
```diff
@@ -45,6 +45,21 @@ bundleAppDir <- function(
...
+  # entire renv/ directory is ignored). Resolve the correct lockfile from the
+  # original appDir and copy it into the bundle so that
+  # parseRenvDependencies() compares against the right lockfile.
+  profileLockfile <- tryCatch(
+    resolveRenvLockFile(appDir),
+    error = function(e) NULL
+  )
+  if (!is.null(profileLockfile) && file.exists(profileLockfile)) {
+    bundleLockfile <- file.path(bundleDir, "renv.lock")
+    file.copy(profileLockfile, bundleLockfile, overwrite = TRUE)
```
aronatkins, line 60: Ah. This replaces the primary `renv.lock` with the named profile lock file.

karawoo: yes, just in the temp directory, not the project source

PR #1308, opened 2026-04-08
remove 2023.03 and 2022.10 from integration testing matrix

### LINKED ISSUE #1307: Connect integration tests fail
Integration tests fail, as observed in https://github.com/rstudio/rsconnect/pull/1306.

- 2023.03.0
- 2022.10.0

Both of these versions are outside our support program, but were in the initial version of this workflow, added just a few months ago (https://github.com/rstudio/rsconnect/pull/1274).

Both versions fail because they lack `cmake`; we may be hitting that because Package Manager no longer hosts binaries for the target distribution (bionic).

karawoo:
closes #1307

### DIFF
--- a/.github/workflows/connect-integration.yaml
+++ b/.github/workflows/connect-integration.yaml
@@ -22,9 +22,6 @@ jobs:
           # These versions use R 4.2 and seem to take forever to install R dependencies
           # - "2024.03.0" # jammy
           # - "2023.09.0" # jammy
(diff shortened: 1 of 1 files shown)


.github/workflows/connect-integration.yaml
aronatkins, line 21: Confirmed that 2024.09.0 is just past our supported version window, so we still span what is supported.
https://docs.posit.co/supported-versions/#posit-connect

aronatkins (approved):

PR #1316, opened 2026-04-28
fix bioconductor repository resolution

### LINKED ISSUE #1314: manifest chooses wrong repository for bioconductor packages in transit
### Description

**Summary**

When using `rsconnect` to create a manifest file for a project that includes Bioconductor packages that happen to be available in CRAN's transit directory, the manifest will get the wrong value for "Repository"

**Repro**
Create a shiny app that uses `S4Vectors` from Bioconductor.

```r
# Session state: R 4.5.3, Posit Cloud, BiocManager 3.22 repos configured.
# S4Vectors 0.48.1 installed from Bioconductor.

options(repos = c(CRAN = "https://cran.rstudio.com/"))
install.packages(c("BiocManager", "rsconnect"))
BiocManager::install(c("S4Vectors"))
packageDescription("S4Vectors")$Version
#> [1] "0.48.1"

# See what rsconnect finds
ap <- rsconnect:::availablePackages(getOption("repos"))
ap["S4Vectors", c("Version", "Repository")]
#> Version                                          Repository
#> "0.49.1-1"    "https://cran.rstudio.com/src/contrib/Transit"

rsconnect:::findRepoUrl("S4Vectors", ap)
#> [1] "https://cran.rstudio.com"
```

Resulting `manifest.json` entry:

```json
"S4Vectors": {
  "Source": "Bioconductor",
  "Repository": "https://cran.rstudio.com",
  "description": { "Version": "0.48.1", "Repository": "Bioconductor 3.22", "...": "..." }
}
```

`https://cran.rstudio.com` doesn't serve S4Vectors 0.48.1. Only the Transit copy at 0.49.1-1 is there, under a different URL. Any deploy target that takes this manifest literally will fail the install.

**Root cause**

Peeking into the Packages on each repository, first you can see Bioconductor has version 0.48.1
Value from https://bioconductor.org/packages/release/bioc/src/contrib/PACKAGES
```
Package: S4Vectors
Version: 0.48.1
Depends: R (>= 4.0.0), methods, utils, stats, stats4, BiocGenerics (>=
        0.53.2)
Suggests: IRanges, GenomicRanges, SummarizedExperiment, Matrix,
        DelayedArray, ShortRead, graph, data.table, RUnit, BiocStyle,
        knitr
License: Artistic-2.0
MD5sum: 97d42c8c8d6bf44fd95675b48dc66d1e
NeedsCompilation: yes
```

Next you can see CRAN has version 0.49.1-1 in Transit, pinned to `(R (>=4.6.0)`
Value from https://cran.rstudio.com/src/contrib/PACKAGES
```
Package: S4Vectors
Version: 0.49.1-1
Depends: R (>= 4.6.0), methods, utils, stats, stats4, BiocGenerics (>=
        0.53.2)
Suggests: IRanges, GenomicRanges, SummarizedExperiment, Matrix,
        DelayedArray, ShortRead, graph, data.table, RUnit, BiocStyle,
        knitr
License: Artistic-2.0
MD5sum: 1f9b83fd10000c017015e8e9f1af4860
NeedsCompilation: yes
Path: Transit
```

Two pieces interact to produce the bad output.

1. `rsconnect:::availablePackages()` in `R/bundlePackage.R` passes a narrower `filters` argument than R's default:

    ```r
    availablePackages <- function(repos) {
      # note that we explicitly exclude the "R_version" filter as we want to ensure
      # that packages which require newer versions of R than the one currently
      # in use can still be marked as available on CRAN -- for example, currently
      # the package "foreign" requires "R (>= 4.0.0)" but older versions of R
      # can still successfully install older versions from the CRAN archive
      filters <- c(
        getOption("rsconnect.available_packages_filters", default = c()),
        "duplicates"
      )
      available.packages(repos = repos, type = "source", filters = filters)
    }
    ```

    R's default is `c("R_version", "OS_type", "subarch", "duplicates")`. rsconnect's override is just `"duplicates"`, which means rows that would otherwise be dropped by `R_version` (packages declaring an R dependency newer than the current session) stay in the matrix.

    The side effect of that choice is that CRAN Transit builds also stay visible even though the depend on a higher version of R than used in this session:

    ```r
    options(repos = BiocManager::repositories())
    ap_default <- available.packages(type = "source")                       # default filters
    ap_rsconnect <- available.packages(type = "source", filters = "duplicates") # rsconnect's

    ap_default[ap_default[,"Package"]=="S4Vectors", "Repository"]
    #> "https://bioconductor.org/packages/3.22/bioc/src/contrib"

    ap_rsconnect[ap_rsconnect[,"Package"]=="S4Vectors", "Repository"]
    #> "https://cloud.r-project.org/src/contrib/Transit"
    ```

    The Transit row only becomes visible once the `R_version` filter is dropped.

2. `rsconnect:::findRepoUrl` then picks the first `Package`-name hit in that matrix, with no version awareness:

    ```r
    function(pkg, availablePackages) {
      idx <- match(pkg, availablePackages[, "Package"])
      if (!is.na(idx)) {
        repo <- availablePackages[[idx, "Repository"]]
        gsub("/src/contrib.*$", "", repo)
      } else {
        NA_character_
      }
    }
    ```

    With `"duplicates"` already having collapsed multiple rows for the same package to the highest-version row, `findRepoUrl` returns whichever Repository URL that row happened to carry. The combination of "keep extra rows visible" plus "pick by name only" is leads to a CRAN Transit URL in a Bioc manifest entry.

    Note I'm not sure this requires the `R_version` filter to be dropped. Transit rows whose R requirement happens to match the current session (e.g. Rcpp, S7, precondition at the moment) collide the same way even under default filters. But dropping `R_version` expands the set of packages that can collide from "only those whose Transit build matches the session's R" to "any package currently in Transit," which is how I hit the Bioconductor case.

**Proposed fix**

Two options, not mutually exclusive.

**Bioconductor fix: swap the lookup order in `standardizeRenvPackage`'s Bioconductor branch.**

The function already receives a separate `biocPackages` matrix alongside the combined `availablePackages`, but only consults it if the packages isn't otherwise found:

```r
else if (pkg$Source == "Bioconductor") {
    pkg$Repository <- findRepoUrl(pkg$Package, availablePackages)
    if (is.na(pkg$Repository)) {
        pkg$Repository <- findRepoUrl(pkg$Package, biocPackages)
    }
}
```

Consulting `biocPackages` first would resolve the specific collision in this report:

```r
else if (pkg$Source == "Bioconductor") {
    pkg$Repository <- findRepoUrl(pkg$Package, biocPackages)
    if (is.na(pkg$Repository)) {
        pkg$Repository <- findRepoUrl(pkg$Package, availablePackages)
    }
}
```

**Broader fix?: make the lookup version-aware.**

The Bioconductor swap only fixes one branch. The same `findRepoUrl(pkg$Package, ...)` call appears in four other places (`R/bundlePackageRenv.R:118`, `:137`, `:159`, and the packrat path at `R/bundlePackagePackrat.R:91`, `:95`), and they all have this same theoretical issue: they look up a package by name and trust whatever Repository URL comes back, even if it points to a different version than what's installed. Transit is the obvious trigger, but the same mismatch can happen any time a package appears in more than one configured repo at different versions.

The fix would be to match on package *and* version, not just name. Two things would need to change together: `availablePackages()` would need to stop dropping rows via the `"duplicates"` filter (since those are the rows you'd be disambiguating against), and `findRepoUrl()` would need to take a version argument. Callers already have `pkg$Version` on hand, so passing it through is the easy part. The harder questions are what to do when no row matches the installed version (the CRAN-archive case the existing comment is guarding against), and whether keeping the wider matrix breaks anyone using the `rsconnect.available_packages_filters` option.

**Impact**

Affects any `rsconnect` user whose local CRAN currently lists a Transit row for a package they have installed from a different source. 

**Environment**

- rsconnect: `<your version>` (run `packageVersion("rsconnect")`)
- R 4.5.3, Posit Cloud, Ubuntu Noble
- BiocManager 3.22

### Your environment

_No response_

karawoo:
Fixes #1314 

Given the pretty narrow scope of the issue I went with the simpler fix. I'm not against doing version-aware lookups in the future, but this should fix the immediate issue.

Making this fixed caused a test to fail due to lazy evaluation. The second commit fixes this so we check up front whether bioc repositories are in the lockfile and skip the lookup when they are.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # rsconnect (development version)
 
+* Fixed an issue where Bioconductor packages could be incorrectly associated
+  with a CRAN repository URL when the same package appeared in CRAN's Transit
+  directory. (#1314)
+
 * `deployApp()` with `logLevel = "verbose"` no longer errors using the `httr2` backend. (#1312)
 
 # rsconnect 1.8.0
--- a/R/bundlePackageRenv.R
+++ b/R/bundlePackageRenv.R
@@ -145,10 +145,9 @@ standardizeRenvPackage <- function(
       }
     }
   } else if (pkg$Source == "Bioconductor") {
-    pkg$Repository <- findRepoUrl(pkg$Package, availablePackages)
+    pkg$Repository <- findRepoUrl(pkg$Package, biocPackages)
     if (is.na(pkg$Repository)) {
-      # Try packages defined from default bioC repos
-      pkg$Repository <- findRepoUrl(pkg$Package, biocPackages)
+      pkg$Repository <- findRepoUrl(pkg$Package, availablePackages)
     }
   } else if (pkg$Source %in% c("Bitbucket", "GitHub", "GitLab")) {
     pkg$Source <- tolower(pkg$Source)
--- a/R/bundlePackageRenv.R
+++ b/R/bundlePackageRenv.R
@@ -47,10 +47,17 @@ parseRenvDependencies <- function(lockfile, bundleDir, snapshot = FALSE) {
     vapply(renvLock$R$Repositories, "[[", "URL", FUN.VALUE = character(1)),
     vapply(renvLock$R$Repositories, "[[", "Name", FUN.VALUE = character(1))
   )
+  bioc <- biocRepos(bundleDir)
+  if (any(bioc %in% repos)) {
+    biocPkgs <- NULL
+  } else {
+    signal("evaluating", class = "rsconnect_biocRepos")
+    biocPkgs <- availablePackages(bioc)
+  }
   deps <- standardizeRenvPackages(
     renvLock$Packages,
     repos,
-    biocPackages = biocPackages(bundleDir)
+    biocPackages = biocPkgs
   )
   if (nrow(deps) == 0) {
     return(data.frame())
@@ -174,10 +181,6 @@ standardizeRenvPackage <- function(
   pkg[manifestPackageColumns(pkg)]
 }
 
-biocPackages <- function(bundleDir) {
-  signal("evaluating", class = "rsconnect_biocRepos") # used for testing
-  availablePackages(biocRepos(bundleDir))
-}
 biocRepos <- function(bundleDir) {
   repos <- getFromNamespace("renv_bioconductor_repos", "renv")(bundleDir)
   repos[setdiff(names(repos), "CRAN")]
--- a/tests/testthat/test-bundlePackageRenv.R
+++ b/tests/testthat/test-bundlePackageRenv.R
@@ -169,6 +169,25 @@ test_that("BioC gets normalized repo", {
   )
 })
 
+test_that("BioC prefers biocPackages over availablePackages (#1314)", {
+  bioc_pkg <- list(Package = "S4Vectors", Source = "Bioconductor")
+
+  avail <- data.frame(
+    Package = "S4Vectors",
+    Repository = "https://cran.rstudio.com/src/contrib/Transit",
+    stringsAsFactors = FALSE
+  )
+
+  bioc <- data.frame(
+    Package = "S4Vectors",
+    Repository = "https://bioconductor.org/packages/3.22/bioc/src/contrib",
+    stringsAsFactors = FALSE
+  )
+
+  result <- standardizeRenvPackage(bioc_pkg, avail, biocPackages = bioc)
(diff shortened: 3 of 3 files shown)


marcosnav (approved):
I agree with using a simple fix for this one for now 👍

Issue #1317, opened 2026-04-28
rsconnect 1.8.0 with workflow fails


novica:
### Description

I am using this workflow: https://solutions.posit.co/operations/deploy-methods/ci-cd/github-actions/ 

Between Feb 4 when it passed and April 28 when it fails (today)  the only change I notice is the update of the rsconnect version from 1.7.0 to 1.8.0. 

Previously it was reporting this:
```
Run rsconnect::writeManifest()
- The project is out-of-sync -- use `renv::status()` for details.
ℹ Capturing R dependencies from renv.lock
✔ Found 95 dependencies
```

Now it fails like this:

```
Run rsconnect::writeManifest()
- The project is out-of-sync -- use `renv::status()` for details.
ℹ Capturing R dependencies from renv.lock
Error in `parseRenvDependencies()`:
! Library and lockfile are out of sync
ℹ Use renv::restore() or renv::snapshot() to synchronise
ℹ Or ignore the lockfile by adding to your .rscignore
```

### Your environment

1.8.0

novica:
Hi again I see this is discussed at #1318 

Let me provide some more context for my issue. It was late yesterday when I was writing.

I notice that I have another step in my workflow, and that is to install `rsconnect`. Maybe in the solution published at https://solutions.posit.co/operations/deploy-methods/ci-cd/github-actions/ it is assumed that `rsconnect` is recorded `renv` already?

Edit: Yes the [example](https://github.com/sol-eng/shiny-app-demo-cicd-github-actions/blob/64a16b0af044de33d0ce10cee41d35bc14129f3d/renv.lock#L863) seems to have `rsconnect` in the `renv.lock`. I am not sure if that is the best way though, rsconnect is just used to deploy, it is not a dependency for the app.

Other than that I have amended the workflow now to what I had in February to include:  `r-version: '4.5.1'` in `setup-r`. No other changes.

The out-of-sync `renv` message is due to few packages being needed (at least in my experience) to deploy a Shiny Golem app. 

```
> renv::status()
The following package(s) are in an inconsistent state:

 package  installed recorded used
 desc     y         y        n   
 pkgbuild y         y        n   
 pkgload  y         y        n   
 processx y         y        n   
```

As a side note, I think it should be possible to just create the manifest file from the lock file. I don't see a reason to have all the environment setup on a deployment action, but maybe I misunderstand something.

karawoo:
[...]

Issue #1315, opened 2026-04-28
Feature Request: Option to bypass local-to-lockfile sync check in deployApp()


daattali:
I'm developing a shiny app contained within a package, and want to deploy the shiny app to shinyapps.io via {rsconnect}. In order for the deployment to be successful, all the dependencies of the package and the package itself need to be listed in the renv.lock file, so it can be installed.

{renv} itself cannot natively allow this to happen (https://github.com/rstudio/renv/issues/2285), so my workaround attempt was to manually add the package to the renv lockfile using `renv::record()`. Then I ran into this error during deployment:

```
ℹ Capturing R dependencies from renv.lock
Error in `parseRenvDependencies()`:
! Library and lockfile are out of sync
ℹ Use renv::restore() or renv::snapshot() to synchronise
ℹ Or ignore the lockfile by adding to your .rscignore
Backtrace:
    ▆
 1. └─rsconnect::deployApp(...)
 2.   └─rsconnect:::bundleApp(...)
 3.     └─rsconnect:::createAppManifest(...)
 4.       └─rsconnect:::bundlePackages(...)
 5.         └─rsconnect:::computePackageDependencies(...)
 6.           └─rsconnect:::parseRenvDependencies(lockfile, bundleDir)
 7.             └─cli::cli_abort(...)
 8.               └─rlang::abort(...)
Execution halted
```

It appears that when deploying an app, if rsconnect notices that I have my own lockfile, rsconnect attempt to do its own dependency check and verify that they are in sync. This check fails. But if deployApp() would use the provided renv lockfile that I have, it should work.

It would be helpful if there was a way to disable the sync check. I'm currently stuck and unsure how to deploy my shiny app that's inside a package.

karawoo:
Hi @daattali, thanks for raising this. Agreed that rsconnect should be more flexible regarding whether the lockfile and project library are in sync or not. We've added the ability to bypass the renv.lock file in #1318, and separately are planning to support deploying from the lockfile only (#1319). If I understand correctly, the latter issue is what you're looking for. I'm going to close this as a duplicate of that issue, but feel free to reopen if you think that doesn't accurately capture your needs.

Issue #1078, opened 2024-05-23
Suddenly unable to deploy shiny app due to renv error: aborting snapshot due to pre-flight validation failure


jzadra:
Hi, I posted this under `renv` issues (see https://github.com/rstudio/renv/issues/1903#issue-2299031623) , but was directed to post it here as it is likely due to how rsconnect is calling `renv` when publishing to shinyapps.io

Kevin Ushey seemed to think it was because of the warning about `pfas.shiny.golem`, which is the shiny app/package I'm developing and publishing.  It seems odd that renv would look for that package.

See https://github.com/rstudio/renv/issues/1903#issuecomment-2127937986 for the references I found in my app to the package name of the app.

=====

Today I am suddenly unable to deploy my golem shiny app to shinyapps.io.  I have made minor changes to the app but nothing I'm aware of that would affect the deployment, and the app works fine locally.

I did upgrade Rstudio to the most recent version this morning.

The error provides nothing to go on for troubleshooting as far as I can see.

I should also note that I don't use renv for this project, so I'm not sure why it's running.

── Preparing for deployment ────────────────────────────────────────────────────
✔ Re-deploying "xxx" using "server: shinyapps.io / username: xxx"
ℹ Looking up application with id "xxx"...
✔ Found application <xxx>
ℹ Bundling 74 files: .here, .Rbuildignore, app.R, CODE_OF_CONDUCT.md, DESCRIPTION, dev/01_start.R, dev/02_dev.R, dev/03_deploy.R, dev/config_attachment.yaml, dev/run_dev.R, man/figures/README-pressure-1.png, man/mod_filter_ui.Rd, man/run_app.Rd, NAMESPACE, NEWS.md, README.md, README.Rmd, scratch/app structure reprex - cicerone.R, …, R/utils_cicerone.R, and R/utils_helpers.R
ℹ Capturing R dependencies with renv
The following package(s) were installed from an unknown source:
- pfas.shiny.golem [0.0.0.9000]
renv may be unable to restore these packages in the future.
Consider reinstalling these packages from a known source (e.g. CRAN).

Traceback (most recent calls last):
9: rsconnect::deployApp(appDir = "~/Documents/R Projects/pfas.shiny.golem", 
       appFileManifest = "/var/folders/mp/w8f9_2c572dfzd9nf4363_lw0000gn/T/02ae-80b4-975c-c128", 
       account = "xxx", server = "shinyapps.io", appName = "xxx", 
[...]

PR #1318, opened 2026-04-28
Add dependencySource parameter to bypass renv.lock during deployment

### LINKED ISSUE #1046: Add posibility of bypassing "Library and lockfile are out of sync" check.
Hey 

With release of rsconnect 1.0.0 - i was really happy to learn that you could use a renv.lock file for specifying deployment dependencies. 

This was a big help since the release script in our azure pipelines, had different dependencies from the actual apps that we were deploying. This helped us sort out a lot of these problems.  However with the functionality that enforce that the Library and lockfile can't be out of sync. We again have this problem. That is the deployment script need to have the same dependencies as specified in the app. 

I understand that the change was made  to avoid having different experiences locally and on the connect server. So therefore i would like to suggest that we might add an argument to deployApp and writeManifest that allows us to bypass the check.

I can also try to create a pull request my self if your are interested? 

I am also interested in other ways to apporach the issue if you have any.

marcosnav:
## Summary

- `deployApp()`, `writeManifest()`, and `appDependencies()` gain a `dependencySource` parameter (default `"strict"`). When set to `"library"`, the `renv.lock` file is ignored and package dependencies are resolved from the local library instead via `renv::snapshot()`.
- A warning and interactive confirmation prompt are shown when `dependencySource = "library"`, making it clear that the deployed content will reflect the local library, not the lockfile.
- Improved error messages when `renv::snapshot()` fails during dependency discovery, with actionable workarounds.

Closes #1046, #1315. Related: #1078, #1317.

## Behavior

When `dependencySource = "library"`:
1. The lockfile branch in `computePackageDependencies()` is skipped entirely.
2. Dependencies fall through to `snapshotRenvDependencies()`, which does a fresh `renv::snapshot()` from the local library.
3. The user sees a warning and (in interactive sessions) a `[Y/n]` confirmation prompt before proceeding.

This avoids the "Library and lockfile are out of sync" error for users who:
- Deploy from CI/CD environments where the local library differs from the lockfile.
- Manually edit the lockfile (e.g. with `renv::record()`).
- Develop Shiny apps packaged as R packages (golem pattern).

### Screenshots
**non-sync library -message "Or set dependencySource = ..."**

Screenshot 2026-05-14 at 11 42 23 a m

**Deploying with `dependencySource = "library"`**

Screenshot 2026-05-14 at 11 43 12 a m

**`writeManifest(dependencySource = "library")` also asks for confirmation**

Screenshot 2026-05-14 at 11 35 38 a m

## Test plan

- [x] Existing `parseRenvDependencies` sync-check test still passes (snapshot updated with new hint)
- [x] New test: `dependencySource = "library"` bypasses lockfile and resolves from local library
- [x] New test: mocked `renv::snapshot` failure produces expected error message (snapshot test)
- [x] New tests for `confirmDependencySourceLibrary()`: user answers Y, empty input (default Y), user answers n (aborts), non-interactive (proceeds with warning)
- [x] Rd files regenerated with `roxygen2::roxygenise()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,6 +19,18 @@
 
 * `deployApp()` with `logLevel = "verbose"` no longer errors using the `httr2` backend. (#1312)
 
+* `deployApp()`, `writeManifest()`, and `appDependencies()` gain a
+  `checkLockfile` parameter. Set `checkLockfile = FALSE` to skip the
+  verification that `renv.lock` versions match the local library. This is
+  useful when deploying from CI/CD environments where the local library
+  differs from the lockfile, or when the lockfile has been manually edited
+  (e.g. with `renv::record()`). (#1046, #1315, #1317)
+
+* Improved error messages when `renv::snapshot()` fails during dependency
+  discovery. The error now suggests actionable workarounds such as
+  `renv::settings$ignored.packages()` for locally-developed packages.
+  (#1078)
+
 # rsconnect 1.8.0
 
 * `rsconnect` now uses [`httr2`](https://httr2.r-lib.org/) as its HTTP client.
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@
 * `deployApp()` with `logLevel = "verbose"` no longer errors using the `httr2` backend. (#1312)
 
 * `deployApp()`, `writeManifest()`, and `appDependencies()` gain a
-  `dependencySource` parameter. Set `dependencySource = "library"` to ignore
+  `dependencyResolution` parameter. Set `dependencyResolution = "library"` to ignore
   the `renv.lock` file and resolve package dependencies by scanning the code.
   The version that is recorded is what is installed in the libraries active in the
   R session (i.e. what is displayed with `.libPaths()`). This is useful when
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,12 +19,11 @@
 
 * `deployApp()` with `logLevel = "verbose"` no longer errors using the `httr2` backend. (#1312)
 
-* `deployApp()`, `writeManifest()`, and `appDependencies()` gain a
-  `checkLockfile` parameter. Set `checkLockfile = FALSE` to skip the
-  verification that `renv.lock` versions match the local library. This is
-  useful when deploying from CI/CD environments where the local library
-  differs from the lockfile, or when the lockfile has been manually edited
-  (e.g. with `renv::record()`). (#1046, #1315, #1317)
+* `deployApp()`, `writeManifest()`, and `appDependencies()` gain an
+  `ignoreLockfile` parameter. Set `ignoreLockfile = TRUE` to ignore the
+  `renv.lock` file and resolve package dependencies from the local library
+  instead. This is useful when deploying from CI/CD environments where the
+  local library doesn't match the lockfile. (#1046, #1315, #1317)
 
 * Improved error messages when `renv::snapshot()` fails during dependency
   discovery. The error now suggests actionable workarounds such as
--- a/NEWS.md
+++ b/NEWS.md
@@ -25,12 +25,10 @@
   The version that is recorded is what is installed in the libraries active in the
   R session (i.e. what is displayed with `.libPaths()`). This is useful when
   deploying from environments where there is a mismatch between the
-  renv lock file (#1046, #1315, #1317)
+  renv lock file and the user's environment. (#1046, #1315, #1317)
 
 * Improved error messages when `renv::snapshot()` fails during dependency
-  discovery. The error now suggests actionable workarounds such as
-  `renv::settings$ignored.packages()` for locally-developed packages.
-  (#1078)
+  discovery. (#1078)
 
 # rsconnect 1.8.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,9 +21,11 @@
 
 * `deployApp()`, `writeManifest()`, and `appDependencies()` gain an
   `ignoreLockfile` parameter. Set `ignoreLockfile = TRUE` to ignore the
-  `renv.lock` file and resolve package dependencies from the local library
-  instead. This is useful when deploying from CI/CD environments where the
-  local library doesn't match the lockfile. (#1046, #1315, #1317)
+  `renv.lock` file and resolve package dependencies by scanning the code. 
+  The version that is recorded is what is installed in the libraries active in the 
+  R session (i.e. what is displayed with `.libPaths()`). This is useful when 
+  deploying from environments where there is a mismatch between the 
+  renv lock file (#1046, #1315, #1317)
 
 * Improved error messages when `renv::snapshot()` fails during dependency
   discovery. The error now suggests actionable workarounds such as
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,12 +19,12 @@
 
 * `deployApp()` with `logLevel = "verbose"` no longer errors using the `httr2` backend. (#1312)
 
-* `deployApp()`, `writeManifest()`, and `appDependencies()` gain an
-  `ignoreLockfile` parameter. Set `ignoreLockfile = TRUE` to ignore the
-  `renv.lock` file and resolve package dependencies by scanning the code. 
-  The version that is recorded is what is installed in the libraries active in the 
-  R session (i.e. what is displayed with `.libPaths()`). This is useful when 
-  deploying from environments where there is a mismatch between the 
+* `deployApp()`, `writeManifest()`, and `appDependencies()` gain a
+  `dependencySource` parameter. Set `dependencySource = "library"` to ignore
+  the `renv.lock` file and resolve package dependencies by scanning the code.
+  The version that is recorded is what is installed in the libraries active in the
+  R session (i.e. what is displayed with `.libPaths()`). This is useful when
+  deploying from environments where there is a mismatch between the
   renv lock file (#1046, #1315, #1317)
 
 * Improved error messages when `renv::snapshot()` fails during dependency
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -136,7 +136,12 @@ appDependencies <- function(
   defer(unlink(bundleDir, recursive = TRUE))
 
   extraPackages <- inferRPackageDependencies(appMetadata)
-  deps <- computePackageDependencies(bundleDir, extraPackages, quiet = TRUE, checkLockfile = checkLockfile)
+  deps <- computePackageDependencies(
+    bundleDir,
+    extraPackages,
+    quiet = TRUE,
+    checkLockfile = checkLockfile
+  )
   deps[c("Package", "Version", "Source", "Repository")]
 }
 
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -114,7 +114,7 @@ appDependencies <- function(
   appFiles = NULL,
   appFileManifest = NULL,
   appMode = NULL,
-  checkLockfile = TRUE
+  ignoreLockfile = FALSE
 ) {
   appFiles <- listDeploymentFiles(appDir, appFiles, appFileManifest)
   appMetadata <- appMetadata(appDir, appFiles = appFiles, appMode = appMode)
@@ -140,7 +140,7 @@ appDependencies <- function(
     bundleDir,
     extraPackages,
     quiet = TRUE,
-    checkLockfile = checkLockfile
+    ignoreLockfile = ignoreLockfile
   )
   deps[c("Package", "Version", "Source", "Repository")]
 }
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -113,7 +113,8 @@ appDependencies <- function(
   appDir = getwd(),
   appFiles = NULL,
   appFileManifest = NULL,
-  appMode = NULL
+  appMode = NULL,
+  checkLockfile = TRUE
 ) {
   appFiles <- listDeploymentFiles(appDir, appFiles, appFileManifest)
   appMetadata <- appMetadata(appDir, appFiles = appFiles, appMode = appMode)
@@ -135,7 +136,7 @@ appDependencies <- function(
   defer(unlink(bundleDir, recursive = TRUE))
 
   extraPackages <- inferRPackageDependencies(appMetadata)
-  deps <- computePackageDependencies(bundleDir, extraPackages, quiet = TRUE)
+  deps <- computePackageDependencies(bundleDir, extraPackages, quiet = TRUE, checkLockfile = checkLockfile)
   deps[c("Package", "Version", "Source", "Repository")]
 }
 
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -114,9 +114,9 @@ appDependencies <- function(
   appFiles = NULL,
   appFileManifest = NULL,
   appMode = NULL,
-  dependencySource = "strict"
+  dependencySource = c("strict", "library")
 ) {
-  dependencySource <- match.arg(dependencySource, c("strict", "library"))
+  dependencySource <- match.arg(dependencySource)
 
   appFiles <- listDeploymentFiles(appDir, appFiles, appFileManifest)
   appMetadata <- appMetadata(appDir, appFiles = appFiles, appMode = appMode)
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -114,9 +114,9 @@ appDependencies <- function(
   appFiles = NULL,
   appFileManifest = NULL,
   appMode = NULL,
-  dependencySource = "default"
+  dependencySource = "strict"
 ) {
-  dependencySource <- match.arg(dependencySource, c("default", "library"))
+  dependencySource <- match.arg(dependencySource, c("strict", "library"))
 
   appFiles <- listDeploymentFiles(appDir, appFiles, appFileManifest)
   appMetadata <- appMetadata(appDir, appFiles = appFiles, appMode = appMode)
--- a/R/appDependencies.R
+++ b/R/appDependencies.R
@@ -114,8 +114,10 @@ appDependencies <- function(
   appFiles = NULL,
   appFileManifest = NULL,
   appMode = NULL,
-  ignoreLockfile = FALSE
+  dependencySource = "default"
 ) {
+  dependencySource <- match.arg(dependencySource, c("default", "library"))
+
   appFiles <- listDeploymentFiles(appDir, appFiles, appFileManifest)
   appMetadata <- appMetadata(appDir, appFiles = appFiles, appMode = appMode)
   if (!needsR(appMetadata)) {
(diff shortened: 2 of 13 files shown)


marcosnav:
Specifically related to #1317, there is some interesting analysis on it.

The sync check itself (`parseRenvDependencies` comparing library vs lockfile versions) has existed since rsconnect 1.1.0. It wasn't new in 1.8.0. What changed is that the new `bundleAppDir()` code in 1.8.0 now copies the renv profile lockfile into the bundle directory even when the user didn't have an `renv.lock` at the root.

The key addition in bundle.R (commit 86d588c2, "Support renv profiles"):
```
  profileLockfile <- tryCatch(
    resolveRenvLockFile(appDir),
    error = function(e) NULL
  )
  if (!is.null(profileLockfile) && file.exists(profileLockfile)) {
    bundleLockfile <- file.path(bundleDir, "renv.lock")
    file.copy(profileLockfile, bundleLockfile, overwrite = TRUE)
  }
```

And `resolveRenvLockFile()` (which replaced the old `renvLockFile()`) now calls `renv::paths$lockfile(project = bundleDir)` first, which respects `RENV_PATHS_LOCKFILE` and renv profiles. If renv is installed and active in any form, this can resolve a lockfile that the old `file.exists(file.path(bundleDir, "renv.lock"))` check would not have found.

The user's CI workflow confirms this, "The project is out-of-sync -- use renv::status() for details" prints in both versions, meaning renv is present and active.

In `1.7.0`:
1. `computePackageDependencies()` checked `file.exists(renvLockFile(bundleDir))`, just `file.path(bundleDir, "renv.lock")`
2. If no root-level `renv.lock` existed in the bundle staging dir, it fell through to Path 3 (`snapshotRenvDependencies`), which creates a fresh lockfile and passes `snapshot = TRUE`... skipping the sync check

In `1.8.0`:
1. `bundleAppDir()` now proactively copies any profile-resolved lockfile into the bundle as `renv.lock`
2. `computePackageDependencies()` uses `resolveRenvLockFile()` which also finds renv-profile lockfiles
3. Now the lockfile exists, so it takes Path 2 (`parseRenvDependencies`) with `snapshot = FALSE`... the sync check runs and fails

So the key point here is, projects that have `renv` active but were previously taking the snapshot path (no sync check) are now hitting the parse path (with sync check) because the lockfile resolution became more aggressive.

The `checkLockfile` parameter we just added is the direct fix for this user `writeManifest(checkLockfile = FALSE)` will work in their CI workflow. 

@karawoo @jonkeane 
The deeper question is whether `bundleAppDir()` should be copying profile lockfiles into the bundle at all when the original `appDir` didn't have a root-level `renv.lock`.

This PR helps to alleviate that, still wondering if this is on the surprising/unexpected side.

karawoo:
Thanks for looking into this @marcosnav! For the case of #1317 I think something is missing, but I'm not sure what.

> computePackageDependencies() checked file.exists(renvLockFile(bundleDir)), just file.path(bundleDir, "renv.lock")
If no root-level renv.lock existed in the bundle staging dir, it fell through to Path 3 (snapshotRenvDependencies), which creates a fresh lockfile and passes snapshot = TRUE... skipping the sync check

I don't think this is correct. If renv was being used but the lockfile was not at the root of the project, deployment failed with this error: https://github.com/rstudio/rsconnect/issues/1122#issuecomment-2792107818. I reproduced this behavior locally with rsconnect 1.7.0

I also tried deploying a project using standard renv (without profiles) from rsconnect 1.7.0. Contrary to #1317 it failed with the same error that we currently see in 1.8.0. Based on my reading of the code this is expected, but I don't have an explanation as to how this would have succeeded in the user's environment with 1.7.0.

> The deeper question is whether bundleAppDir() should be copying profile lockfiles into the bundle at all when the original appDir didn't have a root-level renv.lock.

I do think we want to keep doing this; having renv.lock files at locations other than the project root is an officially supported way to use renv.

R/bundlePackageRenv.R
```diff
@@ -29,7 +29,20 @@ snapshotRenvDependencies <- function(
...
   )
-  renv::snapshot(bundleDir, packages = deps$Package, prompt = FALSE)
+  withCallingHandlers(
+    renv::snapshot(bundleDir, packages = deps$Package, prompt = FALSE),
+    error = function(err) {
+      cli::cli_abort(
+        c(
+          "Failed to snapshot dependencies with renv.",
+          i = "This can happen when a locally-developed package is not installed from a known source.",
+          i = "Use {.code renv::settings$ignored.packages(\"pkg\")} to exclude it from the snapshot.",
```
karawoo: I'd be hesitant to suggest this. If you tell renv to ignore a package that is actually used in the project, then running the content on Connect is most likely going to fail.

karawoo (commented):
As written, `checkLockfile = FALSE` actually does not trust the lockfile, it bypasses checking that the lockfile and library are in sync and records the package metadata from the _library_, NOT the lockfile. At a minimum we'd need to document this differently, but I would like to brainstorm ways we could actually just trust the lockfile.

novica:
> For the case of https://github.com/rstudio/rsconnect/issues/1317 I think something is missing, but I'm not sure what.

@karawoo  I updated my report with some more details. Hope it helps. I can also share links to the gh-actions logs since this is on a public repo.

R/bundlePackage.R
```diff
@@ -58,12 +61,15 @@ computePackageDependencies <- function(
       extraPackages,
       verbose = verbose
     )
-  } else if (!is.null(resolveRenvLockFile(bundleDir))) {
+  } else if (!ignoreLockfile && !is.null(resolveRenvLockFile(bundleDir))) {
```
jonkeane: Doing this here will mean that we fall through to the last `else` where we actually inspect the code fresh using `snapshotRenvDependencies()`. This should still respect the version(s) that are in the renv library, though IIUC, if the renv library isn't populated for some reason this snapshotting is going to go well / will fail.

NEWS.md
```diff
@@ -2,6 +2,17 @@
 
 * `deployApp()` with `logLevel = "verbose"` no longer errors using the `httr2` backend. (#1312)
 
+* `deployApp()`, `writeManifest()`, and `appDependencies()` gain an
+  `ignoreLockfile` parameter. Set `ignoreLockfile = TRUE` to ignore the
+  `renv.lock` file and resolve package dependencies from the local library
+  instead. This is useful when deploying from CI/CD environments where the
+  local library doesn't match the lockfile. (#1046, #1315, #1317)
```
jonkeane:
```suggestion
* `deployApp()`, `writeManifest()`, and `appDependencies()` gain an
  `ignoreLockfile` parameter. Set `ignoreLockfile = TRUE` to ignore the
  `renv.lock` file and resolve package dependencies by scanning the code. 
  The version that is recorded is what is installed in the libraries active in the 
  R session (i.e. what is displayed with `.libPaths()`). This is useful when 
  deploying from environments where there is a mismatch between the 
  renv lock file (#1046, #1315, #1317)
```

I've updated this wording a bit. I'm a little iffy on saying "use this in CI" cause really what we want is what is in #1319 for that. We also might want to defer referencing those issues until we solve that one too (since we believe that to be the real _real_ solution)

R/bundlePackage.R
marcosnav: Yes, good catch that's right, and for cases with unknown-source packages this will hit a circular message suggesting `ignoreLockfile = TRUE` when it won't really help for those cases.

`renv::snapshot()` has a `force = TRUE` parameter that bypasses the pre-flight validation. That's the check that fails on unknown-source packages. WDYT if we also set `force = TRUE` if `ignoreLockfile` is set?

jonkeane: This honestly might still lead folks into the same place they were before, so possibly not worth doing at all.   Looking at the root cause that Kara found: https://github.com/rstudio/rsconnect/issues/1317#issuecomment-434814685

Would the changes on this branch allow that to succeed?

marcosnav: Implemented this, let me know WDYT.

When `ignoreLockfile = TRUE`, pass `force = TRUE` to `renv::snapshot()` so it proceeds past pre-flight validation (e.g. unknown-source packages) instead of aborting. Warnings are still emitted. This makes `ignoreLockfile` useful for apps structured as packages, where the package itself triggers the validation failure.

marcosnav: > This honestly might still lead folks into the same place they were before, so possibly not worth doing at all. Looking at the root cause that Kara found: https://github.com/rstudio/rsconnect/issues/1317#issuecomment-434814685

> Would the changes on this branch allow that to succeed?

Ah, seems I had an outdated comment thread, taking a look

R/bundlePackageRenv.R
```diff
@@ -29,7 +30,24 @@ snapshotRenvDependencies <- function(
...
+      packages = deps$Package,
+      prompt = FALSE,
+      force = force
+    ),
+    error = function(err) {
+      cli::cli_abort(
+        c(
+          "Failed to snapshot dependencies with renv.",
+          i = "This can happen when a locally-developed package is not installed from a known source.",
+          i = "Set {.code ignoreLockfile = TRUE} to force dependency resolution from the local library."
```
karawoo: I don't think this error handling/messaging is quite right, because this error is only going to be thrown when we're _already_ doing dependency resolution from the local library via `renv::snapshot()`. Setting `ignoreLockfile = TRUE` will not help in this case.

Also, if we keep error handling here I'd use `tryCatch` instead of `withCallingHandlers` for this case.

R/deployApp.R
```diff
@@ -930,6 +945,20 @@ bundleApp <- function(
   bundlePath
 }
 
+confirmIgnoreLockfile <- function() {
+  cli::cli_warn(c(
+    "{.file renv.lock} will be ignored.",
+    "!" = "Package dependencies will be resolved from the local library instead.",
+    "!" = "The deployed content may differ from what the lockfile specifies."
```
karawoo: The message about "deployed content" might be a little confusing when this is used in `writeManifest()` where we're not actually deploying content. I think I would just remove this line.

R/deployApp.R
```diff
@@ -930,6 +945,20 @@ bundleApp <- function(
...
+  ))
+  if (is_interactive()) {
+    response <- cli_readline("Do you want to proceed? [Y/n]: ")
+    if (nzchar(response) && tolower(substring(response, 1, 1)) != "y") {
+      cli::cli_abort("Deployment cancelled.")
+    }
+  }
+}
```
karawoo: Ideally in an interactive session the message about ignoring renv.lock would be produced as part of the user prompt, not a separate warning. It is possible for warnings to be suppressed in the user's environment, in which case they would just see "Do you want to proceed? [Y/n]:" and have no idea what they are doing. Could we rework this so that the text is part of `cli_readline` for interactive use? 

For non-interactive use I think it makes sense to show some output but I'm not sure a warning is appropriate given that they have opted into this behavior, so maybe just `cli_inform`.

R/deployApp.R
```diff
@@ -263,12 +269,14 @@ deployApp <- function(
...
   envManagementPy = NULL,
-  packageRepositoryResolutionR = NULL
+  packageRepositoryResolutionR = NULL,
+  ignoreLockfile = FALSE
 ) {
   check_string(appDir)
   check_directory(appDir)
   appDir <- normalizePath(appDir)
 
+  check_bool(ignoreLockfile)
```
karawoo: is there a reason why we only `check_bool()` here and not in the other functions that take `ignoreLockfile`?

tests/testthat/test-bundlePackageRenv.R
```diff
@@ -133,6 +133,28 @@ test_that("errors if library and project are inconsistent", {
...
 })
 
+test_that("ignoreLockfile = TRUE bypasses lockfile and uses local library", {
+  skip_on_cran()
+  skip_if_not_installed("foreign")
+  skip_if_not_installed("MASS")
+
+  withr::local_options(renv.verbose = FALSE)
+
+  app_dir <- local_temp_app(list("foo.R" = "library(foreign); library(MASS)"))
```
karawoo:
```suggestion
  skip_if_not_installed("MASS")

  withr::local_options(renv.verbose = FALSE)

  app_dir <- local_temp_app(list("foo.R" = "library(MASS)"))
```

nit: we're not testing anything with foreign, so we can simplify the test to remove it.

NEWS.md
```diff
@@ -2,6 +2,19 @@
 
 * `deployApp()` with `logLevel = "verbose"` no longer errors using the `httr2` backend. (#1312)
 
+* `deployApp()`, `writeManifest()`, and `appDependencies()` gain an
+  `ignoreLockfile` parameter. Set `ignoreLockfile = TRUE` to ignore the
+  `renv.lock` file and resolve package dependencies by scanning the code. 
+  The version that is recorded is what is installed in the libraries active in the 
+  R session (i.e. what is displayed with `.libPaths()`). This is useful when 
+  deploying from environments where there is a mismatch between the 
+  renv lock file (#1046, #1315, #1317)
```
```suggestion
  deploying from environments where there is a mismatch between the 
  renv lock file and the user's environment. (#1046, #1315, #1317)
```

NEWS.md
```diff
@@ -2,6 +2,19 @@
...
+
+* Improved error messages when `renv::snapshot()` fails during dependency
+  discovery. The error now suggests actionable workarounds such as
+  `renv::settings$ignored.packages()` for locally-developed packages.
+  (#1078)
```
karawoo: I think we're not suggesting this anymore so can remove this line

R/deployApp.R
marcosnav: Thank you! addressed 👍

R/deployApp.R
marcosnav: This doesn't apply anymore

tests/testthat/test-bundlePackageRenv.R
marcosnav: removed it

NEWS.md
marcosnav: done

R/appDependencies.R
```diff
@@ -113,8 +113,11 @@ appDependencies <- function(
   appDir = getwd(),
   appFiles = NULL,
   appFileManifest = NULL,
-  appMode = NULL
+  appMode = NULL,
+  dependencySource = "strict"
 ) {
+  dependencySource <- match.arg(dependencySource, c("strict", "library"))
```
jonkeane: This is the idiomatic way to do this. It will show in autocomplete/docs better what the options are, the first one is the default.

```suggestion
  appMode = NULL,
  dependencySource = c("strict", "library")
) {
  dependencySource <- match.arg(dependencySource)
```

Though, I'm not sure our names are quite right here `dependencySource = "strict"` feels wrong "strict" isn't a source. Maybe that inconsistency is fine, but it sounds really strange to me seeing it written out like this.

R/bundlePackageRenv.R
```diff
@@ -29,7 +30,24 @@ snapshotRenvDependencies <- function(
...
+      bundleDir,
+      packages = deps$Package,
+      prompt = FALSE,
+      force = force
+    ),
+    error = function(err) {
+      cli::cli_abort(
+        c(
+          "Failed to snapshot dependencies with renv.",
+          i = "For example you have a locally-developed package that is not installed from a known source."
```
jonkeane:
```suggestion
          i = "For example, you have a locally-developed package that is installed from disk."
```

R/deployApp.R
```diff
@@ -201,6 +201,14 @@
 #'   `"legacy"`, `"lockfile"`, or `NULL`. The default, `NULL`, will not write
 #'   any values to the bundle manifest and Connect will fall back to the
 #'   server's package repository resolution strategy.
+#' @param dependencySource Controls how R package dependencies are resolved.
+#'   Must be one of `"strict"` or `"library"`. When `"strict"`, the
+#'   `renv.lock` file is used if present and must match the local library.
+#'   When `"library"`, the lockfile is ignored and dependencies are resolved
+#'   from the locally installed library instead. This is useful when the
```
jonkeane:
```suggestion
#'   When `"library"`, the lockfile is ignored and dependencies are resolved
#'   from the available local libraries instead. This is useful when the
```

R/deployApp.R
```diff
@@ -201,6 +201,14 @@
...
+#'   lockfile is out of sync with the local library and cannot be updated
+#'   (e.g. in CI/CD environments). Note that the deployed content will
```
jonkeane:
```suggestion
#'   lockfile is out of sync with the local library and cannot be updated. Note that the deployed content will
```

This actually isn't super helpful in CI/CD what would actually be helpful is to take the renv.lock file and just use that without needing to install all the packages.

jonkeane (commented):
This is so close! A few comments (mostly word smithing). 

It would be good to show off some CLI interactions of what happens when you get prompted here. How does this feel to use?

R/appDependencies.R
marcosnav: Ah, good to know, thank you

marcosnav:
@jonkeane Thank you for the review and suggestions, 
- propagated the match.arg pattern to the other methods as well
- added screenshots to the PR description with the behavior of what this PR introduces

R/bundlePackage.R
```diff
@@ -74,7 +82,8 @@ computePackageDependencies <- function(
       bundleDir,
       extraPackages,
       quiet = quiet,
-      verbose = verbose
+      verbose = verbose,
+      force = dependencySource == "library"
```
karawoo: This will generally be `force = TRUE` in this branch of the code. `force = TRUE` will bypass the pre-flight checks that normally error for things like packages installed from an unknown source. This means that the error message at bundlePackageRenv.R line 45 will be unlikely to be accurate (since `renv::snapshot()` would not error on these packages), and we may generate a lockfile that is not restorable.

marcosnav: Ah yes, good catch! The `force` flag was introduced in a previous iteration where we also wanted to fix locally dev pacakages but didn't get back to it. This is redundant now that this is a scenario we are not trying to fix as part of this work and also that it will be practically the same as `force = TRUE` as you mention. I'll remove it.

marcosnav: Done

karawoo (approved):

karawoo:
Regarding @jonkeane's comment about the argument name, I agree that `dependencySource` feels weird since `"strict"` isn't a source. we could do `dependencyResolution`? I don't feel strongly though, and otherwise this is looking good to me so I gave it the ✅

marcosnav:
> Regarding @jonkeane's comment about the argument name, I agree that `dependencySource` feels weird since `"strict"` isn't a source. we could do `dependencyResolution`? I don't feel strongly though, and otherwise this is looking good to me so I gave it the ✅

Ah, `dependencyResolution` is sounding much nicer. If @jonkeane agrees on that one I'm happy to rename it all the way 👍

karawoo (approved):

PR #1321, opened 2026-05-01
fix: fix history file maintenance race and cap size

### LINKED ISSUE #1320: improve history.dcf management
~RStudio may not consume this information. If this is confirmed, remove the creation and maintenance of the history.dcf file.~

Workbench consumes the history.dcf (see WorkspacesApi.cpp).

We have observed a ~100G history.dcf and a ~200G history.new.dcf file.

A couple problems:

1. Do not grow forever.
2. Do not add duplicates.

Duplicate calls to `saveDeployment()` happen naturally during `deployApp()`. The regular `writeDeploymentRecord()` save overwrites previous data. The deployment history file only prepends (by writing to a copy before overwriting the original).

We still do not know what caused this particular user history to double in size (with what appeared to be a single update, but that is speculation).

aronatkins:
The history file could rapidly expand in size during concurrent deployments. Previously, history mutations landed in history.new.dcf, meaning two calls could each append the existing history.dcf contents, doubling in size each time the race is hit.

Now, write to a uniquely named temporary file. During concurrent deployments, one of the history entries may be lost.

The history file is now capped at 100 records; the most recent are preserved. If the history file has grown beyond 1MB, it is reset. 1M fits 1k-2k entries; for most users, a file this large indicates that they have hit the growth explosion.

fixes #1320

### DIFF
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -194,7 +194,7 @@ addToDeploymentHistory <- function(appPath, deploymentRecord) {
 
   # Truncate the history file when it grows beyond this size.
   # Do not read large files; they may be enormous (#1320).
-  max_bytes <- getOption("rsconnect.max.history.bytes", 1e6)
+  max_bytes <- getOption("rsconnect.max.history.bytes", 1024^2)
 
   # When the history file is reasonably sized, preserve this number of records.
   max_records <- max(1L, getOption("rsconnect.max.history.records", 100L))
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -187,23 +187,36 @@ writeDeploymentRecord <- function(record, filePath) {
   write.dcf(record, filePath, width = 4096)
 }
 
-# Workbench uses to show a list of recently deployed content on user dashboard
+# Workbench uses to show a list of recently deployed content on user dashboard.
 addToDeploymentHistory <- function(appPath, deploymentRecord) {
-  # add the appPath to the deploymentRecord
+  ## deployment history is global; reference back to the app path.
   deploymentRecord$appPath <- appPath
 
-  # write new history file
-  newHistory <- deploymentHistoryPath(new = TRUE)
-  writeDeploymentRecord(deploymentRecord, newHistory)
+  # Truncate the history file when it grows beyond this size.
+  # Do not read large files; they may be enormous (#1320).
+  max_bytes <- getOption("rsconnect.max.history.bytes", 1e6)
+
+  # When the history file is reasonably sized, preserve this number of records.
+  max_records <- getOption("rsconnect.max.history.records", 100L)
 
   history <- deploymentHistoryPath()
-  # append existing history to new history
-  if (file.exists(history)) {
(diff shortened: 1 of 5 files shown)


timtalbot (approved):
I defer to the R experts but this looks like it thoroughly solves the issue by capping the record count and file size.

aronatkins:
> I defer to the R experts but this looks like it thoroughly solves the issue by capping the record count and file size.

The core of the fix is using a uniquely named temporary file when adjusting the history. The caps on the number of records and file size are to help recover from existing large files.

R/deployments.R
```diff
@@ -187,23 +187,36 @@ writeDeploymentRecord <- function(record, filePath) {
...
+  max_bytes <- getOption("rsconnect.max.history.bytes", 1024^2)
+
+  # When the history file is reasonably sized, preserve this number of records.
+  max_records <- max(1L, getOption("rsconnect.max.history.records", 100L))
 
   history <- deploymentHistoryPath()
-  # append existing history to new history
-  if (file.exists(history)) {
+  prior <- if (file.exists(history) && file.size(history) <= max_bytes) {
+    read.dcf(history)
```
karawoo, line 204: do we actually need to parse the dcf, or would it be enough to `readLines()`? I'm wondering if we could `readLines(n = )` where `n` is some reasonable number of lines such that we can still keep recent past records from files that have grown too large, and seamlessly discard the excess.

aronatkins: This is a good question.

The number of lines in a deployment record has changed over time. We can certainly make a guess and read the first N*M lines. That can then feed into `read.dcf`, but the final record is unlikely to be complete and needs to be discarded. We would need to know if `readLines()` consumed the whole file or not when making this determination.

I chose the simpler option: Discard massive files (unlikely in practice, but a protection) and use `read.dcf` to consume the whole file, which means we don't need to fret about whether or not the final record is valid.

But maybe you were suggesting something slightly different: Maintain the last N lines of `history.dcf` and be comfortable with the fact that the final record in that file may be partial.

The other option that I considered, which is similar but slightly more complex, is to read a line at a time from the file and detect the blank-line record separator. Read N records and stop reading.

Any or all of these options are in reach. I chose the current technique because it didn't require other heuristics or manual DCF record probing. Discarding the `history.dcf` only when it was very large felt like a reasonable compromise.

Do any of these other choices feel better?

karawoo: Thanks for the explanation. Reading it laid out this way I think these other options are more trouble than they're really worth. It is a pretty rare edge case that produces the giant files. I'll approve as-is.

karawoo (approved):

R/deployments.R
aronatkins: Leaving this conversation as unresolved only so it remains visible to other folks who might see the PR.

aronatkins:
One additional note: I'm asking the Workbench folks if they're still consuming this data. I'll follow up once I learn the answer.

PR #1323, opened 2026-05-06
add support for deploying nodejs apps with rsconnect

### LINKED ISSUE #1322: Support deploying nodejs content
Support detecting and deploying nodejs content to Connect

see also: 
* rsconnect-python: https://github.com/posit-dev/rsconnect-python/pull/761
* publisher: https://github.com/posit-dev/publisher/pull/4122

karawoo:
closes #1322 

Detects nodejs content on the basis of a `"package.json"` file (in the absence of other R/Python/Quarto indicators). Checks the Connect version we're deploying to and errors if it's <2026.04, since node.js content is not supported on earlier versions of Connect.


tests/testthat/test-appMetadata-nodejs.R
```diff
@@ -0,0 +1,219 @@
+test_that("package.json is detected as nodejs", {
+  dir <- local_temp_app(list(
+    "package.json" = '{"name": "test", "main": "app.js"}',
+    "package-lock.json" = '{"lockfileVersion": 3}',
+    "app.js" = ""
+  ))
+  files <- list.files(dir)
+  result <- inferAppMode(dir, files)
```
jonkeane: Could this be changed to something like this below? That is — do we need to write these files to disk for `inferAppMode` to work?

```suggestion
  result <- inferAppMode(dir,c("package.json", "package-lock.json", "app.js")
```

Or maybe there's a refactoring of `inferAppMode` that would allow that. These tests feel like there's more setup than should be necessary for the unit-style test the matrix of all the possible combinations of files in the tests below.

jonkeane (approved):
This looks good. I have a more meta point about some of the testing we're doing here. I like the coverage, but it seems like maybe we have a slight refactoring for the inferappmode that would make it more encapsulated and easier to test (which definitely does not need to be part of this PR!)

tests/testthat/test-appMetadata-nodejs.R
karawoo: yeah, I agree this could be better but it's already a similar situation for the other app modes so I would be inclined to address it separately. but I'll push an update that puts the nodejs tests closer to those for other app modes so there are fewer places to update if/when we tackle this in the future.

PR #1324, opened 2026-05-15
add function for adding client credentials for Connect Cloud

mslynch:
To bring Posit Connect Cloud into feature parity with shinyappps.io, we'd like to provide a noninteractive way to authenticate for publishing bundled content. This PR adds a `connectCloudClientCredentials` method that can be used to do so using oauth2 client credentials, rather than the interactive device auth flow used by `connectCloudUser`:

```R
rsconnect::connectCloudCredentials(
  clientId="<client_id>",
  clientSecret="<secret>",
  accountName='mslynch'
)
```

This feature is currently only available in the staging environment. To test, create a credential [here](https://login.staging.posit.cloud/identity/credentials) and set `options(rsconnect.connect_cloud_environment="staging")`.


R/accounts.R
```diff
@@ -250,14 +250,142 @@ connectCloudUser <- function(launch.browser = TRUE) {
...
+#' @description
+#' `connectCloudClientCredentials()` registers a Posit Connect Cloud account
+#' using an OAuth 2.0 `client_credentials` grant provided by the Posit Cloud
+#' auth service (https://login.posit.cloud/identity/credentials). Use this
+#' function to authenticate in non-interactive contexts.
+#'
+#' If `accountName` is omitted and the credentials can publish to exactly one
+#' Connect Cloud account, that account is selected automatically. If they can
+#' publish to multiple accounts, you will be prompted to choose; this requires
+#' an interactive session.
```
karawoo: I think this documentation is a bit confusing: this function is intended for non-interactive sessions, but also sometimes might require an interactive session (but if you run it in an interactive session it warns you).

The actual behavior seems ok since you can pass an `accountName` to bypass the interactive menu, so I think I'd just update the documentation to encourage that rather than the interactive fallback. Alternatively, I think it would also be okay to require an `accountName` and skip the interactive menu option altogether.

tests/testthat/test-accounts.R
```diff
@@ -177,6 +177,194 @@ test_that("connectSPCSUser works with explicit snowflakeConnectionName", {
...
+      registered <<- list(...)
+    }
+  )
+
+  simulate_user_input(2)
+  connectCloudClientCredentials(
+    clientId = "client-1",
+    clientSecret = "secret-1",
+    quiet = TRUE
+  )
```
karawoo: in an interactive session when running `devtools::test()` this throws the warning about being intended for non-interactive use. could we please `suppressWarnings()` here to avoid that?

tests/testthat/test-client-connectCloud.R
```diff
@@ -301,3 +305,74 @@ test_that("withTokenRefreshRetry handles 401 with successful token refresh", {
   expect_equal(call_count, 2)
   expect_true(register_called)
 })
+
+test_that("withTokenRefreshRetry uses client_credentials when clientSecret is set", {
+  skip_if_not_installed("webfakes")
```
karawoo: it doesn't look like this test uses webfakes so I think we can remove this.
```suggestion
```

mslynch:
@karawoo I've updated the PR to require an `accountName` and removed the unnecessary `skip_if_not_installed`.

aronatkins (approved):
Nothing stands out to me, but I'll let @karawoo chime in, too.

karawoo (approved):

Issue #1326, opened 2026-05-20
integration test failures caused by missing libuv.so.1


aronatkins:
The Connect preview and Connect release integration tests fail with the following error:

```
Error in dyn.load(file, DLLpath = DLLpath, ...) : 
  unable to load shared object '/opt/rstudio-connect/mnt/app/packrat/lib/x86_64-pc-linux-gnu/4.6.0/fs/libs/fs.so':
  libuv.so.1: cannot open shared object file: No such file or directory
Calls: loadNamespace -> library.dynam -> dyn.load
Execution halted
```

Example runs:
* https://github.com/rstudio/rsconnect/actions/runs/26164024795/job/76963087529
* https://github.com/rstudio/rsconnect/actions/runs/26164024795/job/76963087526

aronatkins:
Most recent connect release pass (https://github.com/rstudio/rsconnect/actions/runs/25925578577/job/76205818493)

```
Pulling image rstudio/rstudio-connect:jammy
...
Linux distribution: Ubuntu 22.04.5 LTS (jammy)
```

Current connect release failure (https://github.com/rstudio/rsconnect/actions/runs/26164024795/job/76963087529)

```
Pulling image ghcr.io/posit-dev/connect:latest
...
Linux distribution: Ubuntu 24.04.4 LTS (noble)
```

The underlying image has changed and these tests now fail.

nealrichardson:
https://github.com/posit-dev/images-connect/pull/92 sounds relevant perhaps, along with https://github.com/posit-dev/with-connect/pull/43. 

@bschwedler any thoughts?

aronatkins:
Waiting on https://github.com/posit-dev/images-connect/pull/107

karawoo:
integration tests are passing now ✅

PR #1328, opened 2026-05-20
avoid using MASS and foreign, as they have R-version restrictions

### LINKED ISSUE #1327: oldrel-3 and oldrel-4 failures caused by MASS install failures
The oldrel-3 (R-4.3.3) and oldrel-4 (R-4.2.3) CI runs fail with errors like:

```
── Error ('test-bundlePackageRenv.R:142:3'): errors if library and project are inconsistent ──
Error: Error: package 'MASS' is not available
Backtrace:
    ▆
 1. └─renv::record("MASS@0.1.1", project = app_dir) at test-bundlePackageRenv.R:142:3
 2.   └─renv:::renv_record_enrich(records[[i]])
 3.     ├─renv:::memoize(...)
 4.     │ └─envir[[key]] %||% value
 5.     └─renv:::renv_record_enrich_impl(record)
 6.       └─renv:::renv_graph_description(record)
 7.         ├─renv:::case(...)
 8.         └─renv:::renv_graph_description_repository(record)
 9.           └─renv:::stopf("package '%s' is not available", package)
```

Example runs:
* https://github.com/rstudio/rsconnect/actions/runs/26164024817/job/76963087578
* https://github.com/rstudio/rsconnect/actions/runs/26164024817/job/76963087563

aronatkins:
Package version restrictions can cause tests to fail if those tests need to obtain a particular package from the repository and the repository version is incompatible with the active version of R. The most current version of MASS, for example, requires R > 4.4.

Note: The integration test failures are tracked by https://github.com/rstudio/rsconnect/issues/1326.

### DIFF
--- a/tests/testthat/test-bundlePackageRenv.R
+++ b/tests/testthat/test-bundlePackageRenv.R
@@ -132,14 +132,11 @@ test_that("gets DESCRIPTION from renv & system libraries", {
 
 
 test_that("errors if library and project are inconsistent", {
-  skip_if_not_installed("foreign")
-  skip_if_not_installed("MASS")
-
   withr::local_options(renv.verbose = FALSE)
 
-  app_dir <- local_temp_app(list("foo.R" = "library(foreign); library(MASS)"))
+  app_dir <- local_temp_app(list("foo.R" = "library(withr)"))
   renv::snapshot(app_dir, prompt = FALSE)
-  renv::record("MASS@0.1.1", project = app_dir)
+  renv::record("withr@0.1.1", project = app_dir)
 
   expect_snapshot(
     parseRenvDependencies(file.path(app_dir, "renv.lock"), app_dir),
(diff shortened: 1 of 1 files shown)


tests/testthat/test-bundlePackageRenv.R
```diff
@@ -132,14 +132,11 @@ test_that("gets DESCRIPTION from renv & system libraries", {
 
 
 test_that("errors if library and project are inconsistent", {
-  skip_if_not_installed("foreign")
-  skip_if_not_installed("MASS")
-
   withr::local_options(renv.verbose = FALSE)
 
-  app_dir <- local_temp_app(list("foo.R" = "library(foreign); library(MASS)"))
+  app_dir <- local_temp_app(list("foo.R" = "library(withr)"))
```
nealrichardson, line 137: How about using a package like `utils` that is included with R? IDK if renv does different things with those packages but you know they will always be available and will always have the same version number as the current R version.

aronatkins: withr is a dev dependency of rsconnect and a CRAN package without R version constraints.

karawoo (approved):
Ah I see, I guess this is a consequence of https://github.com/rstudio/renv/pull/2297. The fix makes sense to me.

kevinushey (approved):
LGTM, but I wonder if this implies a bug or misbehavior in renv::record...

PR #1330, opened 2026-06-02
add nodejs deployment integration test

karawoo:


karawoo:
closing for now, we need an image with connect + node to run these tests.

PR #1331, opened 2026-06-04
report more details if addServer fails

### LINKED ISSUE #1197: provide underlying error when addServer() fails

karawoo:
closes #1197, with less verbose output than #1198. for HTTP errors we report the status, for non-HTTP errors we report the full error message (as it should be fairly short).

```r
rsconnect::addServer("https://this-is-not-a-valid-server")
#> Error in `rsconnect::addServer()`:
#> ! `url` does not appear to be a Posit Connect server.
#> ✖ Failed to perform HTTP request. Caused by error in `curl::curl_fetch_memory()`: ! Could not resolve hostname
#>   [this-is-not-a-valid-server]: Could not resolve host: this-is-not-a-valid-server
#> Run `rlang::last_trace()` to see where the error occurred.
```

```r
rsconnect::addServer("https://google.com")
#> Error in `rsconnect::addServer()`:
#> ! `url` does not appear to be a Posit Connect server.
#> ✖ <https://google.com/__api__/server_settings> responded with HTTP status 404.
#> Run `rlang::last_trace()` to see where the error occurred.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,9 @@
 
 * Improved error messages when `renv::snapshot()` fails during dependency
   discovery. (#1078)
+  
+  * `addServer()` now reports the underlying reason a URL was rejected (e.g.
+  connection error, TLS/certificate problem, or unexpected HTTP status). (#1197)
 
 # rsconnect 1.8.0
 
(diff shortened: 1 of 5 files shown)


aronatkins (approved):
Thanks. This is definitely an improvement.

The response body might be useful, but we can revisit once we have a specific scenario that needs to show it.

PR #1336, opened 2026-06-10
there was no 1.9

aronatkins:


aronatkins:
This branch exists in order to run the pkgdown workflow. main already has the NEWS fix.

PR #1340, opened 2026-06-26
fix: only resolve bioc repositories when a project includes a bioc-sourced package

### LINKED ISSUE #1337: writeManifest()/deployApp() installs BiocManager and contacts bioconductor.org for projects with no Bioconductor dependencies, breaking air-gapped deployments
## Summary

Starting with rsconnect 1.10.0, dependency capture resolves Bioconductor repositories on every `writeManifest()`/`deployApp()` call, even when the project has no package with `Source == "Bioconductor"`. This installs BiocManager into the project library and validates the Bioconductor version online against `bioconductor.org`. In an air-gapped or offline environment the version check fails and capture stops, so CRAN-only content can no longer be published.

This is a regression from 1.8.0, where the Bioconductor repository list was only computed when a Bioconductor-sourced package was actually present.

## Expected behavior

A project whose dependencies are all from CRAN-like repositories should not install BiocManager or contact bioconductor.org during dependency capture.

## Actual behavior

On a CRAN-only renv project, `writeManifest()` installs BiocManager and then errors:

```
ℹ Capturing R dependencies from renv.lock
- Installing BiocManager ...
Error: invalid version specification 'unknown version: Bioconductor version cannot be validated; no internet connection?
    See #troubleshooting section in vignette'
```

## Root cause

In `R/bundlePackageRenv.R`, `parseRenvDependencies()` evaluates the Bioconductor repositories eagerly:

```r
bioc <- biocRepos(bundleDir)            # always runs, every deploy
if (any(bioc %in% repos)) {
  biocPkgs <- NULL
} else {
  biocPkgs <- availablePackages(bioc)
}
deps <- standardizeRenvPackages(renvLock$Packages, repos, biocPackages = biocPkgs)
```

`biocRepos()` calls `renv:::renv_bioconductor_repos()`, which installs BiocManager and resolves the Bioconductor version (a network call to `bioconductor.org/config.yaml`).

In 1.8.0 the same value was passed as a lazy argument and only forced for Bioconductor-sourced packages:

```r
deps <- standardizeRenvPackages(
  renvLock$Packages,
  repos,
  biocPackages = biocPackages(bundleDir)   # promise; only forced for Source == "Bioconductor"
)
```

The eager evaluation was introduced by commit 5ea5a31 ("fix bioconductor repository resolution", PR #1316), which is the PR that fixed issue #1314. `git tag --contains 5ea5a31` returns `v1.10.0` only, so this behavior is specific to the 1.10.0 release. Note that 1.9.0 was never published to CRAN; the CRAN archive goes directly from 1.8.0 (2026-04-10) to the current 1.10.0, so users upgrade from 1.8.0 straight into this behavior.

## Reproduction

1. R session with `repos` pointing only at a CRAN-like mirror (no Bioconductor repo), `R_BIOC_VERSION` unset.
2. A renv project whose renv.lock contains only CRAN packages.
3. Make `bioconductor.org` unreachable (simulate air gap), e.g. blackhole it in /etc/hosts.
4. Fresh R session, then `rsconnect::writeManifest()`.

Result: BiocManager is installed and capture fails with the version-validation error above.

## Suggested fix

Only resolve Bioconductor repositories when the project actually contains a Bioconductor-sourced package, restoring the lazy/conditional behavior from 1.8.0. Computing `bioc` only when needed avoids installing BiocManager and contacting bioconductor.org for CRAN-only content.

## Versions

- rsconnect 1.10.0 (affected). Last working: 1.8.0. 1.9.0 was never released to CRAN.
- Introduced by PR #1316 (commit 5ea5a31), the fix for issue #1314.
- renv (bundled Bioconductor resolution)
- BiocManager 1.30.27
- R 4.2, Linux x86_64

### Your environment

_No response_

joshyam-k:
Fixes #1337

Instead of eagerly evaluating Bioconductor repositories, only do so when the project actually contains a Bioconductor-sourced package. This PR also creates a `isBiocPackage` helper that gets reused across `R/bundlePackageRenv.R`

Following the reproduction steps in #1337:

### Before

```console
> rsconnect::writeManifest(appDir = "proj")
ℹ Capturing R dependencies from renv.lock
  Error: invalid version specification 'unknown version: Bioconductor version cannot be validated;
      no internet connection?
      See #troubleshooting section in vignette'
```

### After

```console
> options(repos = c(CRAN = "https://cloud.r-project.org"))
> rsconnect::writeManifest(appDir = "proj")
ℹ Capturing R dependencies from renv.lock
✔ Found 2 dependencies
```

### DIFF
--- a/R/bundlePackageRenv.R
+++ b/R/bundlePackageRenv.R
@@ -59,12 +59,18 @@ parseRenvDependencies <- function(lockfile, bundleDir, snapshot = FALSE) {
     vapply(renvLock$R$Repositories, "[[", "URL", FUN.VALUE = character(1)),
     vapply(renvLock$R$Repositories, "[[", "Name", FUN.VALUE = character(1))
   )
-  bioc <- biocRepos(bundleDir)
-  if (any(bioc %in% repos)) {
-    biocPkgs <- NULL
+  # Only resolve Bioconductor repositories when the project actually contains a
+  # Bioconductor-sourced package.
+  if (hasBiocPackage(renvLock$Packages)) {
+    bioc <- biocRepos(bundleDir)
+    if (any(bioc %in% repos)) {
+      biocPkgs <- NULL
+    } else {
+      signal("evaluating", class = "rsconnect_biocRepos")
+      biocPkgs <- availablePackages(bioc)
+    }
   } else {
-    signal("evaluating", class = "rsconnect_biocRepos")
-    biocPkgs <- availablePackages(bioc)
+    biocPkgs <- NULL
   }
   deps <- standardizeRenvPackages(
     renvLock$Packages,
@@ -118,13 +124,9 @@ standardizeRenvPackage <- function(
   # Convert renv source to manifest source/repository
   # https://github.com/rstudio/renv/blob/0.17.2/R/snapshot.R#L730-L773
 
-  if (
-    is.null(pkg$Repository) &&
-      !is.null(pkg$RemoteRepos) &&
-      grepl("bioconductor.org", pkg$RemoteRepos)
-  ) {
-    # Work around bug where renv fails to detect BioC package installed by pak
-    # https://github.com/rstudio/renv/issues/1202
+  # Normalize Bioconductor packages to Source == "Bioconductor", including those
+  # installed by pak that renv records without a Bioconductor source.
+  if (isBiocPackage(pkg)) {
     pkg$Source <- "Bioconductor"
   }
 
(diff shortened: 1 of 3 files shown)


R/bundlePackageRenv.R
```diff
@@ -59,12 +59,18 @@ parseRenvDependencies <- function(lockfile, bundleDir, snapshot = FALSE) {
     vapply(renvLock$R$Repositories, "[[", "URL", FUN.VALUE = character(1)),
     vapply(renvLock$R$Repositories, "[[", "Name", FUN.VALUE = character(1))
   )
-  bioc <- biocRepos(bundleDir)
-  if (any(bioc %in% repos)) {
-    biocPkgs <- NULL
+  # Only resolve Bioconductor repositories when the project actually contains a
```
joshyam-k, line 62: sorry- I hate when github makes a diff look confusing like this. We're just wrapping the lines that were previously 62-64 and 66-67 with a new `hasBiocPackage(...)` gate.

jonkeane (approved):
This looks good to me, I would love to get @karawoo 's take as well, but tests pass and the code looks right.

tests/testthat/test-bundlePackageRenv.R
```diff
@@ -130,6 +130,29 @@ test_that("gets DESCRIPTION from renv & system libraries", {
...
+    biocRepos = function(...) stop("biocRepos() should not be called")
+  )
+
+  app_dir <- local_temp_app(list("foo.R" = "library(foreign); library(MASS)"))
+  renv::snapshot(app_dir, prompt = FALSE)
+
+  expect_no_condition(
+    deps <- parseRenvDependencies(file.path(app_dir, "renv.lock"), app_dir),
+    class = "rsconnect_biocRepos"
+  )
```
karawoo: I think this `expect_no_condition` is redundant since we're mocking `biocRepos` to error, which would immediately fail the test if the code were trying to look up bioconductor packages. It's harmless to leave it though I think.

karawoo (approved):

PR #1341, opened 2026-06-29
default `contentCategory = "mcp"` for mcptools MCP servers

simonpcouch:
Related to posit-dev/mcptools#106 and posit-dev/plumber2#89. mcptools is now an adopter of plumber2's `_server.yml` standard. For now, I'm documenting in mcptools that folks should set `deployApp(contentCategory)` explicitly, but this change would let them just `deployApp()`. :)


simonpcouch:
Closing in favor of #1343.

PR #1343, opened 2026-06-30
default `contentCategory = "mcp"` for mcptools MCP servers

simonpcouch:
Same changes as #1341, but PRed from a branch rather than a fork. See that PR description for context. cc @joshyam-k, thanks for the add!


joshyam-k (approved):
neat- thanks!

PR #1347, opened 2026-07-07
Notify about newer versions of rsconnect

### LINKED ISSUE #1342: Notify users when a new version of rsconnect is available
This request is coming from a painful situation I encountered when I was using push-button deployment in RStudio. 

I'm on R version 4.4.1, for which CRAN only has prebuilt rsconnect binaries up to version 1.7.0. 

```
> install.packages("rsconnect")
Installing package into ‘/Users/zverham/Library/R/arm64/4.4/library’
(as ‘lib’ is unspecified)
trying URL 'https://cloud.r-project.org/bin/macosx/big-sur-arm64/contrib/4.4/rsconnect_1.7.0.tgz'
Content type 'application/x-gzip' length 822937 bytes (803 KB)
```

The actual current version of rsconnect is 1.10.0 - more recent versions of rsconnect use the v1 connect apis for publishing. The unversioned APIs have been removed from the product. 

This means that, for me, push-button deployment fails with the following error: 

```
Error in `POST()`:
! <http://localhost:3939/__api__/applications/3/upload> failed with HTTP
  status 404
404 page not found
Backtrace:
    ▆
 1. └─rsconnect::deployApp(...)
 2.   └─client$uploadApplication(application$id, bundlePath)
 3.     └─rsconnect:::POST(...)
 4.       └─rsconnect:::httpRequestWithBody(...)
 5.         └─rsconnect:::handleResponse(...)
 6.           └─rsconnect (local) reportError(response$content)
 7.             └─cli::cli_abort(...)
 8.               └─rlang::abort(...)
```

Without any clear indicator that (a) I'm on an out-of-date rsconnect version, or (b) that install.packages on its own won't give me the most recent rsconnect version. I have to either install from source or target a PPM repository that provides the prebuilt binary. 

The actual situation I was in has to do with how CRAN manages binary builds on older R versions (I think) - but one thing that would have made it easier to understand what was happening would be having rsconnect detect if a newer version exists and telling me.

karawoo:
Fixes #1342 

This checks for newer versions of rsconnect in the user's configured repositories at two times:
* when the package is attached (in an interactive session)
* on an HTTP error

The latter is arguably sort of a weird place to check and report a newer version, but it is the only case that would have saved @zackverham in #1342 because he was using push-button deployment in RStudio, not deploying from the R console. We intentionally do not imply that upgrading rsconnect will fix whatever error the user hit since we have no way of knowing whether that's true.

The check looks at the user's configured repositories and does not error if it can't determine the latest version. Requests are limited to `rsconnect.update_check_timeout` (default 2s) so we don't delay `library(rsconnect)` or error reporting excessively. Results are also cached (default 1 day). The whole feature can be turned off with `options(rsconnect.check_updates = FALSE)` or `RSCONNECT_CHECK_UPDATES=false`.

### DIFF
--- a/R/cookies.R
+++ b/R/cookies.R
@@ -71,15 +71,17 @@ storeCookies <- function(requestURL, cookieHeaders) {
   lapply(cookies, function(co) {
     # Remove any duplicates
     # RFC says duplicate cookies are ones that have the same domain, name, and path
-    hostCookies <<- hostCookies[ # nolint
+    # nolint start
+    hostCookies <<- hostCookies[
       !(co$name == hostCookies$name & co$path == hostCookies$path),
     ]
 
     # append this new cookie on
-    hostCookies <<- rbind( # nolint
+    hostCookies <<- rbind(
       as.data.frame(co, stringsAsFactors = FALSE),
       hostCookies
     )
+    # nolint end
   })
 
   # Save this host's cookies into the cookies store.
--- a/R/cookies.R
+++ b/R/cookies.R
@@ -71,12 +71,12 @@ storeCookies <- function(requestURL, cookieHeaders) {
   lapply(cookies, function(co) {
     # Remove any duplicates
     # RFC says duplicate cookies are ones that have the same domain, name, and path
-    hostCookies <<- hostCookies[
+    hostCookies <<- hostCookies[ # nolint
       !(co$name == hostCookies$name & co$path == hostCookies$path),
     ]
 
     # append this new cookie on
-    hostCookies <<- rbind(
+    hostCookies <<- rbind( # nolint
       as.data.frame(co, stringsAsFactors = FALSE),
       hostCookies
     )
--- a/R/version-check.R
+++ b/R/version-check.R
@@ -230,6 +230,23 @@ updateStartupMessage <- function(installed, latest) {
   )
 }
 
+# Emitted at the start of every deployApp() call so that captured deploy
+# logs always record the rsconnect version in use, plus the newer available
+# version when one is known.
+deployVersionMessage <- function(installed, latest = NULL) {
+  if (is.null(latest)) {
+    paste0("This is rsconnect ", installed)
+  } else {
+    paste0(
+      "This is rsconnect ",
+      installed,
+      " (",
+      latest,
+      " is latest available)"
+    )
+  }
+}
+
 # A neutral bullet appended to HTTP error messages. Deliberately avoids
 # implying that updating will resolve the error -- rsconnect has no way of
 # knowing whether an outdated version caused the failure.
--- a/R/version-check.R
+++ b/R/version-check.R
@@ -10,71 +10,201 @@
 #
 # The check never blocks deployment or throws: any failure (offline,
 # air-gapped repositories, timeout, etc.) is swallowed and simply results in
-# no notifiication.
+# no notification. The result (including a failed/unknown result) is cached
+# to disk for rsconnect.update_check_interval seconds (default 1 day), so
+# the network is only actually touched once per interval.
+
+# Fetches and parses a single repo's PACKAGES index, returning the version
+# listed for `pkg`, or NULL if it can't be determined. Uses httr2 rather
+# than available.packages()/download.file() because httr2::req_timeout()
+# reliably bounds the connection phase too -- options(timeout=) does not.
+latestVersionInRepo <- function(pkg, repo, timeout) {
+  url <- paste0(sub("/+$", "", repo), "/src/contrib/PACKAGES")
+
+  packages <- tryCatch(
+    httr2::request(url) |>
+      httr2::req_timeout(timeout) |>
+      httr2::req_perform() |>
+      httr2::resp_body_string(),
+    error = function(e) NULL
+  )
+
+  if (is.null(packages)) {
+    return(NULL)
+  }
+
+  parsePackagesVersion(packages, pkg)
+}
+
+# Pure parsing step, split out from latestVersionInRepo() so it's testable
+# without a network call.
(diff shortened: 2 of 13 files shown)


edavidaja (approved):
I'm kind of meh on messaging in the http error case but otherwise lgtm

karawoo:
I'm also a bit meh on it, but included it because in the push-button publishing case the `.onAttach` hook doesn't really help. I could go either way though. @aronatkins do you have any strong opinions here?

R/version-check.R
```diff
@@ -0,0 +1,244 @@
...
+      idx <- match(pkg, fields[, "Package"])
+      if (is.na(idx)) NULL else unname(fields[idx, "Version"])
+    },
+    error = function(e) NULL
+  )
+}
+
+# Returns the newest version of `pkg` available from `repos`, or NULL if it
+# can't be determined.
+latestAvailableVersion <- function(
```
aronatkins, line 54: Confirmed:

```r
rsconnect:::latestAvailableVersion()
#> [1] "1.10.1"
rsconnect:::latestAvailableVersion("shiny")
#> [1] "1.14.0"
```

R/version-check.R
```diff
@@ -0,0 +1,244 @@
...
+# air-gapped repositories, timeout, etc.) is swallowed and simply results in
+# no notification. The result (including a failed/unknown result) is cached
+# to disk for rsconnect.update_check_interval seconds (default 1 day), so
+# the network is only actually touched once per interval.
+
+# Fetches and parses a single repo's PACKAGES index, returning the version
+# listed for `pkg`, or NULL if it can't be determined. Uses httr2 rather
+# than available.packages()/download.file() because httr2::req_timeout()
+# reliably bounds the connection phase too -- options(timeout=) does not.
+latestVersionInRepo <- function(pkg, repo, timeout) {
```
aronatkins, line 21: Confirmed (manually) the following:

```r
rsconnect:::latestVersionInRepo("rsconnect","https://cran.rstudio.com", 2)
#> [1] "1.10.1"
rsconnect:::latestVersionInRepo("rsconnect","https://cran.rstudio.com", 1)
#> [1] "1.10.1"
rsconnect:::latestVersionInRepo("rsconnect","https://cran.rstudio.com", 0.1)
#> NULL
```

R/version-check.R
```diff
@@ -0,0 +1,244 @@
...
+#
+# The check never blocks deployment or throws: any failure (offline,
+# air-gapped repositories, timeout, etc.) is swallowed and simply results in
+# no notification. The result (including a failed/unknown result) is cached
+# to disk for rsconnect.update_check_interval seconds (default 1 day), so
+# the network is only actually touched once per interval.
+
+# Fetches and parses a single repo's PACKAGES index, returning the version
+# listed for `pkg`, or NULL if it can't be determined. Uses httr2 rather
+# than available.packages()/download.file() because httr2::req_timeout()
```
aronatkins, line 19: `available.packages` is generally loaded once and cached in a session and is often already needed in order to compute the manifest during `deployApp`. Pity that we can't constrain its duration, as it would make much of this code unnecessary.

R/version-check.R
```diff
@@ -0,0 +1,244 @@
...
+  paste0(
+    "A newer version of rsconnect (",
+    latest,
+    ") is available from your repositories. You have ",
+    installed,
+    "."
+  )
+}
+
+# A neutral bullet appended to HTTP error messages. Deliberately avoids
```
aronatkins, line 250: An additional option: Emit the current rsconnect version and latest version on every call to `rsconnect::deployApp()`. This is useful for troubleshooting when we capture deploy logs and probably desirable in addition to what is already here.

> This is rsconnect 1.10.1
> This is rsconnect 1.10.1 (1.11.0 is latest available)

R/version-check.R
```diff
@@ -0,0 +1,244 @@
...
+  interval <- getOption("rsconnect.update_check_interval", 60 * 60 * 24)
+  fresh <- !is.null(cached) &&
+    identical(cached$repos, reposKey) &&
+    (as.numeric(Sys.time()) - cached$checked) < interval
+
+  if (fresh) {
+    latest <- cached$latest
+  } else {
+    latest <- latestAvailableVersion(pkg, repos)
+    writeVersionCheckCache(pkg, latest, reposKey, cacheDir)
```
aronatkins, line 194: Confirmed that `...../R/rsconnect/cache/rsconnect-update-check.dcf` exists with contents:

```dcf
Checked: 1784222580.96541
Latest: 1.10.1
Repos: b94ec174
```

If I edit the "Latest" version, restart my session, and load rsconnect, I see:

```
A newer version of rsconnect (1.12.0) is available from your repositories. You have 1.10.1.9000.
```

aronatkins (approved):
The changes look good. Would this be surprising to users? Do other R packages perform similar up-to-date checks?

karawoo:
> Would this be surprising to users? Do other R packages perform similar up-to-date checks?

It's not something I've seen in other R packages, but I _hope_ it's not too surprising to users. Startup messages generally are fairly common, ggplot2 uses them to display random ggplot2 tips. And users always have the option to disable the check & message.

R/cookies.R
```diff
@@ -71,12 +71,12 @@ storeCookies <- function(requestURL, cookieHeaders) {
   lapply(cookies, function(co) {
     # Remove any duplicates
     # RFC says duplicate cookies are ones that have the same domain, name, and path
-    hostCookies <<- hostCookies[
+    hostCookies <<- hostCookies[ # nolint
```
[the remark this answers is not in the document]
karawoo: making this suggested change will just bring back the linting failures 😭

PR #1353, opened 2026-07-28
Migrate rsconnect deployment records from shinyapps.io to Connect Cloud

vrsarah:
Adds a way to migrate an app's local deployment record from shinyapps.io to Posit Connect Cloud, so that the next `deployApp()` call — including the RStudio IDE's Publish button — routes to the already-migrated Connect Cloud content instead of the original server. New functions added:

- **migrateToConnectCloud(appPath, contentId, cloudAccount, appName, account, server)** - resolves and validates the local source deployment record first, then calls the Connect Cloud API to confirm the target content exists, writes a new `.dcf` record, and removes the superseded source record
- **ensureConnectCloudAccount()** - an internal guard called at the start of `migrateToConnectCloud()`. If no Connect Cloud account is registered, it offers a `cli_menu()` choice between the browser device-auth flow or pointing the user at `connectCloudClientCredentials()` for CI/non-interactive use.

Closes rstudio/lucid-server#5700

### Manual Testing

- Registered a live shinyapps.io account and deployed a Shiny app to it.
- Deployed the same app to Connect Cloud to create a real migration target, then used `forgetDeployment()` to reset to shinyapps.io as the only active record (simulating the pre-migration state).
- Ran `migrateToConnectCloud()` against the live shinyapps.io record with the real Connect Cloud content ID — succeeded, and `deployments()` afterward confirmed exactly one record remained, correctly pointing at `connect.posit.cloud` with the right appId, and the shinyapps.io record gone.


R/deployments.R
```diff
@@ -322,3 +322,141 @@ forgetDeployment <- function(
...
+  newRecord <- deploymentRecord(
+    name = sourceRecord$name,
+    title = content$title %||% sourceRecord$title,
+    username = ccInfo$name,
+    account = ccInfo$name,
+    server = "connect.posit.cloud",
+    hostUrl = serverInfo("connect.posit.cloud")$url,
+    appId = contentId,
+    bundleId = NULL, # Connect Cloud does not use bundle IDs
+    url = content$url %||% "",
```
karawoo: This doesn't seem to be putting a correct url into the deployment record. When I deploy after migrating I see:

Image

When I select `1` _then_ it records the correct URL in the record and subsequent deploy prompts show it. But I think we should be able to generate the right URL during the migration.  `awaitCompletion` does it like so:

```r
url = paste0(connectCloudUrls()$ui, "/", ccInfo$name, "/content/", contentId)
```

though `ccInfo$name` is the local account name, so the link would be wrong for a collaborator migrating content owned by a different account.

R/deployments.R
```diff
@@ -322,3 +322,141 @@ forgetDeployment <- function(
...
+        i = "Call {.fun rsconnect::connectCloudUser} or {.fun rsconnect::connectCloudClientCredentials} first."
+      )
+    )
+  }
+
+  idx <- cli_menu(
+    "No Posit Connect Cloud account registered.",
+    "Choose a setup method:",
+    choices = c(
+      "Browser (device auth): {.fun connectCloudUser}",
```
samperman: I think we should tell the user which path is recommended. (Does saying "device auth" add anything? Most users won't care what that is.)


```suggestion
      "Browser (Recommended): {.fun connectCloudUser}",
```

R/deployments.R
```diff
@@ -322,3 +322,141 @@ forgetDeployment <- function(
...
+    url = content$url %||% "",
+    envVars = sourceRecord$envVars[[1L]]
+  )
+  newPath <- deploymentConfigFile(
+    appPath,
+    newRecord$name,
+    newRecord$account,
+    newRecord$server
+  )
+  writeDeploymentRecord(newRecord, newPath)
```
karawoo: What if a deployment record already exists for this account/name? I think we probably should prompt the user rather than overwriting it.

tests/connectcloud-integration/README.md
karawoo, line 1: Setting up integration testing with connect cloud seems like a good idea but if these tests aren't set up to actually run then I'd lean toward removing them for now.

R/deployments.R
```diff
@@ -322,3 +322,141 @@ forgetDeployment <- function(
...
+#'   `https://connect.posit.cloud/{account}/content/{contentId}`.
+#' @param cloudAccount Local name of the Connect Cloud account to write the
+#'   new record under. When `NULL` and exactly one Connect Cloud account is
+#'   registered, it is used automatically.
+#' @param name,account,server Filters to disambiguate the source deployment
+#'   record when `appPath` has records for multiple deployments.
+#'
+#' @return The path to the new deployment record file, invisibly.
+#' @export
+migrateDeployment <- function(
```
karawoo: `migrateDeployment` seems too generic of a name given that it only supports migrating a record to Connect Cloud specifically. Though I'm not sure what to suggest instead `migrateDeploymentToConnectCloud` is way too long.

vrsarah: yeah I agree 🤔 I'll rename it to `migrateToConnectCloud`

R/deployments.R
vrsarah: Good idea. It now checks for an existing record at the target path before writing and prompts the user to overwrite or abort

R/deployments.R
vrsarah: Thanks for catching this! I resolved the account name against `client$getAccounts()$data` so that it gets the actual owning account's slug when building the URL

R/deployments.R
vrsarah: updated

R/deployments.R
```diff
@@ -322,3 +322,181 @@ forgetDeployment <- function(
...
+  # accounts the caller has a role on, since the content may belong to a
+  # different account (e.g. a team account) than the one authenticating
+  # this request -- using `ccInfo$name` unconditionally would build a URL
+  # under the wrong account for that case.
+  ownerAccount <- Find(
+    function(a) identical(a$id, content$account_id),
+    client$getAccounts()$data
+  )
+  contentUrl <- if (is.null(ownerAccount)) {
+    ""
```
karawoo: This suggests that the user doesn't have access to the account the content is in, so the next deploy would likely fail. Possibly better to fail early here rather than fall back to an empty URL?

R/deployments.R
```diff
@@ -322,3 +322,181 @@ forgetDeployment <- function(
...
+  # The content response carries only the owning account's id
+  # (`account_id`), not a ready-made URL or slug. Resolve it against the
+  # accounts the caller has a role on, since the content may belong to a
+  # different account (e.g. a team account) than the one authenticating
+  # this request -- using `ccInfo$name` unconditionally would build a URL
+  # under the wrong account for that case.
+  ownerAccount <- Find(
+    function(a) identical(a$id, content$account_id),
+    client$getAccounts()$data
+  )
```
karawoo, line 428: Would you mind extracting this out into a helper that `awaitCompletion` can use as well (instead of how it currently constructs the URL based on the current user)? Otherwise the first deploy after migrating will overwrite the URL to an incorrect one if the content belongs to a different account. This is really a preexisting bug in `awaitCompletion` but I think it makes sense to fix here.

vrsarah: ah yeah, that makes sense! Refactored it

R/deployApp.R
```diff
@@ -754,9 +754,18 @@ deployApp <- function(
   }
   if (!quiet) {
     if (deploymentSucceeded) {
-      cli::cli_alert_success(
-        "Successfully deployed to {.url {application$url}}"
-      )
+      if (isTRUE(nzchar(application$url))) {
```
vrsarah, line 757: This is to distinguish "succeeded with URL" from "succeeded, no URL".

R/deployments.R
```diff
@@ -322,3 +322,176 @@ forgetDeployment <- function(
...
+#' @param name,account,server Filters to disambiguate the source deployment
+#'   record when `appPath` has records for multiple deployments.
+#'
+#' @return The path to the new deployment record file, invisibly.
+#' @export
+migrateToConnectCloud <- function(
+  appPath = ".",
+  contentId,
+  cloudAccount = NULL,
+  name = NULL,
```
karawoo: could we call this `appName` instead of `name`? that would be consistent with other functions, and `disambiguateDeployments` at line 378 prompts the user to provide `appName`, not `name`.

R/client-connectCloud.R
```diff
@@ -92,6 +114,45 @@ connectCloudClient <- function(service, authInfo) {
...
+      path <- paste0(
+        "/accounts?has_user_role=true&include_total=true&limit=",
+        pageSize,
+        "&offset=",
+        offset
+      )
+      response <- withTokenRefreshRetry(GET, path)
+      allAccounts <- c(allAccounts, response$data)
+      offset <- offset + length(response$data)
+      if (length(response$data) == 0 || offset >= response$total) {
```
karawoo:
```suggestion
      if (length(response$data) == 0 || isTRUE(offset >= as.numeric(response$total))) {
```

just a little extra defensiveness if `response$total` is ever `NULL`/missing/character

karawoo (approved):
two very minor comments but once those are addressed this looks good to me to merge

vrsarah:
Thanks for the thorough review @karawoo 🙏

PR #1358, opened 2026-08-10
Reconstruct migration record from contentId when no local record exists

vrsarah:
Follow-up to #1353. Closes https://github.com/rstudio/lucid-server/issues/5706

For users whose local record is gone, the merged version had no recovery path - this closes that gap without requiring the shinyapps source bundle or `appId`. When no source record is found:

- The Connect Cloud record is reconstructed from `contentId` alone (which `getContent()` already verifies exists).
- The record `name` comes from `appName`, or is derived from the content's title via `generateAppName()` when `appName` is omitted - Connect Cloud content has no server-side name to reuse, only a title.
- The source-removal step is skipped (there is nothing to remove).

### Testing
1. Run `migrateToConnectCloud(appDir, contentId)` from a working directory where there is no deployment record.
2. It should create a Connect Cloud record from `contentId`.
3. `deployApp(appDir)` should send deployments to connect.posit.cloud.

### DIFF
--- a/R/deployments.R
+++ b/R/deployments.R
@@ -362,6 +362,7 @@ migrateToConnectCloud <- function(
   server = NULL
 ) {
   check_string(contentId)
+  check_string(appName, allow_empty = FALSE, allow_null = TRUE)
 
   # Ensure a Connect Cloud account exists; prompt to register one if not.
   ensureConnectCloudAccount()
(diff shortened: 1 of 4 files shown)


R/deployments.R
```diff
@@ -438,13 +460,13 @@ migrateToConnectCloud <- function(
...
-    envVars = sourceRecord$envVars[[1L]]
+    envVars = if (!is.null(sourceRecord)) sourceRecord$envVars[[1L]] else NULL
   )
 
   writeDeploymentRecord(newRecord, newPath)
 
-  # Remove the superseded source record.
-  if (unlink(sourceRecord$deploymentFile) != 0L) {
+  # Remove the superseded source record, when there was one.
+  if (!is.null(sourceRecord) && unlink(sourceRecord$deploymentFile) != 0L) {
```
samperman, line 488: If i'm reading this right, if the passed in `appName` doesn't match the existing record, then `sourceRecord` is null so nothing is removed. this would result in two records:
```
rsconnect/connect.posit.cloud/cc-account/{appName}.dcf
rsconnect/shinyapps.io/u/{someOtherOldName}.dcf
```

maybe we should query unfiltered first and only reconstruct if there really is no existing record. otherwise let disambiguateDeployments complain 🤔

R/deployments.R
```diff
@@ -359,55 +367,44 @@ migrateToConnectCloud <- function(
...
-        "Abort"
-      ),
-      not_interactive = c(
-        i = "Remove the existing record, or pass a different {.arg appName}/{.arg cloudAccount}, then retry."
-      )
+  # The record name comes from the source record when we have one, otherwise
+  # from `appName`. Both are known without a network call, so we can check for
+  # a target-path collision before doing any Connect Cloud API work. Without
+  # either, the name is derived from the content title after it is fetched.
+  recordName <- if (!is.null(sourceRecord)) sourceRecord$name else appName
```
samperman, line 417: what if `appName` is blank? should probably validate it (like we do `contentId`)

tests/testthat/test-migrateToConnectCloud.R
```diff
@@ -385,14 +387,255 @@ test_that("migrateToConnectCloud() aborts with no deployment records", {
...
+  expect_equal(rec$appId, "abc123")
+  expect_equal(rec$name, "myapp")
+  expect_equal(rec$title, "My App")
+  expect_equal(rec$url, "https://connect.posit.cloud/cc-account/content/abc123")
+})
+
+test_that("migrateToConnectCloud() derives the record name from the content title when no source and no appName", {
+  appDir <- withr::local_tempdir()
+
+  local_mocked_bindings(
```
samperman: lots of similar looking duplication for these local_mocked_bindings calls.. a helper method something like this would make the tests more readable:
```
local_migrate_mocks(title = "My App", getContent = NULL)
```

R/deployments.R
vrsarah: yeah, updated it to validate appName and error if it's blank

tests/testthat/test-migrateToConnectCloud.R
vrsarah: good idea, done!

R/deployments.R
vrsarah: that's true.. that would've left two records behind. Fixed it so `appName` isn't used as a filter anymore. Now if there are existing records but none matches the `appName`, it errors instead of reconstructing

samperman (approved):

karawoo (approved):

PR #1359, opened 2026-08-14
Add Posit Connect Cloud support to applications() and collaborator management functions

vrsarah:
Adds Posit Connect Cloud (PCC) support to `applications()` and to the five collaborator-management functions (`addAuthorizedUser`, `removeAuthorizedUser`, `showUsers`, `showInvited`, `resendInvitation`), and soft-deprecates those five functions.

`applications()` previously aborted for PCC accounts; it now lists PCC content via a paginated content request and returns the same column shape as the other server types. The five collaborator functions gain PCC code paths for inviting, listing, and removing members and pending invitations. On PCC, these functions identify their target content by the local deployment record's content id rather than by application name.

The five collaborator functions are soft-deprecated with `lifecycle::deprecate_warn()`, firing once per call on both shinyapps.io and PCC.

### Notes for reviewers

- The deprecation message wording is a placeholder and will be finalized before release; the `when` version is the current development version and should be bumped to the release version before CRAN submission.
- On PCC, `sendEmail = FALSE` is not honored (the platform always emails invitees) and `showInvited()`/`resendInvitation()` cannot surface an invitation link/code.
- `showUsers()` and `showInvited()` now build their data frames with atomic columns on all backends, so ids are returned as character and empty results are typed zero-row data frames (a visible change for existing shinyapps.io callers, noted in NEWS).

### DIFF
--- a/R/applications.R
+++ b/R/applications.R
@@ -93,8 +93,8 @@ applications <- function(account = NULL, server = NULL) {
     )
     res <- lapply(apps, function(x) {
       # On PCC the content list response does not expose a separately served-app
-      # URL; the browsable content URL is the best available and is used for both
-      # url and config_url.
+      # URL. url = the browsable content view page; config_url = the settings
+      # page for that content item.
       contentUrl <- paste0(contentUrlBase, x$id %||% "")
       data.frame(
         id = x$id %||% NA_character_,
@@ -104,7 +104,7 @@ applications <- function(account = NULL, server = NULL) {
         status = NA_character_,
         size = NA_character_,
         instances = NA_integer_,
-        config_url = contentUrl,
+        config_url = paste0(contentUrl, "/settings/info"),
         created_time = x$created_time %||% NA_character_,
         updated_time = x$updated_time %||% NA_character_,
         guid = NA_character_,
--- a/R/applications.R
+++ b/R/applications.R
@@ -59,21 +59,36 @@ applications <- function(account = NULL, server = NULL) {
       stringsAsFactors = FALSE
     )
     if (length(apps) == 0) return(empty)
+    # Resolve the owning account's real server-side slug once. All items belong
+    # to accountDetails$accountId (listApplications filters by it), so one
+    # getAccounts() call covers every row. Using the resolved slug rather than
+    # accountDetails$name (the local alias) prevents wrong-account URLs when the
+    # remote slug differs from the alias stored in the local config.
+    pccAccts <- client$getAccounts()$data
+    pccOwner <- Find(function(a) identical(a$id, accountDetails$accountId), pccAccts)
+    if (is.null(pccOwner)) {
+      cli::cli_abort(
+        c(
+          "Unable to determine the Connect Cloud account for content listing.",
+          i = "You may not have access to the account this content belongs to."
+        )
+      )
+    }
+    contentUrlBase <- paste0(connectCloudUrls()$ui, "/", pccOwner$name, "/content/")
     res <- lapply(apps, function(x) {
+      # On PCC the content list response does not expose a separately served-app
+      # URL; the browsable content URL is the best available and is used for both
+      # url and config_url.
+      contentUrl <- paste0(contentUrlBase, x$id %||% "")
       data.frame(
         id = x$id %||% NA_character_,
         name = x$title %||% NA_character_,
         title = x$title %||% NA_character_,
-        url = x$current_revision$url %||% NA_character_,
+        url = contentUrl,
         status = NA_character_,
         size = NA_character_,
         instances = NA_integer_,
-        # accountDetails$name is the authenticated caller's account, which matches
-        # the account_id filter passed to listApplications — content listed here
-        # always belongs to this account, so this URL is correct.
-        config_url = paste0(
-          connectCloudUrls()$ui, "/", accountDetails$name, "/content/", x$id %||% ""
-        ),
+        config_url = contentUrl,
         created_time = x$created_time %||% NA_character_,
         updated_time = x$updated_time %||% NA_character_,
         guid = NA_character_,
--- a/R/applications.R
+++ b/R/applications.R
@@ -327,6 +327,7 @@ getLogs <- function(
     server = server,
     account = account
   )
+  checkShinyappsServer(deployment$server)
   accountDetails <- accountInfo(deployment$account, deployment$server)
   client <- clientForAccount(accountDetails)
   application <- getAppByName(client, accountDetails, deployment$name)
--- a/R/applications.R
+++ b/R/applications.R
@@ -51,21 +51,32 @@ applications <- function(account = NULL, server = NULL) {
 
   if (isPCC) {
     empty <- data.frame(
-      id = character(), name = character(), title = character(),
-      url = character(), status = character(), size = character(),
-      instances = integer(), config_url = character(),
-      created_time = character(), updated_time = character(),
+      id = character(),
+      name = character(),
+      title = character(),
+      url = character(),
+      status = character(),
+      size = character(),
+      instances = integer(),
+      config_url = character(),
+      created_time = character(),
+      updated_time = character(),
       guid = character(),
       stringsAsFactors = FALSE
     )
-    if (length(apps) == 0) return(empty)
+    if (length(apps) == 0) {
+      return(empty)
+    }
     # Resolve the owning account's real server-side slug once. All items belong
     # to accountDetails$accountId (listApplications filters by it), so one
     # getAccounts() call covers every row. Using the resolved slug rather than
     # accountDetails$name (the local alias) prevents wrong-account URLs when the
     # remote slug differs from the alias stored in the local config.
     pccAccts <- client$getAccounts()$data
-    pccOwner <- Find(function(a) identical(a$id, accountDetails$accountId), pccAccts)
+    pccOwner <- Find(
+      function(a) identical(a$id, accountDetails$accountId),
+      pccAccts
+    )
     if (is.null(pccOwner)) {
       cli::cli_abort(
         c(
@@ -74,7 +85,12 @@ applications <- function(account = NULL, server = NULL) {
         )
       )
     }
-    contentUrlBase <- paste0(connectCloudUrls()$ui, "/", pccOwner$name, "/content/")
+    contentUrlBase <- paste0(
+      connectCloudUrls()$ui,
+      "/",
+      pccOwner$name,
+      "/content/"
+    )
     res <- lapply(apps, function(x) {
       # On PCC the content list response does not expose a separately served-app
       # URL; the browsable content URL is the best available and is used for both
--- a/R/applications.R
+++ b/R/applications.R
@@ -92,19 +92,22 @@ applications <- function(account = NULL, server = NULL) {
       "/content/"
     )
     res <- lapply(apps, function(x) {
-      # On PCC the content list response does not expose a separately served-app
-      # URL. url = the browsable content view page; config_url = the settings
-      # page for that content item.
-      contentUrl <- paste0(contentUrlBase, x$id %||% "")
+      # url = the standalone served URL handed to app consumers (consistent with
+      # shinyapps.io/Connect); config_url = the dashboard settings page.
+      # Prefer the revision's served URL (the vanity/custom URL when set); fall
+      # back to the constructed content-id URL for content not yet published,
+      # where current_revision (or its url) is absent.
+      contentId <- x$id %||% ""
+      dashboardUrl <- paste0(contentUrlBase, contentId)
       data.frame(
         id = x$id %||% NA_character_,
         name = x$title %||% NA_character_,
         title = x$title %||% NA_character_,
-        url = contentUrl,
+        url = x$current_revision$url %||% connectCloudStandaloneUrl(contentId),
         status = NA_character_,
         size = NA_character_,
         instances = NA_integer_,
-        config_url = paste0(contentUrl, "/settings/info"),
+        config_url = paste0(dashboardUrl, "/settings/info"),
         created_time = x$created_time %||% NA_character_,
         updated_time = x$updated_time %||% NA_character_,
         guid = NA_character_,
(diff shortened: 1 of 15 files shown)


man/addAuthorizedUser.Rd
```diff
@@ -36,12 +36,22 @@ custom message to send in email invitation. Defaults to NULL, which
...
-This function works only for ShinyApps servers.
+This function works for ShinyApps and Posit Connect Cloud. On Posit
+Connect Cloud, the content's account must be an organization account.
+The \code{sendEmail} argument is ignored on Posit Connect Cloud;
+PCC always sends an invitation email.
+
+On Posit Connect Cloud, the content is resolved from the local deployment
+record in \code{appDir}; the call must run from the directory that contains
+the \code{rsconnect/} deployment record. \code{appName} selects among
+multiple records in the same directory.
```
vrsarah: We could make it accept `contentId` so that a deployment record is not required, but I don't think we would want to change the API surface and introduce a new parameter ~for a feature that's going to be deprecated immediately~. Same for the other user management functions.

man/showUsers.Rd
```diff
@@ -18,13 +18,28 @@ there are multiple options, you'll be prompted to pick one.
...
 Use \code{\link[=accounts]{accounts()}} to see the full list of available options.}
 }
+\value{
+A data frame with one row per authorized user. Columns always
+present: \code{id}, \code{email}, \code{account}. The \code{account}
+column is populated on shinyapps.io only and is \code{NA} on Posit Connect
+Cloud. On Posit Connect Cloud the data frame additionally includes
+\code{display_name} and \code{role} (e.g. \code{"viewer"} or
+\code{"collaborator"}) from the API response.
+}
```
vrsarah, line 40: We could get the account name for the associated PCC user, but that adds an extra API call.

R/applications.R
```diff
@@ -298,6 +358,7 @@ getLogs <- function(
     server = server,
     account = account
   )
+  checkShinyappsServer(deployment$server)
```
vrsarah, line 364: getLogs() is a shinyapps.io-only API, so this guard was necessary when adding PCC support to applications()

R/client-connectCloud.R
```diff
@@ -416,6 +456,89 @@ connectCloudClient <- function(service, authInfo) {
...
+
+    listApplicationAuthorization = function(appId) {
+      pageSize <- 100
+      offset <- 0
+      allItems <- list()
+      repeat {
+        path <- paste0(
+          "/contents/",
+          appId,
+          "/users",
```
vrsarah: This might look confusing, but `/contents/{id}/users` is the endpoint that returns the content's authorized users on PCC (for shinyapps.io it's `GET /applications/{id}/authorization`)

R/auth.R
```diff
@@ -59,16 +205,31 @@ addAuthorizedUser <- function(
   emailMessage = NULL
 ) {
   accountDetails <- accountInfo(account, server)
-  checkShinyappsServer(accountDetails$server)
+  if (!isPositConnectCloudServer(accountDetails$server)) {
+    checkShinyappsServer(accountDetails$server)
+  }
+  lifecycle::deprecate_warn(
+    "1.11.0",
```
vrsarah: This is a guess, but we'll likely bump from 1.10.1 since we're making API changes.

R/applications.R
```diff
@@ -43,17 +43,77 @@ applications <- function(account = NULL, server = NULL) {
...
+      connectCloudUrls()$ui,
+      "/",
+      pccOwner$name,
+      "/content/"
+    )
+    res <- lapply(apps, function(x) {
+      # On PCC the content list response does not expose a separately served-app
+      # URL. url = the browsable content view page; config_url = the settings
+      # page for that content item.
+      contentUrl <- paste0(contentUrlBase, x$id %||% "")
```
samperman: Hmm. this is tricky. As a user, i think i would rather see the standalone url here (we already have `config_url` below. The standalone is the url that could be handed to an app consumer and is also more consistent with shinyapps.io and Posit Connect AFAICT

Two complications:
1. i don't know if the standalone url is actually available in this PCC API response. might be worth revisiting if not
2. the rest of the connect cloud code in `rsconnect` is making a similar decision. if we go down this route we might consider changing `connectCloudContentUrl()` `awaitCompletion()`, `deployApp()`, and `migrateToConnectCloud()`

Maybe that means its too late since it would change behavior of existing methods, but I think its worth a conversation

samperman: This is what each backend does now on this branch:

| Backend | `url` | `config_url` |
| --- | --- | --- |
| **Connect** | verbatim from the API — served content URL (vanity URL if set) | built client-side: `<server>/connect/#/apps/<id>` |
| **shinyapps.io** | verbatim from the API — `https://<account>.shinyapps.io/<app>/` | hardcoded: `https://www.shinyapps.io/admin/#/application/<id>` |
| **Connect Cloud** | built client-side: `<ui>/<slug>/content/<id>` (product UI page) | that same URL + `/settings/info` |

vrsarah: I think we should keep the product UI page URL (https://connect.posit.cloud/{account}/content/{content_id}) for `applications()$url`, since that's what the existing functions return and I'd want to it to match `deployments()$url` for the same content. We're also leaning towards that same product UI URL in the other Slack thread. If we later want the standalone URL, we can change it everywhere at once

samperman: he other slack thread was specifically what gets returned on a deploy... i think this is orthogonal to that discussion. My original thinking those should all match too, but i've backed off on that. There are separate fields on the application object here and if we choose to only show the dashboard url after deployments that seems ok. Is there a programmatic way to get the standalone url via this library? It feels like `applications()` should tell you that info

samperman: both Connect and Shinyapps.io return the standalone url for `application.url`

R/client-connectCloud.R
```diff
@@ -168,8 +168,48 @@ connectCloudClient <- function(service, authInfo) {
     withTokenRefreshRetry = withTokenRefreshRetry,
 
     listApplications = function(accountId, filters = list()) {
-      # TODO: call the real API when available (api doesn't support filtering by name yet)
-      return(list())
+      pageSize <- 100
+      offset <- 0
+      allItems <- list()
+      repeat {
+        path <- paste0(
```
samperman: should this filter out deleted content?

R/client-connectCloud.R
```diff
@@ -168,8 +168,48 @@ connectCloudClient <- function(service, authInfo) {
...
+          "/contents?account_id=",
+          accountId,
+          "&include_total=true&limit=",
+          pageSize,
+          "&offset=",
```
samperman, line 200: is offset reliable if there is no order provided?

R/applications.R
vrsarah: ah, I meant the other PCC functions return the UI URL (the ones you mentioned above)

R/auth.R
```diff
@@ -107,38 +275,49 @@ removeAuthorizedUser <- function(
...
+  # note — the root issue is a correctness bug, not just warning noise.
+  api <- clientForAccount(accountDetails)
+  users <- showUsers_impl(api, application$id)
+
+  user <- as.character(user)
+  # Match id first (UUID strings on PCC, numeric-as-character on shinyapps.io),
+  # then fall back to email. The old is.numeric() branch missed PCC UUID ids.
+  if (user %in% users$id) {
+    user <- users[users$id == user, ]
+  } else if (user %in% users$email) {
```
samperman, line 325: there is an edge case here... the connect cloud api might set the email to `REDACTED` ... if so, this code will report the user as "not found" which isn't quite accurate. 🤷

R/applications.R
samperman: i know. i guess my point is I think _this one_ should set `application.url` to the standalone url. 🤷

vrsarah: I focused too much on making the content url consistent across PCC functions 😅 you're right, we don't have to keep the applications URL consistent with deployments. We can get the standalone URL from `current_revision.url` (covers vanity/custom domains) - it's included in the list response, so no extra lookup needed - and fall back to the constructed URL

R/client-connectCloud.R
vrsarah: actually.. nope, this is wrong

R/auth.R
vrsarah: good point, I'll improve the message!

R/auth.R
```diff
@@ -107,38 +275,59 @@ removeAuthorizedUser <- function(
...
+    user <- users[users$email == user, ]
```
karawoo:
```suggestion
    user <- users[which(users$id == user), ]
  } else if (user %in% users$email) {
    user <- users[which(users$email == user), ]
```

If an email is redacted then `users$email == user` will produce NAs. Those NAs when treated as indices in this subset will produce all-NA rows that will cause an error later when we do `if (is.na(user$id))` (there will be more than one value in `user$id`, which results in a `the condition has length > 1` error).

I think this suggested edit should fix it but having a test case too would probably be worth it

R/client-connectCloud.R
```diff
@@ -168,8 +179,56 @@ connectCloudClient <- function(service, authInfo) {
     withTokenRefreshRetry = withTokenRefreshRetry,
 
     listApplications = function(accountId, filters = list()) {
-      # TODO: call the real API when available (api doesn't support filtering by name yet)
-      return(list())
+      pageSize <- 100
+      offset <- 0
+      allItems <- list()
+      repeat {
```
karawoo: this pagination pattern appears three times in this PR plus there's one existing instance in `getAccounts`, probably worth writing a helper

R/auth.R
```diff
@@ -149,16 +338,29 @@ removeAuthorizedUser <- function(
...
+#'   present: \code{id}, \code{email}, \code{account}. The \code{account}
+#'   column is populated on shinyapps.io only and is \code{NA} on Posit Connect
+#'   Cloud. On Posit Connect Cloud the data frame additionally includes
+#'   \code{display_name} and \code{role} (e.g. \code{"viewer"} or
+#'   \code{"collaborator"}) from the API response.
+#' @note This function works for ShinyApps and Posit Connect Cloud.
+#'
+#'   On Posit Connect Cloud, the content is resolved from the local deployment
+#'   record in \code{appDir}; the call must run from the directory that contains
+#'   the \code{rsconnect/} deployment record. \code{appName} selects among
```
karawoo: > the call must run from the directory that contains the \code{rsconnect/} deployment record

I don't think this is true, you can run it from a different directory as long as you pass in `appDir` right?

R/auth.R
```diff
@@ -1,3 +1,108 @@
...
+# original three-column shape (id, email, account).
+showUsers_impl <- function(api, applicationId, isPCC = FALSE) {
+  res <- api$listApplicationAuthorization(applicationId)
+  rows <- lapply(res, function(x) {
+    id <- as.character(x$user$id %||% NA_character_)
+    email <- as.character(x$user$email %||% NA_character_)
+    if (is.na(id) && is.na(email)) {
+      cli::cli_abort(
+        c(
+          "Unexpected response from Connect Cloud: a user record has neither an {.field id} nor an {.field email}.",
```
karawoo: this error message mentions connect cloud even if we're running against shinyapps.io

karawoo (commented):
Could you say more about why you're adding connect cloud functionality that's immediately deprecated for the collaborator management functions?

R/auth.R
vrsarah: ah yes, this comment is wrong 😬

vrsarah:
> Could you say more about why you're adding connect cloud functionality that's immediately deprecated for the collaborator management functions?

My understanding is these are the shinyapps.io-era collaborator functions, and we want to steer people to the Connect Cloud UI rather than grow rsconnect's collaborator API. Thinking about it more though, I'm not sure the deprecation should cover PCC too 🤔 if PCC collaborator management is meant to be a supported rsconnect path, we'd want to gate the warning to shinyapps.io only. I'll re-confirm the intended scope with my team before finalizing.

R/auth.R
vrsarah: you're right, fixed 🙏

R/auth.R
vrsarah: oops, fixed 🥹

vrsarah:
@karawoo So it sounds like we actually don't want to deprecate until we add Connect Cloud support to the Connect API package.. I'll remove those flags from the user management functions. Thanks for asking!

R/auth.R
```diff
@@ -1,3 +1,92 @@
...
+        )
+      )
+    }
+    if (isPCC) {
+      data.frame(
+        id = id,
+        email = email,
+        account = NA_character_,
+        display_name = as.character(x$user$display_name %||% NA_character_),
+        role = as.character(x$role %||% NA_character_),
```
samperman, line 27: why is account NA but role might not be? doesn't the role apply to a specific account?

samperman (approved):

R/auth.R
```diff
@@ -107,38 +252,53 @@ removeAuthorizedUser <- function(
...
-      user <- users[users$id == user, ]
-    } else {
-      stop("User ", user, " not found", call. = FALSE)
-    }
+  # resolve content exactly once: use impl so showUsers() does not call
+  # resolveContentTarget() a second time (a second interactive prompt could
+  # return a different record, causing removeApplicationUser to act on the
+  # wrong content).
+  api <- clientForAccount(accountDetails)
+  users <- showUsers_impl(api, application$id)
```
karawoo:
```suggestion
  users <- showUsers_impl(
    api,
    application$id,
    isPCC = isPositConnectCloudServer(accountDetails$server)
  )
```

karawoo (approved):

R/auth.R
vrsarah: `account` is a shinyapps.io-only field (the collaborator's account name) that Connect Cloud user record doesn't include, so it's NA there and kept only for column consistency. `role` is a Connect Cloud-specific field the API returns - its presence isn't tied to account, and the user is identified by id/email/display_name on Connect Cloud.

vrsarah:
I added an optional `contentId` support in a separate PR for reviewer's convenience: https://github.com/rstudio/rsconnect/pull/1360

PR #1362, opened 2026-08-29
Fix unset Connect Cloud environment variables

### LINKED ISSUE #1361: Connect Cloud: Secrets for which no env var is set get overridden
### Description

When `deployApp()` updates existing Connect Cloud content, `client$updateContent()` builds the `secrets` payload like this:

https://github.com/rstudio/rsconnect/blob/main/R/client-connectCloud.R#L235-L244

```r
secrets <- unname(Map(
  function(name, value) {
    list(
      name = name,
      value = value
    )
  },
  envVars,
  Sys.getenv(envVars)
))
```

`Sys.getenv(envVars)` is called with no `unset =` argument, so if one of the env var names stored in the deployment record (`deployment$envVars`) is not set in the current R session/shell, it resolves to `""` — an actual empty string — and that empty string is sent to the API as the secret's `value`. This overwrites/clears the previously configured secret on Connect Cloud instead of leaving it unchanged.

Note that you don't need to pass `envVars` again to hit this: only the *names* of env vars are stored in the deployment record, not their values (see `updateDeployment()` in `R/deploymentTarget.R`, which falls back to `previous$envVars[[1]]` when `envVars` isn't supplied). So a name set once keeps getting resent on every subsequent `deployApp()` call, evaluated against whatever the variable's value happens to be *at that later deploy time* — even if the caller never mentions `envVars` again.

This is inconsistent with how the same situation is handled for plain Posit Connect in `client$setEnvVars()`:

https://github.com/rstudio/rsconnect/blob/main/R/client-connect.R#L147-L160

```r
setEnvVars = function(guid, vars) {
  path <- v1_url("content", guid, "environment")
  body <- unname(Map(
    function(name, value) {
      list(
        name = name,
        value = if (is.na(value)) NULL else value
      )
    },
    vars,
    Sys.getenv(vars, unset = NA)
  ))
  PATCH_JSON(service, authInfo, path, body)
}
```

Here `Sys.getenv(vars, unset = NA)` combined with `if (is.na(value)) NULL else value` means an unset var is sent with `value = NULL` instead of `""`. Note this still sends an explicit JSON `null` for the entry (via `toJSON(..., null = "null")`) rather than omitting the entry outright, so it's not clearly correct either — it's untested what the Connect API does with an explicit `null` value here.

### Steps to reproduce

1. `Sys.setenv(MY_SECRET = "hunter2")`
2. `rsconnect::deployApp(envVars = c("MY_SECRET"))` — deploys to Connect Cloud, storing the secret `MY_SECRET = "hunter2"` on the server, and storing the *name* `MY_SECRET` (not the value) in the local deployment record.
3. `Sys.unsetenv("MY_SECRET")` (or: run in a fresh shell/session where it was never set).
4. `rsconnect::deployApp()` — note `envVars` is **not** passed this time. The stored name is reused automatically from the deployment record.
5. The secret `MY_SECRET` on the server gets overwritten with an empty string, silently, with no indication to the caller that this happened.

### Expected behavior

When resolving a stored env var name's current value via `Sys.getenv()` at deploy time, if the variable is unset, that name should be **omitted entirely** from the API payload (no entry sent for it at all), not sent with `value = ""` (Connect Cloud) or `value = null` (Connect). This leaves the existing secret on the server untouched, matching the documented behavior that "removing an environment variable from `envVars` will leave it unchanged on the server."

### Your environment

_No response_

LeonidasZhak:
## Summary

Connect Cloud deployments omit environment variables that are unset in the current R session from the `secrets` payload. Explicitly set empty values remain sendable. This update also merges current `main` so the PR is no longer conflicted.

## Thanks to maintainers

Thanks to the rsconnect maintainers for maintaining the deployment clients and for the clear issue report.

## Issue or motivation

Fixes #1361. A deployment record stores environment variable names for later deployments. If one of those variables is absent later, the Connect Cloud client previously sent `value = ""`, clearing the existing secret.

## Root cause

`Sys.getenv(envVars)` uses an empty string for an unset variable, so the client could not distinguish an absent variable from one explicitly set to an empty value. During the conflict repair, `main` had also added guards for `envVars = NULL` and `character(0)`; the final helper preserves that empty-array behavior while filtering only absent variables.

## Change

Add a Connect Cloud-specific helper that returns an empty JSON array for no variables, uses `unset = NA` for named variables, and filters only absent variables before constructing the payload. Use it for both content creation and updates. Add a regression test that captures both request bodies and checks that an explicitly empty value is preserved while an unset variable is omitted. Merge current `main` and keep the existing Connect Cloud fixes.

## Tests

- `Rscript -e 'devtools::load_all(quiet = TRUE); testthat::test_file("tests/testthat/test-client-connectCloud.R", reporter = "summary")'`: passed.
- `air format --check R/client-connectCloud.R tests/testthat/test-client-connectCloud.R NEWS.md`: passed.
- `git diff --cached upstream/main --check`: passed before commit.
- `lintr::lint_package()`: passed, no lints.
- `R CMD build .`: passed, including vignette creation.
- `_R_CHECK_CRAN_INCOMING_REMOTE_=false R CMD check --as-cran --no-manual rsconnect_1.11.0.9000.tar.gz`: passed with 1 existing development-version NOTE.
- `_R_CHECK_FORCE_SUGGESTS_=false _R_CHECK_CRAN_INCOMING_REMOTE_=false R CMD check --no-manual rsconnect_1.11.0.9000.tar.gz`: passed with Status: OK.

## Scope

This does not change Connect Cloud authentication, secret names, explicitly empty values, deployment record storage, `deployApp(appId=)` behavior, or user-management endpoints. I did not run credentialed Connect/shinyapps.io integration workflows locally because they require repository secrets.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # rsconnect (development version)
 
+* Connect Cloud deployments no longer send empty values for environment
+  variables that are unset in the current R session. (#1361)
+
 # rsconnect 1.11.0
 
 * rsconnect checks whether a newer version of itself is available from your
--- a/R/client-connectCloud.R
+++ b/R/client-connectCloud.R
@@ -47,6 +47,22 @@ cloudContentTypeFromAppMode <- function(appMode) {
   )
 }
 
+cloudSecrets <- function(envVars) {
+  values <- Sys.getenv(envVars, unset = NA)
+  keep <- !is.na(values)
+
+  unname(Map(
+    function(name, value) {
+      list(
+        name = name,
+        value = value
+      )
+    },
+    envVars[keep],
+    values[keep]
+  ))
+}
+
 # Creates a client for interacting with the Connect Cloud API.
 connectCloudClient <- function(service, authInfo) {
   # Generic retry wrapper. If a request fails with 401 Unauthorized, it will
@@ -191,16 +207,7 @@ connectCloudClient <- function(service, authInfo) {
         primary_file = primaryFile
       )
 
-      secrets <- unname(Map(
-        function(name, value) {
-          list(
-            name = name,
-            value = value
-          )
-        },
-        envVars,
-        Sys.getenv(envVars)
-      ))
+      secrets <- cloudSecrets(envVars)
 
       json <- list(
         account_id = accountId,
@@ -232,16 +239,7 @@ connectCloudClient <- function(service, authInfo) {
         path <- paste0(path, "?new_bundle=true")
       }
 
-      secrets <- unname(Map(
-        function(name, value) {
-          list(
-            name = name,
-            value = value
-          )
-        },
-        envVars,
-        Sys.getenv(envVars)
-      ))
+      secrets <- cloudSecrets(envVars)
 
       json <- list(
         secrets = secrets,
--- a/tests/testthat/test-client-connectCloud.R
+++ b/tests/testthat/test-client-connectCloud.R
@@ -51,6 +51,62 @@ test_that("awaitCompletion", {
   expect_null(result$error)
 })
 
+test_that("Connect Cloud omits unset environment variables from secrets", {
+  local_mocked_bindings(
+    POST_JSON = function(service, authInfo, path, json) {
+      list(id = "content123", json = json)
+    },
+    PATCH_JSON = function(service, authInfo, path, json) {
+      list(id = "content123", json = json)
+    },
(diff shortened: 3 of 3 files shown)


joshyam-k (approved):
thank you for this @LeonidasZhak! this is looking good and I can shepherd it the rest of the way through.

PR #1371, opened 2026-09-04
document update-check options and generate options.Rd from roxygen

karawoo:
#1347 Introduced some new rsconnect options but didn't document them in options.Rd. I wasn't aware of that file and it wouldn't have occurred to me to edit a .Rd file directly.

I find editing roxygen2 a lot more pleasant so I moved the contents of options.Rd to options.R and rewrote the contents as roxygen2 comments. Also added documentation for the new options from #1347.

Left is the current appearance on main, right is how it looks with this change (the main difference is the options are a bulleted list now): 

Screenshot 2026-09-04 at 2 34 48 PM


jonkeane (approved):
Ooph, I didn't know about this separate non-roxygen options file either. Thanks for this!

PR #1373, opened 2026-09-09
fix(connect): insert dashboard_url into legacy API response

### LINKED ISSUE #1372: Redeploy opens content URL instead of the Connect access/settings page
*Summary*

After upgrading from rsconnect 1.7.0 to rsconnect 1.10.0, *redeploying* existing content opens the content URL (`https://<server>/content/<guid>`) in the browser instead of the Connect access page (`https://<server>/connect/#/apps/<guid>/access`). Initial deployments still open the access page.

Previously both initial deployments and redeployments opened the access page.

*Steps to reproduce*

1. Deploy content to Connect (e.g. from the RStudio/Workbench Publish button or `rsconnect::deployApp()`) → browser opens `.../connect/#/apps/<guid>/access` :white_check_mark:
2. Redeploy the same content → browser opens `.../content/<guid>` :x:

Reproduced with rsconnect 1.11.0 against Connect 2026.06.1, and originally reported with rsconnect 1.10.0 against an older Connect version — so it does not appear to be Connect-version dependent.

*Expected*

Redeployments open the access/settings page, as they did in 1.7.0.

*Possible cause*

`openURL()` (`R/deployApp.R:1062-1078`) opens `<dashboard_url>/access` when `application$dashboard_url` is set, and otherwise falls back to `application$url` (the content URL).

https://github.com/rstudio/rsconnect/pull/1280 ("use v1 APIs for create/upload/deploy") migrated the create path to Connect's v1 API and removed the `configureApplication()` call that previously supplied a config URL for both paths:

• Initial deploy: `createApplication()` uses v1 `/content` and returns `dashboard_url` (`R/client-connect.R:77-80`) → access page.
• Redeploy: `getApplication()` still uses the unversioned `/applications/{id}` endpoint, which does not return `dashboard_url` (`R/client-connect.R:104-109`) → `dashboard_url` is `NULL` → content URL.

nealrichardson:
Fixes #1372

Because saved deployment records only contain numeric content ids and not GUIDs, we have to use the unversioned `applications/:id/` API to look them up. That API does not include `dashboard_url` in its response, which meant that, when deploying content for the first time using the v1 APIs, `openURL()` (called on success in `deployApp()`) would open the `dashboard_url`, but when re-deploying and looking up from the deployment record, it would fall back to opening the standalone content URL in the browser. And unfortunately today, there is no way in Connect to get from the standalone URL back into the Connect dashboard if you want to view or edit any settings. #1280 introduced this discrepancy in refactoring some but not all of the API call sites to use v1 APIs.

The fix here just constructs the `dashboard_url` and injects it into the unversioned response in `getApplication()`. It turns out we already had the logic elsewhere, so I factored out a helper and used it in both places.


karawoo (approved):
this has always annoyed me, thank you

PR #1375, opened 2026-09-10
Fix deployApp(appId=) for Connect Cloud

### LINKED ISSUE #1367: deployApp(appId=) fails for Connect Cloud: the client implements no getApplication()
`deployApp(appId = ...)` cannot be used with Posit Connect Cloud. `findDeploymentTargetByAppId()` calls `getApplication()`, which calls `client$getApplication()`, and the Connect Cloud client does not define that method.

This matters for CI. On Connect Cloud, content cannot be found by name either (`applications()` aborts, and `listApplications()` maps `name` to the title), so the content id is the only stable identifier — and it is exactly the identifier `deployApp()` cannot accept.

## Reprex, no network

```r
info   <- rsconnect:::accountInfo("<your account>", "connect.posit.cloud")
client <- rsconnect:::clientForAccount(info)

"getApplication" %in% names(client)
#> FALSE

sort(names(client))
#> "awaitCompletion" "createContent" "currentUser" "getAccounts" "getAuthorization"
#> "getContent" "listApplications" "publish" "service" "updateContent"
#> "uploadBundle" "withTokenRefreshRetry"
```

`getContent()` is present and does the same job; `getApplication()` is the name the shared code path asks for.

## Reprex, against Connect Cloud

From a directory with no `rsconnect/` record, which is what a fresh CI checkout looks like:

```r

rsconnect::deployApp(
  appDir  = "<an app directory>",
  appId   = "<content id from the dashboard URL>",
  account = "<your account>",
  server  = "connect.posit.cloud"
)
```

Observed:

```
Error in client$getApplication(appId, "unknown") : attempt to apply non-function
```

Expected: the content that `appId` names receives the deployment.

## Cause

`R/applications.R`:

```r
getApplication <- function(account, server, appId) {
  ...
  client$getApplication(appId, "unknown"),
```

`R/client-connectCloud.R` returns a list of methods that includes `getContent` but not `getApplication`.

## Workaround

Write the deployment record from the content id first, then deploy by name. `migrateToConnectCloud()` (1.11.0) reconstructs the record when none exists, which makes this usable in CI:

```r
rsconnect::migrateToConnectCloud(
  appPath      = appDir,
  contentId    = "<content id>",
  cloudAccount = "<your account>",
  appName      = "<name>"
)
rsconnect::deployApp(appDir = appDir, appName = "<name>", forceUpdate = TRUE, ...)
```

## Suggestion

Aliasing `getApplication` to `getContent` in the Connect Cloud client would make `appId` work, since the shared code path only needs the content record back.

Worth noting alongside this: on `main`, the name-based rediscovery in `findDeploymentTargetByAppName()` is now skipped for Connect Cloud. With `appId` also unusable, code that has lost its local record has no way left to target existing content, and a deployment that used to update content will instead create a duplicate.

vrsarah:
`deployApp(appId = "<content-id>")` on Posit Connect Cloud failed with the error:
```
Error in client$getApplication(appId, "unknown") : attempt to apply non-function
``` 
because the Connect Cloud client implemented `getContent()` but not `getApplication()`. 

This PR adds `getApplication(applicationId, deploymentRecordVersion)` to the `connectCloudClient()` returned list. It delegates to the existing `getContent(applicationId)` and sets `content$name <- content$title` so that `createDeploymentFromApplication()` downstream gets a non-NULL `name` field (matching the convention already used in `listApplications`).

Fixes #1367

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # rsconnect (development version)
 
+* `deployApp(appId=)` now works for Posit Connect Cloud content. The Connect
+  Cloud client previously implemented no `getApplication()` method, so
+  deploying to an existing content item by id (rather than via a local
+  deployment record) errored with "attempt to apply non-function". (#1367)
+
 * `showUsers()` and `showInvited()` now always return a data frame with
   atomic `character`/`logical` columns, including a typed 0-row data frame
   (rather than `NULL`) when there are no results. User ids are now always
--- a/R/client-connectCloud.R
+++ b/R/client-connectCloud.R
@@ -275,6 +275,12 @@ connectCloudClient <- function(service, authInfo) {
 
     getContent = getContent,
 
+    getApplication = function(applicationId, deploymentRecordVersion) {
+      content <- getContent(applicationId)
+      content$name <- content$title
+      content
+    },
+
     updateContent = function(
       contentId,
       envVars,
--- a/tests/testthat/test-client-connectCloud.R
+++ b/tests/testthat/test-client-connectCloud.R
@@ -1042,3 +1042,38 @@ test_that("listApplicationInvitations accumulates multiple pages and keeps accep
   expect_equal(result[[2]]$id, "inv-2")
   expect_equal(result[[3]]$id, "inv-3")
 })
+
+test_that("getApplication() delegates to getContent and sets name = title", {
+  skip_if_not_installed("webfakes")
+
+  content_app <- webfakes::new_app()
+  content_app$use(webfakes::mw_json())
+  content_app$get("/contents/:id", function(req, res) {
+    res$set_status(200L)$send_json(
+      list(
+        id = "content-abc",
+        title = "My App Title",
+        state = "active",
+        account_id = "acct-1"
+      ),
+      auto_unbox = TRUE
+    )
+  })
+  app <- webfakes::local_app_process(content_app)
+  service <- parseHttpUrl(app$url())
+
+  authInfo <- list(
+    server = "connect.posit.cloud",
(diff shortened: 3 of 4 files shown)


R/client-connectCloud.R
```diff
@@ -275,6 +275,12 @@ connectCloudClient <- function(service, authInfo) {
 
     getContent = getContent,
 
+    getApplication = function(applicationId, deploymentRecordVersion) {
+      content <- getContent(applicationId)
+      content$name <- content$title
```
karawoo, line 280: I think this wants to be wrapped in `generateAppName`.

```suggestion
      content$name <- generateAppName(content$title, unique = FALSE)
```

vrsarah: ah, yes!

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # rsconnect (development version)
 
+* `deployApp(appId=)` now works for Posit Connect Cloud content. The Connect
+  Cloud client previously implemented no `getApplication()` method, so
+  deploying to an existing content item by id (rather than via a local
+  deployment record) errored with "attempt to apply non-function". (#1367)
```
karawoo: Could we remove the detail about `The Connect Cloud client previously implemented no `getApplication()` method`? that's an internal implementation detail that won't mean much to an end user.

karawoo (approved):
could we update the news to remove the internal implementation details? otherwise looks good to me

NEWS.md
vrsarah: oops 🤦‍♀️ will remove it!

PR #1376, opened 2026-09-10
Fix null primary_file for manifest deploys to Connect Cloud

### LINKED ISSUE #1366: deployApp(manifestPath=) sends primary_file: null to Connect Cloud, which rejects it
Deploying to Posit Connect Cloud from a pre-generated `manifest.json` always fails. The API rejects the request because `primary_file` is null, and it is null because supplying `appMode` from the manifest skips the inference that would have set it.

## Reprex, no network

`appMetadata()` sets `inferredPrimaryFile` only in the branch that infers `appMode`. `deployApp(manifestPath=)` takes `appMode` from the manifest, so that branch never runs:

```r
d <- tempfile(); dir.create(d)
writeLines("library(shiny); shinyApp(fluidPage(), function(input, output) {})", file.path(d, "app.R"))

inferred <- rsconnect:::appMetadata(appDir = d, appFiles = "app.R")
supplied <- rsconnect:::appMetadata(appDir = d, appFiles = "app.R", appMode = "shiny")

inferred$appPrimaryDoc        # NULL
inferred$inferredPrimaryFile  # "app.R"   <- the value the API needs

supplied$appPrimaryDoc        # NULL
supplied$inferredPrimaryFile  # NULL      <- nothing left to send
```

## Reprex, against Connect Cloud

```r
d <- tempfile(); dir.create(d)
writeLines("library(shiny); shinyApp(fluidPage(), function(input, output) {})", file.path(d, "app.R"))
rsconnect::writeManifest(appDir = d)

rsconnect::deployApp(
  appDir       = d,
  manifestPath = file.path(d, "manifest.json"),
  appName      = "reprex-primary-file",
  account      = "<your account>",
  server       = "connect.posit.cloud"
)
```

Observed:

```
Field `body.next_revision.primary_file` error: Input should be a valid string
```

Expected: the deployment succeeds, with `primary_file` taken from the manifest.

## Cause

`R/deployApp.R` builds the value as `appMetadata$appPrimaryDoc %||% appMetadata$inferredPrimaryFile`, in all three places that talk to Connect Cloud (content creation, `updateContent()`, and the later save). Both halves are NULL for a Shiny application deployed from a manifest: `appPrimaryDoc` comes from `manifest$metadata$primary_rmd`, which is null for Shiny, and `inferredPrimaryFile` is only ever assigned inside `if (is.null(appMode))` in `R/appMetadata.R`.

`inferredPrimaryFile` has no other assignment anywhere in the package, so a manifest deployment can never populate it.

## Workaround

Reproduce the rule that `inferAppMode()` uses, and inject it:

```r
ns <- asNamespace("rsconnect")
original <- get("appMetadata", envir = ns)
utils::assignInNamespace("appMetadata", function(...) {
  m <- original(...)
  if (is.null(m$appPrimaryDoc) && is.null(m$inferredPrimaryFile)) {
    m$inferredPrimaryFile <- "app.R"   # or "server.R", per inferAppMode()
  }
  m
}, ns = "rsconnect")
```

Deploying from a committed manifest is not optional in our case: some of our applications depend on packages that are not on CRAN, so a CI runner cannot resolve dependencies and regenerate a manifest.

## Suggestion

The manifest branch of `deployApp()` already reads `manifest$files`. The same rule `inferAppMode()` applies to a directory could be applied to that file list, which would make `primary_file` correct for both Shiny styles without touching the inference path.

vrsarah:
When deploying to Posit Connect Cloud from a pre-generated `manifest.json` via `deployApp(manifestPath=)`, the deploy always failed with:

```
Field `body.next_revision.primary_file` error: Input should be a valid string
```

**Root cause:** In `R/appMetadata.R`, `inferredPrimaryFile` is initialized to `NULL` and only assigned inside the `if (is.null(appMode))` block. When `appMode` is supplied by the manifest path, that block is skipped entirely, so `inferredPrimaryFile` stays `NULL`. Connect Cloud receives `null` for `primary_file` and rejects the request.

To fix the issue, this PR adds an `else` branch that calls `inferAppMode()` solely to obtain `primaryFile`, without touching the caller-supplied `appMode`. This is safe for all other deploy targets — they ignore `inferredPrimaryFile` — and costs one extra file scan only when `appMode` is pre-supplied.

### Before / After

```r
d <- tempfile(); dir.create(d)
writeLines("library(shiny); shinyApp(fluidPage(), function(input,output){})", file.path(d,"app.R"))

# Before fix
appMetadata(appDir=d, appFiles="app.R", appMode="shiny")$inferredPrimaryFile
#> NULL   ← Connect Cloud receives null → 422 error

# After fix
appMetadata(appDir=d, appFiles="app.R", appMode="shiny")$inferredPrimaryFile
#> [1] "app.R"
```

Fixes #1366

### DIFF
--- a/R/appMetadata.R
+++ b/R/appMetadata.R
@@ -40,6 +40,16 @@ appMetadata <- function(
       appMode <- appModeResult$appMode
       inferredPrimaryFile <- appModeResult$primaryFile
     }
+  } else {
+    # appMode was supplied (e.g. from a manifest), so the inference block above
+    # is skipped -- but Connect Cloud still needs a primary_file. Infer just the
+    # primary file from the same file-list rules, keeping the supplied appMode.
+    inferredPrimaryFile <- inferAppMode(
+      appDir,
+      appFiles,
+      usesQuarto = quarto,
+      isShinyappsServer = isShinyappsServer
+    )$primaryFile
   }
 
   appPrimaryDoc <- inferAppPrimaryDoc(
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # rsconnect (development version)
 
+* Deploying to Posit Connect Cloud from a pre-generated `manifest.json`
+  (`deployApp(manifestPath=)`) no longer fails with a null `primary_file`.
+  The primary file is now inferred from the file list even when `appMode` is
+  supplied. (#1366)
+
 * `showUsers()` and `showInvited()` now always return a data frame with
   atomic `character`/`logical` columns, including a typed 0-row data frame
   (rather than `NULL`) when there are no results. User ids are now always
--- a/tests/testthat/test-appMetadata.R
+++ b/tests/testthat/test-appMetadata.R
@@ -229,6 +229,38 @@ test_that("appMetadata returns NULL plumberInfo for nodejs", {
   expect_null(metadata$plumberInfo)
 })
 
+test_that("inferredPrimaryFile is populated when appMode is supplied (app.R)", {
+  dir <- withr::local_tempdir()
+  writeLines(
+    "library(shiny); shinyApp(fluidPage(), function(input, output) {})",
+    file.path(dir, "app.R")
+  )
+  metadata <- appMetadata(dir, appFiles = "app.R", appMode = "shiny")
+  expect_equal(metadata$inferredPrimaryFile, "app.R")
+})
+
+test_that("inferredPrimaryFile is populated when appMode is supplied (server.R)", {
+  dir <- withr::local_tempdir()
+  writeLines("", file.path(dir, "server.R"))
+  writeLines("", file.path(dir, "ui.R"))
+  metadata <- appMetadata(
+    dir,
+    appFiles = c("server.R", "ui.R"),
+    appMode = "shiny"
+  )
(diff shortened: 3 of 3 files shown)


R/appMetadata.R
```diff
@@ -40,6 +40,16 @@ appMetadata <- function(
       appMode <- appModeResult$appMode
       inferredPrimaryFile <- appModeResult$primaryFile
     }
+  } else {
+    # appMode was supplied (e.g. from a manifest), so the inference block above
+    # is skipped -- but Connect Cloud still needs a primary_file. Infer just the
+    # primary file from the same file-list rules, keeping the supplied appMode.
+    inferredPrimaryFile <- inferAppMode(
```
karawoo, line 47: what happens if the inferred app mode differs from the one that's passed in?

vrsarah: That's a good question 🤔 The primary file is paired with whatever mode we detect, so if the manifest's app mode doesn't match, we'd send a primary file that belongs to the other app mode. That doesn't necessarily make it the wrong file though, so publishing still works.

It only bites us when the file is genuinely wrong for the declared mode. In that case, the app publishes successfully but fails at runtime (connect-engine). I'll update the comment and include explanation about that case

karawoo (approved):

PR #1377, opened 2026-09-10
Fix Connect Cloud redeploy: stale upload token and upload=FALSE

### LINKED ISSUE #1369: deployApp(upload = FALSE) fails on Connect Cloud with object 'bundle' not found
`deployApp(upload = FALSE)` publishes the existing bundle and then errors while saving the deployment record, because `bundle` is only assigned for shinyapps.io in that branch.

## Reprex

```r

# a directory that already has a deployment record for Connect Cloud
rsconnect::deployApp(
  appDir  = "<an app directory>",
  appName = "<name>",
  account = "<your account>",
  server  = "connect.posit.cloud",
  upload  = FALSE
)
```

Observed — note that the publish itself starts, so the failure is after the work:

```
── Deploying to server ────────────────────────────────────────
ℹ Publish started.
## Begin Publishing Log ##
## End Publishing Log ##
Error in rsconnect::deployApp(...) : object 'bundle' not found
Calls: <Anonymous> -> saveDeployment -> deploymentRecord -> %||%
```

Expected: either the deployment record is saved without a bundle id, or `upload = FALSE` is rejected up front for Connect Cloud.

## Cause

In `R/deployApp.R`, `bundle` is assigned in the `if (upload)` branch. The `else` branch assigns it only for shinyapps.io:

```r
} else {
  # redeploy existing bundle
  if (isShinyappsServer(accountDetails$server)) {
    bundle <- application$deployment$bundle
  }
}
```

The Connect Cloud path that follows then reads it unconditionally:

```r
if (isPositConnectCloudServer(accountDetails$server)) {
  client$publish(application$id)
  ...
  saveDeployment(
    recordPath,
    deployment = deployment,
    application = application,
    bundleId = bundle$id,      # <- object 'bundle' not found
    metadata = metadata
  )
```

`bundle <- NULL` is set for Connect Cloud inside the upload branch ("PCC doesn't use bundle objects like other servers"), so `bundle$id` is fine when `upload = TRUE` and undefined when it is FALSE.

## Suggestion

Initialising `bundle <- NULL` before the `if (upload)` branch fixes it, since `NULL$id` is NULL and `deploymentRecord()` already treats `bundleId` as optional.

## Context

I reached for `upload = FALSE` while trying to recover content that had no `current_revision`, where a normal deployment fails with `Invalid token`. It was not the right tool for that, but the error was confusing enough to be worth reporting on its own.

### LINKED ISSUE #1370: Connect Cloud content with no current_revision can never be deployed again: Invalid token
Posit Connect Cloud content that has no `current_revision` fails **every** subsequent `deployApp()` with `Invalid token`, permanently. The cause is an assumption in `deployApp()` that a null `current_revision` means new content. It does not: republished content has none either, and then rsconnect never asks the API for a fresh bundle and uploads against a token that expired an hour after it was minted.

The failure looks like an authentication problem, which is what made it expensive to find. Nothing is wrong with the credentials.

## The assumption

`R/deployApp.R`:

```r
# Change _visibility_ & set env vars before uploading contents
if (isPositConnectCloudServer(accountDetails$server)) {
  # no update needed if we just created the content
  # current revision will be null only when creating new content
  if (!is.null(application$current_revision)) {
    ...
    application <- client$updateContent(
      application$id, deployment$envVars, newBundle = upload, primaryFile, appMetadata$appMode
    )
```

`updateContent(newBundle = TRUE)` is `PATCH /contents/{id}?new_bundle=true`, and it is the only request that mints a bundle and returns a fresh `source_bundle_upload_url`. When `current_revision` is null it is skipped, and the upload a few lines later uses whatever URL the content object already carried:

```r
uploadUrl <- application$next_revision$source_bundle_upload_url
success <- client$uploadBundle(bundlePath, uploadUrl)
if (!success) {
  cli::cli_abort("Could not upload bundle.")
}
```

That token has an `exp` one hour after it was minted, so after an hour every deployment of that content fails, forever.

## How content gets into that state

`POST /contents/{id}/republish` did it for us. Afterwards the content had no `current_revision` and a **published** `next_revision`, and it stayed that way — the API does not promote it.

## Reprex

```r
info   <- rsconnect:::accountInfo("<your account>", "connect.posit.cloud")
client <- rsconnect:::clientForAccount(info)
id     <- "<content id of a working, deployed app>"

# put the content into the state
client$withTokenRefreshRetry(rsconnect:::POST_JSON, paste0("/contents/", id, "/republish"), list())

c1 <- client$getContent(id)
is.null(c1$current_revision$id)   # TRUE
c1$next_revision$status           # "published"

# now wait for the upload token to age out (one hour), then deploy as usual
rsconnect::deployApp(appDir = "<the app>", appName = "<name>", account = "<your account>",
                     server = "connect.posit.cloud", forceUpdate = TRUE)
```

Observed:

```
>> {"error":"Invalid token"}
Error in `rsconnect::deployApp()`:
! Could not upload bundle.
```

Inspecting the token shows why — it is the bundle from the original creation, long expired:

```r
tok <- sub(".*token=", "", client$getContent(id)$next_revision$source_bundle_upload_url)
jsonlite::fromJSON(rawToChar(jsonlite::base64url_dec(strsplit(tok, ".", fixed = TRUE)[[1]][2])))
#> $bundle_id ... $exp 1788468025   # an hour after the content was created
```

Expected: `deployApp()` asks for a new bundle whatever the state of the content, and the deployment succeeds.

## Workaround

Make the request that rsconnect skips, just before deploying:

```r
if (is.null(client$getContent(id)$current_revision$id)) {
  client$updateContent(id, envVars = character(), newBundle = TRUE,
                       primaryFile = "app.R", appMode = "shiny")
}
```

That creates a pending revision with a fresh token, which the deployment then finds. Deleting and recreating the content is not necessary.

## Dead ends, for the record

None of these recover the content: `POST /contents/{id}/publish`, a second `republish`, `deployApp(upload = FALSE)` (which fails separately, #1369), and `POST /revisions/{id}/refresh_upload_url` — the last returns `409 Revision has already been published; cannot refresh upload URL`.

## Suggestion

Drop the `!is.null(application$current_revision)` guard, or base it on something that actually distinguishes new content, such as whether this call created it. The comment above the guard is the bug: it states an invariant the API does not hold.

vrsarah:
This PR fixes the two issues:

- **#1370**: Redeploying Posit Connect Cloud content whose initial publish failed permanently failed with "Invalid token". Such content is left with a null `current_revision` (no revision was ever promoted), and the guard `!is.null(application$current_revision)` treated that as newly-created content — so `updateContent` (the only call that mints a fresh `source_bundle_upload_url`) was skipped, and the deploy reused an expired token. The fix keys off whether this call created the content instead, so existing content always gets a fresh upload URL.

- **#1369**: `deployApp(upload = FALSE)` on Posit Connect Cloud errored with `object 'bundle' not found` because `bundle` was only assigned inside the `if (upload)` branch, but read unconditionally afterward.

### Fixes

**#1370**: Replace the `current_revision` guard with a `createdContent` flag that tracks whether _this call_ created or 404-recreated the content:
- `isNewContent <- is.null(deployment$appId)` before the create/lookup block
- `isNewContent <<- TRUE` inside the PCC 404 handler (recreated content already has a pending revision)
- Guard becomes `if (!isNewContent)` — existing content always gets `updateContent`; freshly created/recreated content skips it

**#1369**: Initialize `bundle <- NULL` once before the `if (upload)` block; remove the now-redundant `bundle <- NULL` inside the PCC upload branch. `bundle$id` is `NULL` when `upload = FALSE`, which `deploymentRecord()` already handles as optional.

### DIFF
--- a/R/deployApp.R
+++ b/R/deployApp.R
@@ -525,6 +525,8 @@ deployApp <- function(
     checkConnectSupportsNodejs(client)
   }
 
+  createdContent <- is.null(deployment$appId)
+
   if (is.null(deployment$appId)) {
     taskStart(quiet, "Creating content on server...")
     if (isPositConnectCloudServer(accountDetails$server)) {
@@ -563,6 +565,7 @@ deployApp <- function(
           application
         },
         rsconnect_http_404 = function(err) {
+          createdContent <<- TRUE
           application <- applicationDeleted(
             client,
             deployment,
@@ -615,9 +618,10 @@ deployApp <- function(
 
   # Change _visibility_ & set env vars before uploading contents
   if (isPositConnectCloudServer(accountDetails$server)) {
-    # no update needed if we just created the content
-    # current revision will be null only when creating new content
-    if (!is.null(application$current_revision)) {
+    # Existing content: mint a fresh bundle + upload URL. Skipped only when we
+    # just created (or recreated) the content in this call, which already has a
+    # pending revision.
+    if (!createdContent) {
       taskStart(quiet, "Updating content...")
       # Use appPrimaryDoc if available, otherwise fall back to inferredPrimaryFile
       primaryFile <- appMetadata$appPrimaryDoc %||%
@@ -649,6 +653,8 @@ deployApp <- function(
       taskComplete(quiet, "Environment variables updated")
     }
   }
+
+  bundle <- NULL
 
   if (upload) {
     python <- getPythonForTarget(python, accountDetails)
@@ -687,7 +693,6 @@ deployApp <- function(
       if (!success) {
         cli::cli_abort("Could not upload bundle.")
       }
-      bundle <- NULL # PCC doesn't use bundle objects like other servers
     } else if (isShinyappsServer(accountDetails$server)) {
       bundle <- uploadShinyappsBundle(
         client,
--- a/tests/testthat/test-deployApp.R
+++ b/tests/testthat/test-deployApp.R
@@ -337,3 +337,210 @@ test_that("openURL() launches the browser on success with a valid url", {
   )
   expect_true(launched)
 })
+
+# PCC deploy: updateContent / createdContent guard (#1370, #1369) ----------
+
+# Shared PCC test content fixture: existing content with NULL current_revision.
+pcc_existing_content_null_revision <- list(
+  id = "content-abc",
+  url = "https://connect.posit.cloud/myaccount/content/content-abc",
+  current_revision = NULL,
+  next_revision = list(
+    id = "rev-new",
+    source_bundle_upload_url = "https://upload.example.com/bundle"
+  )
+)
+
+# Shared PCC test content fixture: existing content with a current_revision.
+pcc_existing_content_with_revision <- utils::modifyList(
+  pcc_existing_content_null_revision,
+  list(current_revision = list(id = "rev-old"))
+)
+
+# Set up a PCC server, account, and (optionally) a re-deploy record.
+local_pcc_deploy_env <- function(
+  appDir,
+  appId = "content-abc",
+  env = parent.frame()
+) {
+  local_temp_config(env = env)
+  addTestServer(
+    url = "https://connect.posit.cloud",
+    name = "connect.posit.cloud"
+  )
+  addTestAccount("myaccount", server = "connect.posit.cloud")
+  if (!is.null(appId)) {
+    addTestDeployment(
+      appDir,
+      appName = "myapp",
+      appId = appId,
+      account = "myaccount",
+      server = "connect.posit.cloud"
+    )
+  }
+}
+
+test_that("#1370: existing PCC content with NULL current_revision calls updateContent", {
+  skip_on_cran()
+  appDir <- local_temp_app(list("app.R" = "library(shiny)"))
+  local_pcc_deploy_env(appDir)
+
+  update_content_called <- FALSE
+  local_mocked_bindings(
+    clientForAccount = function(...) {
+      list(
+        getContent = function(id) pcc_existing_content_null_revision,
+        updateContent = function(id, envVars, newBundle, primaryFile, appMode) {
+          update_content_called <<- TRUE
+          pcc_existing_content_null_revision
+        },
+        uploadBundle = function(bundlePath, url) TRUE,
+        publish = function(id) invisible(NULL),
+        awaitCompletion = function(revisionId) {
+          list(
+            success = TRUE,
+            url = "https://connect.posit.cloud/myaccount/content/content-abc"
+          )
+        }
+      )
+    },
+    bundleApp = function(...) {
+      tmp <- tempfile(fileext = ".tar.gz")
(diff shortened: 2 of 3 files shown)


tests/testthat/test-deployApp.R
```diff
@@ -337,3 +337,210 @@ test_that("openURL() launches the browser on success with a valid url", {
...
+  if (!is.null(appId)) {
+    addTestDeployment(
+      appDir,
+      appName = "myapp",
+      appId = appId,
+      account = "myaccount",
+      server = "connect.posit.cloud"
+    )
+  }
+}
```
karawoo, line 381: let's move these into helper.R

tests/testthat/test-deployApp.R
```diff
@@ -337,3 +337,210 @@ test_that("openURL() launches the browser on success with a valid url", {
...
+
+test_that("#1370: existing PCC content with NULL current_revision calls updateContent", {
```
karawoo, line 383: could you please remove the issue numbers from the test descriptions?

R/deployApp.R
```diff
@@ -525,6 +525,10 @@ deployApp <- function(
     checkConnectSupportsNodejs(client)
   }
 
+  # New content (no existing target, or the target was deleted and recreated
+  # below) already has a fresh pending revision, so it skips updateContent().
+  isNewContent <- is.null(deployment$appId)
```
karawoo, line 530: "was deleted and recreated below" is confusing because it sounds like something below has happened before this line. could you clarify this comment?

tests/testthat/test-deployApp.R
```diff
@@ -337,3 +337,210 @@ test_that("openURL() launches the browser on success with a valid url", {
...
+  update_content_called <- FALSE
+  local_mocked_bindings(
+    clientForAccount = function(...) {
+      list(
+        getContent = function(id) pcc_existing_content_null_revision,
+        updateContent = function(id, envVars, newBundle, primaryFile, appMode) {
+          update_content_called <<- TRUE
+          pcc_existing_content_null_revision
+        },
+        uploadBundle = function(bundlePath, url) TRUE,
```
karawoo, line 397: can you add an assertion that the url provided is the new url and not the old one?

R/deployApp.R
```diff
@@ -563,6 +567,7 @@ deployApp <- function(
           application
         },
         rsconnect_http_404 = function(err) {
+          isNewContent <<- TRUE
```
karawoo, line 570: could this function include `isNewContent` as part of what it returns instead of using `<<-`?

R/deployApp.R
vrsarah: agreed, updated the comment!

tests/testthat/test-deployApp.R
vrsarah: oops 😬

karawoo (approved):

PR #1378, opened 2026-09-10
Fix Connect Cloud deploy crash with no env vars on R < 4.2.0

vrsarah:
`deployApp()` to Posit Connect Cloud fails with:

```
Error in mapply(FUN = f, ..., SIMPLIFY = FALSE):
  zero-length inputs cannot be mixed with those of non-zero length
```

when the deployment has **no environment variables**, on R < 4.2.0.

### Root cause

Both `createContent` and `updateContent` in `R/client-connectCloud.R` build the `secrets` payload with:

```r
secrets <- unname(Map(
  function(name, value) list(name = name, value = value),
  envVars,
  Sys.getenv(envVars)
))
```

When `envVars` is `NULL` or `character(0)`, `Sys.getenv(envVars)` returns the **entire process environment** (a long named character vector). `Map()` then receives a zero-length first argument and a non-zero-length second argument. R >= 4.2.0 tolerates this and returns an empty list silently; R < 4.2.0 throws the error above.

### Fix

Guard both call sites with `length(envVars) > 0` so `Sys.getenv()` is never called on an empty vector:

```r
secrets <- if (length(envVars) > 0) {
  unname(Map(function(name, value) list(name = name, value = value),
             envVars, Sys.getenv(envVars)))
} else {
  list()
}
```

This matches the existing `length(...) > 0` guard pattern used elsewhere in the codebase (e.g. `R/deployApp.R` around the `setEnvVars` call).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # rsconnect (development version)
 
+* Deploying to Posit Connect Cloud content that has no environment variables
+  no longer fails on R < 4.0.0 with "zero-length inputs cannot be mixed with
+  those of non-zero length". The Connect Cloud client no longer calls
+  `Sys.getenv()` on an empty vector when building the secrets payload.
+
 * `showUsers()` and `showInvited()` now always return a data frame with
   atomic `character`/`logical` columns, including a typed 0-row data frame
   (rather than `NULL`) when there are no results. User ids are now always
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # rsconnect (development version)
 
 * Deploying to Posit Connect Cloud content that has no environment variables
-  no longer fails on R < 4.0.0 with "zero-length inputs cannot be mixed with
+  no longer fails on R < 4.2.0 with "zero-length inputs cannot be mixed with
   those of non-zero length". The Connect Cloud client no longer calls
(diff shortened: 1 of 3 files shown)


R/client-connectCloud.R
```diff
@@ -246,16 +246,20 @@ connectCloudClient <- function(service, authInfo) {
...
+              name = name,
+              value = value
+            )
+          },
+          envVars,
+          Sys.getenv(envVars)
+        ))
+      } else {
+        list()
+      }
```
karawoo: This works but reads kind of weirdly in R. I would do

```suggestion
      secrets <- list()
      if (length(envVars) > 0) {
        secrets <- unname(Map(
          function(name, value) {
            list(
              name = name,
              value = value
            )
          },
          envVars,
          Sys.getenv(envVars)
        ))
      }
```

and the same below

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # rsconnect (development version)
 
+* Deploying to Posit Connect Cloud content that has no environment variables
+  no longer fails on R < 4.2.0 with "zero-length inputs cannot be mixed with
+  those of non-zero length". The Connect Cloud client no longer calls
+  `Sys.getenv()` on an empty vector when building the secrets payload.
```
```suggestion
  those of non-zero length".
```

karawoo (approved):

